PR #1, opened 2019-04-05
add layout functions

paulf81:
Adding the set of layout tools from wind_tools, the script is admittedly not up to our current standards, but I was finding helpful in doing this EDF pre-study so propose to throw it in and clean it up later, probable cleaning targets would be removing pandas, adding some documentation, and not sure what happened but the distance plot not looking perfect, but still useful


paulf81:
Cool if we just pull the trigger on this one?  It breaks nothing, just adds some less than perfect code...

rafmudaf:
merged and branch deleted

PR #5, opened 2019-04-25
Update running procedures

paulf81:
Submitting this pull request of Raf and my work today now before it includes too many changes.  This is most of the overhaul though that will affect API though, but please do have a quick look over to make sure you agree with these changes.   Note I updated all the examples to demonstrate new methods so perhaps start there...

 Main ideas:
- Interface provides most ways in to floris -- fi.floris.farm.... should no longer appear in examples
- No more direct setting of floris variables, variables are set either through reinitialize flow field, or through calculate wake, depending on what is appropriate
- Removed convenience wrappers and duplicate definitions from floris/floris.farm/floris.farm.flow_field since the interface provides a clean access, these could just lead to confusion if variables are defined twice at different levels

-- I'm still working on through
- Turbine Map I believe could be further cleaned
- Speeding up the visualization calls, and some of the calls more generally


bayc:
Wow, that's a lot of changes, great work! Will need to go through in depth tomorrow, but at first pass, minor things so I don't forget:

1) The getters in farm.py will need their documentation updated so that the examples of setting the values are removed.
2) Convenience typo in floris_utilities.py, lines 30 and 49.

paulf81:
None of my commits have passed goin

paulf81:
You're both right, I need to update the docs, and @rafmudaf maybe you and I can track down the CI issues?  We can also discuss a bit more on the best balance of convenience for users and coders

bayc:
Last night it looked like the test files just needed to be updated. It seemed they were using functions/methods that had been eliminated.

rafmudaf:
@bayc I have some updates for the tests but yes its generally just updating the api. I'll fix those asap!

ejsimley:
Nice idea to set variables through only reinitialize flow field and calculate wake.

1) Not directly related to these changes, but it would be nice to adjust the domain bounds when plotting horizontal planes for wind directions other than 270. For example in example_0005 figure 2, the plots don't show much of the wakes for wind directions 170 and 45. 

2) I can redo documentation for farm after these changes are merged.

paulf81:
A bunch more commits, now tools has FlowData while only simulation has FlowField, clearing up the ambiguity.  Also, flow-data lets you set the domain size, and I added an option to extract only points near hub-height for more quickly making horizontal cut throughs.  Fixed github.com references also.  I updated and checked all the examples, @rafmudaf we may need to double check the tests, and I know the docs are now needing updating

paulf81:
I've updated now the example documentation, are we ok to pull into develop?

rafmudaf:
I added one commit removing some commented code and using a built-in attribute to get the model string. Otherwise, my only comment is that it might be nice to allow setting the domain bounds even when using the curl model.

rafmudaf:
But I think that could be dealt with later. I think this pull request solves the problem of having a large domain when its not needed.

paulf81:
Yeah, we can keep improving, I'm going to merge it in, thanks all!

PR #8, opened 2019-05-14
New vis example

paulf81:
**Complete this sentence**
Adds an example that shows how to do a quiver cut-through visualization, and export to vtk

Adds a quiver visualization method to the visualization module, and an example which uses it


paulf81:
@rafmudaf could you check on flow_data.py?  one issue which came up which I'm not quite sure how to fix, in the function "save_as_vtk", a number of values are written to file using .format, they then appear in the file as floats, and paraview errors trying to read them.  Could we force the integer values (dimensions, point_data, and "Uavg 3...") to be ints?

rafmudaf:
@paulf81 I've added a fix for that in the Vec3 class. It now checks if all of the arguments are the same type rather than forcing them all to be `float`. Now the vtk file works straight from the example. Check it out!
wakes

paulf81:
Wow, the visualization is looking great!  What a nice touch!

paulf81:
I checked your version @rafmudaf and works fine, I also add a documentation file, good to go?

bayc:
@raf that looks awesome, great work! @paulf81 looking through the code, I am wondering if you need a separate `example_input_curl.json`. You can just use the original input file and change the model to `curl` in the example file using the `set_wake_model` method, right? The curl model parameters are the same in the `example_input.json` file, but I did notice the Cp, Ct, and wind speed tables had changed. Is there a reason for this change?

paulf81:
That is a great catch @bayc , I updated the code in this commit

PR #11, opened 2019-06-11
Several small fixes

paulf81:
**Complete this sentence**
THIS PULL REQUEST __ Is__ READY TO MERGE

A small set of bug fixes to correct for calculation of ct, and catch issues in very low wind speeds or ct values over 1.0


paulf81:
Need to update regression tests...

ejsimley:
I thought we could set Cp to 0 when below the minimum wind speed. Does this seem reasonable? I realized that Ct = 0.0 will cause some divide by zero errors in the wake calculations, so I'll change that back to 0.99

paulf81:
Right, the alternative was to either error if they don't define down to 0, or else force a 0 answer into the table, but this seemed simplest and mostly correct.  The current behavior can be chaotic if you ask for values below the defined speeds.  Since most turbines cut in 4/5 this seems pretty safe, what do you think?

jrannoni:
I agree with most of that.  But I was worried that you would run into what @ejsimley described and so instead of getting 0.0 power you get NAN.

PR #12, opened 2019-06-12
Add a new option to compute yaw loss by w-exp

paulf81:
This is not a real pull request, I was just thinking this pull format, which quickly highlights changes a convenient way to get your opinion on how to do this, I'll explain more on slack

### DIFF
--- a/floris/simulation/turbine.py
+++ b/floris/simulation/turbine.py
@@ -402,27 +402,6 @@ class Turbine():
         """
         return np.cbrt(np.mean(self.velocities**3))
 
-    # @property
-    # def Cp(self):
-    #     """
-    #     This property returns the power coeffcient of a turbine.
-
-    #     This property returns the coefficient of power of the turbine 
-    #     using the rotor swept area average velocity, interpolated from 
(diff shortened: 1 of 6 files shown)


floris/simulation/turbine.py
```diff
@@ -424,6 +424,29 @@ def Cp(self):
         """
         return self._fCp(self.average_velocity)
 
+    # @property
+    def Cp_new(self, scaled_velocity):
```
nhamilto: Ultimately, we should avoid names with '_new'. This will undoubtedly lead to issues in the future. We can add a deprication warning simply update the existing namespace.

paulf81:
Ok, this version should be more ready

floris/simulation/turbine.py
paulf81: this should be fixed now

ejsimley:
Should we set p_w = pP/3? This would match the power in region 2 that we are used to, after cubing the yaw effective wind speed.

And would it make sense to use yaw effective wind speed to calculate C_T? I know you saw higher thrust after yawing above rated, so maybe this doesn't work.

Issue #15, opened 2019-06-17
Bug report


jfst:
**Bug description**
The calculate_wake() function produces

 different result even though the input is the same.

**To Reproduce**
Run the code in the Additional context section. The input file is the one provided in the example directory of the floris package.

**Expected behavior**
The total output power differ for the same input wind direction (-270) when calc_wake() is called between with another wind direction (-90) .

Wind speed/dir:  8.0 / -270
Power:  **4.470527954624492**

Wind speed/dir:  8.0 / -90
Power:  5.276564072664706

Wind speed/dir:  8.0 / -270
Power:  **5.276564072664707**

Wind speed/dir:  8.0 / -90
Power:  5.276564072664706

**Screenshots, if applicable**
<Add screenshots to help explain your problem.>

**Floris Version**
v1.1.2

**System Information (please complete the following information):** 
 - OS: Win10
 - Library versions
   * matplotlib: 3.0.3
   * numpy: 1.16.3
   * pytest: 4.3.1
   * scipy: 1.1.0
   * Sphinx: 1.8.5

**Additional context**
`import numpy as np
import floris.tools as wfct

fi = wfct.floris_utilities.FlorisInterface('example_input.json')
print()

fi.reinitialize_flow_field(wind_direction = 0)
print('Wind speed/dir: ', fi.floris.farm.wind_speed, '/', fi.floris.farm.wind_direction)
fi.calculate_wake()
print('Power: ', np.sum(fi.get_turbine_power())/1e6)
print()

fi.reinitialize_flow_field(wind_direction = 180)
print('Wind speed/dir: ', fi.floris.farm.wind_speed, '/', fi.floris.farm.wind_direction)
fi.calculate_wake()
print('Power: ', np.sum(fi.get_turbine_power())/1e6)
print()

fi.reinitialize_flow_field(wind_direction = 0)
print('Wind speed/dir: ', fi.floris.farm.wind_speed, '/', fi.floris.farm.wind_direction)
fi.calculate_wake()
print('Power: ', np.sum(fi.get_turbine_power())/1e6)
print()

fi.reinitialize_flow_field(wind_direction = 180)
print('Wind speed/dir: ', fi.floris.farm.wind_speed, '/', fi.floris.farm.wind_direction)
fi.calculate_wake()
print('Power: ', np.sum(fi.get_turbine_power())/1e6)
print()`

bayc:
[...]

PR #16, opened 2019-06-20
Fix to correctly reinitialize turbine turbulence intensity when reinitializing the flow field.

bayc:
**Complete this sentence**

A bug was identified where a turbine's turbulence intensity (TI) was retained even after reinitializing the flow field. Code was changed to correctly reset the TI using the `reinitialize_turbines()` method. This was a combined effort from @ejsimley , @rafmudaf , and @bayc .

https://github.com/NREL/floris/issues/15, raised by @jfst

This will change the power output of wind farms when re-calculated at different wind directions.

Below, the problem is demonstrated before the fix, and then the results after the fix are shown.

Running the `example_input.json` file for an input wind direction of 270 deg, the farm and turbine powers produced are:

__Case 1 before the bug fix:__
```
Farm Power:  4.494816398178318
Turbine Powers:  [1690358.3095915292, 557049.8894976295, 1690358.3095915292, 557049.8894976295]
```
After reinitializing the flow field, and then re-calculating the wake for an input wind direction of 90 deg should return the same farm power and reversed turbine powers, but instead returned:

__Case 2 before the bug fix:__
```
Farm Power:  5.235071691416997
Turbine Powers:  [927177.5361169698, 1690358.3095915292, 927177.5361169698, 1690358.3095915292]
```

This can be seen in the images below, where the first case is shown in the first image and the second case in the second image. The wakes of the upstream turbine in the second image can be seen to be less strong than the wakes in the first image. This is because the TI for the downstream turbines of the first case is retained and uses by the upstream turbines in the second case. By correctly reinitializing the TI for all turbines when the flow field is reinitialized, the two cases return the same powers and the wakes appear equal (shown in the 3rd and 4th figures).

__Case 1 after the bug fix:__
```
Farm Power:  4.494816398178318
Turbine Powers:  [1690358.3095915292, 557049.8894976295, 1690358.3095915292, 557049.8894976295]
```

__Case 2 after the bug fix:__
```
Farm Power:  4.494816398178318
Turbine Powers:  [557049.8894976297, 1690358.3095915292, 557049.8894976295, 1690358.3095915292]
```

__Figure 1: First case with a wind direction of 270 deg before the bug fix.__
old_wd_270

__Figure 2: Second case with a wind direction of 90 deg before the bug fix.__
old_wd_90

__Figure 3: First case with a wind direction of 270 deg after the bug fix.__
new_wd_270

__Figure 4: Second case with a wind direction of 90 deg after the bug fix.__
new_wd_90

The code used to produce these values and images is posted below.

```
import numpy as np
import floris.tools as wfct

fi = wfct.floris_utilities.FlorisInterface('example_input.json')

fi.reinitialize_flow_field(wind_direction = 270)
print('Wind speed/dir: ', fi.floris.farm.wind_speed, '/', fi.floris.farm.wind_direction)
fi.calculate_wake()
print('Farm Power: ', fi.get_farm_power()/1e6)
print('Turbine Powers: ', fi.get_turbine_power())
print()

# Initialize the horizontal cut
hor_plane_1 = wfct.cut_plane.HorPlane(
    fi.get_flow_data(),
    fi.floris.farm.turbines[0].hub_height
)

fi.reinitialize_flow_field(wind_direction = 90)
print('Wind speed/dir: ', fi.floris.farm.wind_speed, '/', fi.floris.farm.wind_direction)
fi.calculate_wake()
print('Farm Power: ', fi.get_farm_power()/1e6)
print('Turbine Powers: ', fi.get_turbine_power())
print()

# Initialize the horizontal cut
hor_plane_2 = wfct.cut_plane.HorPlane(
    fi.get_flow_data(),
    fi.floris.farm.turbines[0].hub_height
)

# Plot and show
fig, ax = plt.subplots()
wfct.visualization.visualize_cut_plane(hor_plane_1, ax=ax)
fig, ax = plt.subplots()
wfct.visualization.visualize_cut_plane(hor_plane_2, ax=ax)
plt.show()
```

### DIFF
--- a/floris/simulation/flow_field.py
+++ b/floris/simulation/flow_field.py
@@ -311,8 +311,6 @@ class FlowField():
             self.wind_veer = wind_veer
         if turbulence_intensity is not None:
             self.turbulence_intensity = turbulence_intensity
-            for turbine in self.turbine_map.turbines:
-                turbine.turbulence_intensity = self.turbulence_intensity
         if air_density is not None:
             self.air_density = air_density
             for turbine in self.turbine_map.turbines:
@@ -335,7 +333,7 @@ class FlowField():
 
         # reinitialize the turbines
         for turbine in self.turbine_map.turbines:
-            turbine.reinitialize_turbine()
+            turbine.reinitialize_turbine(self.turbulence_intensity)
 
     def calculate_wake(self, no_wake=False):
         """
--- a/floris/simulation/turbine.py
+++ b/floris/simulation/turbine.py
@@ -93,7 +93,7 @@ class Turbine():
             raise ValueError(
                 "Turbine.grid_point_count must be the square of a number")
 
-        self.reinitialize_turbine()
+        self.reinitialize_turbine(turbulence_intensity=0.0)
 
         # initialize derived attributes
         self.grid = self._create_swept_area_grid()
@@ -286,7 +286,7 @@ class Turbine():
             rotated_z
         )
 
-    def reinitialize_turbine(self):
+    def reinitialize_turbine(self, turbulence_intensity):
         """
         This method sets the velocities at the turbine's rotor swept 
         area grid points to zero.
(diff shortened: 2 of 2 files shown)


bayc:
Yea @ejsimley , Raf and I were thinking the same thing, but since it doesn't have knowledge and we were wanting to get the fix in, we went with `None` so that if a user calls the function and doesn't supply a value then the code will fail when it tries to use the value. I'm thinking we can re-work this in the next major version, along with Nicholas `wake_turbulence.py` module.

PR #17, opened 2019-06-24
Hot fix vis

paulf81:
This pull shows a solution to the "all blue" cut through visualization.  I believe the issue comes from a numerical issue causing the grid data function to rule certain values extrapolation instead of interpolation.  To show, I've added some temp examples which will/won't generate the blue plot issue depending which line you comment/uncomment near line 73 in cut_plane.py, let me know if this works for you and if this solution (a small perturbation to break the tie) works


ejsimley:
This fix works for me, thanks Paul! I'm checking to see if it solves the problem for the larger wind farm David was looking at.

paulf81:
I revised this, the previous version I think wouldn't have helped if the minimum x, or y, was positive, this version should be robust, and will remove the issue

PR #18, opened 2019-06-28
Add a json update tool

paulf81:
Here I add a first pass function which converts the current JSON file to the new one proposed by @nhamilto .  Doing this also raised a few questions for me I wanted to clarify to make sure I understand all changes:

1) The new version, in the curl model, includes parameters initial, constant, ai, downstream, but don't these belong in the turbulence model?
2) I think in the past we defined gauss/jimenez twice, then we moved to one time, is the idea we now go back to defining twice to allow for greater modularity?  Just want to be sure,


nhamilto:
Hey Paul. I'll look at your new tool. As to your questions:

1. The curl model requires the initial, constant, ai, and downstream TI parameters. Previously they were part of the general 'parameters' level of the JSON input file. After breaking this up into separate parameter dictionaries for 'wake_velocity_parameters', 'wake_turbulence_parameters', and 'wake_deflection_parameters', @bayc and I figured out that the Curl model never made a call to wake_turbulence.py, and thus couldn't see the relevant TI parameters. We added them to the Curl wake_velocity model as a workaround.
2. There are indeed multiple definitions named gauss/jimenez/ishihara but they live in multiple parameter dictionaries. We could certainly reorganize into a single dictionary for each type. If we do so, only part of the dictionary would be used for wake_velocity, wake_turbulence, etc., but it might simplify the input.

paulf81:
That sounds good then, I just wanted to be sure to ask to make sure these changes were intentional, so long as they are, I'm ok with it, I'm gonna clean up the script just a little now...

bayc:
Yea @paulf81 , the updater looks good. As for the input dictionary and its divisions, I think I am currently torn. Having a dictionary for each module makes sense in that you can then define individually what `wake_velocity` uses and what `wake_deflection` uses, but from a user stand point, it means having to update parameters in two locations when tuning, which can be an easy place to introduce mistakes. Not sure which way to go...

paulf81:
I'm also torn about it, it's the kind of thing (not noticing that ka is definied twice, once for recovery, once for deflection) where I can too easily imagine making mistakes myself, but on the other hand, like Nicolas is saying, it's a more accurate and complete input,

Just off the top of my head, what if there was only one definition of gauss, but included like:
ka_defecit: ....
ka_deflection: ...

To make as clear as possible where each is going and to remind users ka is definied twice, and also nudge them that all things being equal they should be the same?

PR #20, opened 2019-07-05
Resample in bootstrap weighted by dist to bin cent

paulf81:
Small update that adds weights to the bootstrap resampling in energy ratio to favor points closest to bin center.  Allows a compromise between bigger bins, but prioritizing points in center of bin to outer-most


bayc:
Yea, looks good to me too!

paulf81:
Done, thanks!

PR #24, opened 2019-07-22
Add option to allow a deflection multipler def

paulf81:
**Complete this sentence**
Ready to merge

Allow an option to specify the deflection multiplier in the json.  If not defined, default to the current 1.2,


bayc:
Hey @paulf81 , this seems alright to me. A simple way to include something for advanced usage but not require it/handle it if needed. I would suggest a more descriptive print statement, maybe even just adding "deflection" in front of multiplier would do it.

paulf81:
Nice suggestion @bayc , cool if I merge now?

PR #30, opened 2019-09-26
Change bounds in FlorisInterface.get_hub_height_flow_data

knutss:
I tried to use the x_bounds and y_bounds arguments in the FlorisInterface.get_hub_height_flow_data function, the bounds are passed to flow_field.set_bounds, but then flow_field.reinitialize_flow_field calls self.set_bounds without an argument and the bounds are reset. To fix this I suggest to pass a bounds_to_set argument to reinitialize_flow_field, and set the bounds there.

BR,
Knut S. Seim

### DIFF
--- a/floris/simulation/flow_field.py
+++ b/floris/simulation/flow_field.py
@@ -293,6 +293,10 @@ class FlowField():
             with_resolution: A :py:class:`floris.utilities.Vec3` object 
                 that defines the flow field resolution at which to 
                 calculate the wake (default is *None*).
+            bounds_to_set: A list of values representing the mininum 
+                and maximum values for the domain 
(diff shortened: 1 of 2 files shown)


rafmudaf:
@knutss I've fixed the merge conflicts here. Could you give me (@rafmudaf) write access to your repository? You can do this by adding me as a contributor.

rafmudaf:
I'm asking for access because I've rewritten a bit of the git history and will need to force-push before merging. Since you started this branch off of the `master` branch, it was a bit behind the `develop` branch, but I've fixed that.

rafmudaf:
Thanks @knutss, feel free to take me off of your repo now.

Issue #29, opened 2019-08-29
PyPI deploy


markusdregi:
Dear Floris developers,

I see that you have multiple versions here at _GitHub_ that is not deployed to PyPI yet. Do you have any plans of getting versions later then `1.0.0` on PyPI any time soon?

Reference: https://pypi.org/project/FLORIS/#history

Would you be interested in a pull request that would have _Travis_ automatically deploy new versions of Floris to PyPI whenever you create new git tags? We have done so for multiple of our repositories in _Equinor_ and it works really smoothly.

Best regards,
Markus Dregi

rafmudaf:
Hi @markusdregi 

You're right, we have been lazy in keeping `pypi` up to date. And yes, a pull request for TravisCI to automatically deploy new tags to `pypi` would be most welcome!

Thank you for any help in this area.

Rafael

markusdregi:
> And yes, a pull request for TravisCI to automatically deploy new tags to pypi would be most welcome!

Perfect, then a PR might be on its way ;)

Cheers

PR #33, opened 2019-10-08
Automatic release to pypi on tags

sondreso:
Adds deploy settings to travis configurations. This will publish the package to PyPi when a new tag is added on Github.

Before merging you should push a commit with where the username and password placeholders are replace with your username and encrypted password (see https://docs.travis-ci.com/user/deployment/pypi/). 

Best regards, 
Sondre Sortland


rafmudaf:
Thanks @sondreso. This will solve Issue #29.

rafmudaf:
@sondreso The `setup.py` currently fails on install since `VERSION` is no longer defined. Not sure why TravisCI didn't run on this pull request. I'll fix that so you can get feedback immediately.

rafmudaf:
I like the PyPi stuff here but haven't been able to get the version info to work correctly with `setuptools_scm`. My biggest issue is that the version is given as `1.1.5.dev1+gf0cdb9a.d20191008` but the latest tag is `v1.1.4`. This is described in the [docs](https://pypi.org/project/setuptools-scm/) but I find it very misleading to roll the version by default. Instead, the version should be the latest tag with the number of commits ahead and the current commit. This probably isn't an issue when deploying, but it is a problem when running a local editable install as many of us do.

sondreso:
Alright, then i will change the PR to only use the automatic pypi deploy, with the old manual version number. It will not be an issue as long as you remember to bump the version number manually with new releases.

PR #36, opened 2019-12-06
Propose speed ups

paulf81:
This is another "fake" pull request, just a convenient way to show some more ideas I've been trying.  Matt Shields asked Jen and I to look at if the 30s it takes FLORIS to compute a baseline AEP for 4 turbines is correct.  That baseline AEP can be checked using the aep_calcuation.py example.  On my computer, the proposed changes reduce the time it takes down to 15s.

I've been using cProfile and snakeviz to identify the slow parts and they're mostly what you'd expect, appends and list comprehensions which could be re-cast as numpy vector operations.  I don't think I'm proposing to actually merge these specific changes, but show what we might do to speed up FLORIS if we want to.  Below I'll copy my notes I took as I went along

~~
 NOTES
Starting from modded example from Matt shields, initial run completes in 30.57 s

# Removing print out of progreess
29 s

# ERG WRONG BRANCH!!!

Starting over on optimization branch with printouts…
Running examples/profile_examples/aep_calculation.py

BASELINE
Total time = 29.8 s

FIRST EDIT 
REMOVE PRINTOUTS
Total time = 29.9 s

Ok, checking perf in snakeviz

SECOND EDIT
In optimization .py around line 820, there is a dataframe being built up using append, switching this to a dict based method according to:
https://stackoverflow.com/questions/27929472/improve-row-append-performance-on-pandas-dataframes

Total time is now 20 s!

THIRD EDIT

Try removing calls to utility/cosd function and put in time np.cos(np.radians
directly

19.98 s, not a very big improvement, undoing this for now… (only 2% better, look for better opportunities, referring)

Fourth EDIT

In turbine.py, replace list comps with a call to scipi kdtree, to do a search for nearest neighbors

17.3 s

A little better, but probably simple matrix math could be better, try that next…

Fifth Edit

Replace tree with a distance matrix and argmin, 

16.3 s

Sixth Edit

Precompute and store interp functions in turbine module

15 s


paulf81:
@bayc @rafmudaf  I think this is now cleaned up and ready to, ok to pull in?

bayc:
@paulf81 This looks good, thanks for tackling this! I just made some adjustments to the line formatting to better match the rest of the code.

The kd approach for nearest neighbors was interesting, hadn't seen that before! And the dictionary updates replacing the appends to the df makes sense.

One thing I am unsure of is the purpose of the `np.mean` in the previous code, [line 212](https://github.com/NREL/floris/pull/36/files#diff-ea90fa8f5e8cb1f301357ed51500b255R212) in the latest version of `turbine.py`. I noticed the new code still returns 9 velocity values across the rotor, but couldn't track down what exactly the previous code was doing. Otherwise, it all looks good to me.

floris/simulation/turbine.py
```diff
@@ -186,19 +193,39 @@ def calculate_swept_area_velocities(self, wind_direction, local_wind_speed, coor
...
-        zPts = np.array([point[1] for point in self.grid])
-
-        # interpolate from the flow field to get the flow field at the grid points
-        dist = [np.sqrt((coord.x1 - x_grid)**2 + (coord.x2 + yPts[i] - y_grid) **
-                        2 + (self.hub_height + zPts[i] - z_grid)**2) for i in range(len(yPts))]
-        idx = [np.where(dist[i] == np.min(dist[i])) for i in range(len(yPts))]
-        data = [np.mean(u_at_turbine[idx[i]]) for i in range(len(yPts))]
-        return np.array(data)
+        
+        # PREVIOUS MEHTHOD
```
rafmudaf: Can we remove the previous method? Its in the git history...

paulf81:
Ok, cleaned up and talked to you all, merging it

PR #37, opened 2019-12-17
Add gch to optimization

paulf81:
Another semi-request, I think I've uncovered a general issue, run example
example_0006b_weird_behavior_Gauss.py

This is upstream of finalizing my checkouts of gch...


rafmudaf:
These have been merged into develop in pull request #40.

PR #38, opened 2020-01-02
Break out the GCH wake models into distinct model classes

rafmudaf:
**Complete this sentence**
THIS PULL REQUEST IS READY TO MERGE

This pull request creates a new class for the Gauss-Curl hybrid wake models. Before, this wake model was built into the existing Gauss model.

None, but we talked about this in a meeting.

Primarily `WakeVelocity` and `WakeDeflection`.

See that the automated tests pass.


bayc:
Hey @rafmudaf , this looks good to me! I see some to-do's in there, so we can just add those to the list and get this working.

rafmudaf:
@bayc that's right, I've added some to-do's that we should address before merging into `develop`. Thanks for the review.

PR #44, opened 2020-01-08
Fix interpolation issues

paulf81:
I fixed the seg fault issue by setting up a new FLORIS enviornment in conda.  Then I reverted the 'nearest' flags back to 'cubic' interpolation.

Also, I confirmed the heterogenous example works now, needed GCH added to het json


rafmudaf:
Nicely done

PR #47, opened 2020-01-09
Propose turbine change code

paulf81:
This pull doesn't currently work, sharing to get opinions on current implementation and next steps.  Basically, @jrannoni and I were talking and thinking a simple way to start enabling farms of mixed turbines, without complexifying the input, is keep one turbine on input, and then add others in code.  To that end I've been working on a new interface function and example.  You can see in the changed files how I started going about it, and where I get stuck

### DIFF
--- a/examples/example_0008_modify_turbine.py
+++ b/examples/example_0008_modify_turbine.py
@@ -11,6 +11,9 @@
 
 # See read the https://floris.readthedocs.io for documentation
 
+# This example illustrates changing the properties of some of the turbines
+# This can be used to setup farms of different turbines
+
 import matplotlib.pyplot as plt
 import floris.tools as wfct
 
--- a/examples/example_0008_modify_turbine.py
+++ b/examples/example_0008_modify_turbine.py
@@ -1,0 +1,31 @@
+# Copyright 2019 NREL
+
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+# this file except in compliance with the License. You may obtain a copy of the
+# License at http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+
+# See read the https://floris.readthedocs.io for documentation
+
+import matplotlib.pyplot as plt
+import floris.tools as wfct
+
+# Initialize the FLORIS interface fi
+fi = wfct.floris_interface.FlorisInterface("example_input.json")
+
+fi.change_turbine([1],{'rotor_diameter':300})
+
+# # Calculate wake
+# fi.calculate_wake()
+
+# # Get horizontal plane at default height (hub-height)
+# hor_plane = fi.get_hor_plane()
+
+# # Plot and show
+# fig, ax = plt.subplots()
+# wfct.visualization.visualize_cut_plane(hor_plane, ax=ax)
+# plt.show()
--- a/examples/example_0008_modify_turbine.py
+++ b/examples/example_0008_modify_turbine.py
@@ -20,7 +20,7 @@ import floris.tools as wfct
 # Initialize the FLORIS interface fi
 fi = wfct.floris_interface.FlorisInterface("example_input.json")
 
-# Change turbine 0, 3 to have a 35 m rotor diamater
+# Change turbine 0 and 3 to have a 35 m rotor diameter
 fi.change_turbine([0,3],{'rotor_diameter':35})
 
 # Calculate wake
--- a/examples/example_0008_modify_turbine.py
+++ b/examples/example_0008_modify_turbine.py
@@ -17,15 +17,16 @@ import floris.tools as wfct
 # Initialize the FLORIS interface fi
 fi = wfct.floris_interface.FlorisInterface("example_input.json")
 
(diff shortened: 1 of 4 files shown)


paulf81:
talking with @jrannoni and @bayc this was simpler then I realized, I add the new function, and a simple example, look ok?

floris/tools/floris_interface.py
```diff
@@ -820,6 +820,42 @@ def get_farm_AEP(self, wd, ws, freq, yaw=None):
...
+            turbine_change_dict (dict): dictionary of changes to make.  All 
+                key values should be from the JSON turbine/properties set.
+                Any key values not specified will be copied from the original
+                JSON values
+        """
+
+        #TODO This may not work as intended if floris_interface not initialized with JSON
+
+        # First re-read the input file to get the baseline parameters
+        json_dict = self.floris.input_reader._parseJSON(self.input_file)
```
rafmudaf: Can you use the `input_dict` on the floris object instead of parsing again? Also, you should generally avoid using anything with a leading underscore.

floris/tools/floris_interface.py
```diff
@@ -820,6 +820,42 @@ def get_farm_AEP(self, wd, ws, freq, yaw=None):
...
+                turbine_dict["properties"][key] = turbine_change_dict[key]
+                print("Changing key %s to the following value..." % key)
+                print('...,',turbine_dict["properties"][key])
+            else:
+                print('Leaving key %s as defined in input file %s' % (key,self.input_file))
+
+        # Now go through turbine list and re-init any in turb_num_array
+        for t_idx in turb_num_array:
+            print('Updating turbine: %00d' % t_idx)
+            self.floris.farm.turbines[t_idx].__init__(turbine_dict)
```
rafmudaf: Why call the `__init__` method like this?

rafmudaf: I'm not sure what actually happens when you set the turbines attribute on Farm since thats a virtual property with a getter-method only (`@propert`). I think you should use the turbine array in `TurbineMap` instead.

paulf81: If you try the example, it seems to fly, but I admit it seems precarious.  I started out using the TurbineMap, but it was trickier to say, set the second turbine, because TurbineMap is a dict keyed off coordinates, instead of indexed array I think?  Maybe I could work some more with you on doing this correctly?  This was just my first pass...

floris/tools/floris_interface.py
paulf81: That's a good catch, I missed that floris anyway saves that, I'll change it

paulf81: Hmm, it's not clear the resultant dict is saved, it seems to be read in and passed in line 88 of floris.py, I think that's why I thought I needed to read it again...

floris/tools/floris_interface.py
paulf81: also on the __init__ question, it's because the init function builds the interpolation function, if the user changes the cp/ct tables, these functions will need to be updated...

floris/tools/floris_interface.py
rafmudaf: Good point. As an alternative, I propose that this modifies the Turbine objects in place rather than going back to the input file and starting from scratch. All properties defined in the input file exist in the Turbine class and for any that don't have an explicit setter-method we can add one. This also takes care of the situation in your TODO where FLORIS was instantiated from a dictionary instead of an input file. What do you think?

floris/tools/floris_interface.py
rafmudaf: I think we could either add an id tag to the turbines or actually keep an array in TurbineMap. That would be more explicit and avoid having two objects get out of sync.

floris/tools/floris_interface.py
paulf81: I like this idea, but I wasn't sure how to implement it.  Do you have any suggestions, or would like to take a crack at it?

floris/tools/floris_interface.py
paulf81: I'm not sure I follow, will you be at NREL next week and we can discuss live?

paulf81:
hey @rafmudaf check it now, I think it's much simpler and in line with our discussion

rafmudaf:
@bayc could you give this a second look? There are some changes to the Turbine class that you may want to see.

examples/example_0008_modify_turbine.py
```diff
@@ -0,0 +1,35 @@
...
+# This example illustrates changing the properties of some of the turbines
+# This can be used to setup farms of different turbines
+
+import matplotlib.pyplot as plt
+import floris.tools as wfct
+
+# Initialize the FLORIS interface fi
+fi = wfct.floris_interface.FlorisInterface("example_input.json")
+
+# Change turbine 0, 3 to have a 35 m rotor diamater
```
bayc: The comment could be clearer that you are changing turbines 0 and 3:

\# Change turbine 0 and 3 to have a 35 m rotor diameter

floris/tools/floris_interface.py
```diff
@@ -820,6 +820,28 @@ def get_farm_AEP(self, wd, ws, freq, yaw=None):
             AEP_sum = AEP_sum + self.get_farm_power()*freq[i]*8760
         return AEP_sum
 
+
+    def change_turbine(self, turb_num_array, turbine_change_dict):
+        """
+        Change turbine properties of given turbines
+
+        Args:
+            turb_num_array (list): list of turbine indeces to change
```
bayc: indeces -> indices

PR #81, opened 2020-04-03
move existing examples

paulf81:
Just starting...


bayc:
@paulf81 @rafmudaf I had to add a check for non-gauss models after the restructuring of the `base_gauss_model` as non-gauss models didn't have `calculate_VW` or `correction_steps` methods, where they would have previously just passed through with right flags set to false. We should decide if this is how we want to handle this. See my changes in the commit: https://github.com/NREL/floris/pull/81/commits/69c158d16e08a1bbc48f87e78b5938d47e678319.

PR #93, opened 2020-04-16
Add input file converter (AKA Preprocessor)

rafmudaf:
**Complete this sentence**

Adds an input file converter called "preprocessor". This package allows converting from an existing input file to the latest supported version. For all new version releases, this package should be updated accordingly.

None

None, this is an additional package.


rafmudaf:
The preprocessor is a command line tool separate from FLORIS (thats why I put it in the `share/` directory), and its not really meant to be scripted. So, rather than storing an old input file and scripting, I suggest adding the command line information to the documentation and README. What do you think?

PR #94, opened 2020-04-16
Expand test coverage

rafmudaf:
**Complete this sentence**

This pull request expands the test coverage within the "simulation" package a bit. It also Disables the regression tests so that we can actually see something meaningful from the automated tests instead of a perpetual red X. And the number of emails from GitHub should drop.

Tests


rafmudaf:
Yes, they aren't remove but just disabled in Github Actions. Whenever it makes sense, they can be reenabled.

PR #96, opened 2020-04-19
revert wrap functions to work with arrays

paulf81:
Hey Raf, I think you edited these functions, but they threw errors in my code when I passed in arrays, so I was hoping to revert these two if that's ok?

### DIFF
--- a/floris/utilities.py
+++ b/floris/utilities.py
@@ -149,7 +149,7 @@ def tand(angle):
     return np.tan(np.radians(angle))
 
 
-def wrap_180(angle):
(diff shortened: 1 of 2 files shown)


rafmudaf:
Ok, no problem here. Before committing this change, these two additional things should be updated:

- [ ] Docstring should clearly indicate that the arguments may be either a single value or a list and that the return type is a list
- [ ] The unit test should be updated since it is currently failing

paulf81:
@rafmudaf is this ok now?

PR #100, opened 2020-04-21
Gauss vel models

paulf81:
Adding new docstrings to gaussian wake models


rafmudaf:
@bayc could you take one more look at this? I've made some changes to all docstrings and I think they're all now consistent with the format we came up with.

paulf81:
Thanks guys, sorry there was so much left to fix..

PR #121, opened 2020-05-09
make weighting function depend on whether or not bins overlap

paulf81:
**Complete this sentence**

This fixes a few issues with energy ratio:

- Better handling of empty bins
    -- This was only an issue for small data sets but good to fix
- Passes through the bootstrap size (was forcing default)
- Correct bin weighting to be uniform in non-overlapped bins and gaussian when overlapped

energy_ratio.py

### DIFF
--- a/floris/tools/energy_ratio.py
+++ b/floris/tools/energy_ratio.py
@@ -364,9 +364,13 @@ def calculate_balanced_energy_ratio(
 
         # else:
         #     baseline_weight = np.ones_like(reference_power_baseline_wd)
-        baseline_weight = gaussian(wind_dir_array_baseline_wd,
-                                   wind_direction_bin,
-                                   wind_direction_bin_radius / 2.0)
+        # if wind_direction_bin_radius > 1.5:
+        if wind_direction_bin_p_overlap > 5.:
+            baseline_weight = gaussian(wind_dir_array_baseline_wd,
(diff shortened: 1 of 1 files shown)


floris/tools/energy_ratio.py
```diff
@@ -463,35 +472,63 @@ def calculate_balanced_energy_ratio(
...
-                p_change_bs,
-                percentiles,
-                central_estimate=p_change_array[i],
-                method='simple_percentile')
+        # print('wd',wind_direction_bin)
+        # print(ratio_base_bs)
+        # print(np.mean(ratio_base_bs))
+
+        # Trim out the nans
+        # print('size before',len(ratio_base_bs))
```
bayc, line 480: Do we know how often NANs occur? I think we can push to remove this kind of code by addressing the root cause.

paulf81: Fairly certain all nans I've seen so far are fully-empty bins, which the new code handles better.  Still, it's all a bit experimental would it be ok to leave some of this for now?

bayc: Yea, for sure, we can leave it for now, but something we should definitely work on handling differently.

paulf81: works for me!

PR #127, opened 2020-05-29
CCBlade example hotfix

nhamilto:
**Complete this sentence**
THIS PULL REQUEST __ IS __ READY TO MERGE

Updated `examples/scaling_example/get_scaled_params_with_cc_interface.py` to use dict in the output of `rotor.evaluate()` to accommodate updated function in CCBlade module.

`examples/scaling_example/get_scaled_params_with_cc_interface.py`


paulf81:
hi Nicholas, I just downloaded and tested this.  I had to pull update and then re-install CC-blade, but then can confirm it does work, and it makes sense we should be compliant with most recent version.  Nice catch!!

One quick thing, could you change the pull as to go to develop instead of master?  That way it can roll into master with the rest?

nhamilto:
Hi @paulf81 . That was an oversight. I meant for it to go into `develop` in the first place. Good catch!

PR #135, opened 2020-06-10
Small safety when only one rotor point

ebranlard:
**Complete this sentence**
THIS PULL REQUEST IS  READY TO MERGE

Small fix if you only have one rotor point (e.g. at the center), then we don't want points to be at y-R but at y.


bayc:
Thanks for the pull request @ebranlard , and apologies in the delay for addressing it. We will soon have another pull request that enables a variable number of grid points based off of the most recent develop branch, including just one grid point (with GCH turned off). Due to the impending pull request (with modifications for several other changes to the code), I'm going to close this one.

PR #141, opened 2020-06-22
Update: GCH vortex calculation, effective yaw, tuning

jrannoni:
This request has updated the following:

1. Updated vortex equations: V and W coordinates were flipped and W was missing a minus sign.
2. Updated the yaw added wake recovery.
3. Removed "yaw_recovery_alpha" parameter.
4. Slightly modified how the effective yaw is computed: take the average of the spanwise velocity across the rotor disk and compute the yaw angle that would have produced that average spanwise velocity across the disk (previously relied on rotation, but now relies on V).
5. Updated the TI tuning.
6. Change the wake combination method for V and W components to use linear superposition.
7. Updated and added examples.


floris/simulation/wake_deflection/base_velocity_deflection.py
```diff
@@ -124,40 +124,49 @@ def calculate_effective_yaw_angle(self, x_locations, y_locations,
...
             W = flow_field.w
+            Uinf = np.mean(flow_field.wind_map.grid_wind_speed)
+
+            eps = self.eps_gain * D  # Use set value
+            xLocs = x_locations - coord.x1
+            idx = np.where((np.abs(xLocs) < D / 4))
 
-            yLocs = y_locations - coord.x2
-            zLocs = z_locations - (HH)
+            yLocs = y_locations[idx] + 0.01 - coord.x2
```
bayc, line 135: @jrannoni Wondering what purpose the offsets have? Is it avoiding some error? And if so, should we check for that error?

PR #144, opened 2020-06-29
Feature/yaw opt normalization

bayc:
**Complete this sentence**

Adds normalization to the yaw optimizations in the scipy folder.

None.

`floris/tools/scipy`

Improves the optimization timing and performance.

None.


floris/tools/optimization/scipy/yaw_wind_rose.py
```diff
@@ -171,8 +171,57 @@ def __init__(
             unc_options=unc_options,
         )
 
+        self._get_initial_farm_power()
```
ejsimley, line 174: I think this makes more sense in reinitialize_opt_wind_rose so it gets called every time the object is updated.

floris/tools/optimization/scipy/yaw_wind_rose_parallel.py
```diff
@@ -513,7 +515,10 @@ def optimize(self):
         with MPIPoolExecutor() as executor:
             if self.ti is None:
                 for df_opt_one in executor.map(
-                    self._optimize_one_case, self.ws.values, self.wd.values
+                    self._optimize_one_case,
+                    self.ws.values,
+                    self.wd.values,
+                    self.df_base.power_baseline.values,
```
ejsimley, line 521: the super init is already calling self._get_initial_farm_power(), so maybe it makes sense for the parallel class to use the initial power values that were calculated there for consistency.

ejsimley: After thinking about it, if this were the case the parallel class would not be able to take advantage of parallelizing when get_initial_farm_power() is called in the parent class.

ejsimley:
I made some changes so that a) _get_initial_farm_power is called in reinitialize_opt_wind_rose instead of init for YawOptimizationWindRose (so that it gets recomputed every time the parameters get reinitialized), and b) when a YawOptimizationWindRoseParallel object gets initialized the super init does not call _get_initial_farm_power (since initial farm powers are calculated differently in the parallel version). @bayc do these changes seem reasonable to you?

ejsimley:
Hey Chris, I just made some minor updates to the code. I think it is ready to merge now.

floris/tools/optimization/scipy/yaw.py
```diff
@@ -289,10 +299,26 @@ def reinitialize_opt(
...
+            (
+                self._norm(
+                    self.bnds[i][0], self.minimum_yaw_angle, self.maximum_yaw_angle
+                ),
+                self._norm(
+                    self.bnds[i][1], self.minimum_yaw_angle, self.maximum_yaw_angle
+                ),
+            )
+            for i in range(self.nturbs)
+        ]
```
ejsimley, line 321: I changed how the bounds are set and normalized if `bnds` is defined in the arguments. Before, `bnds_norm` was getting set to (0,1) for each turbine, even if `bnds` specified different bounds for each turbine.

floris/tools/optimization/scipy/yaw.py
```diff
@@ -385,10 +418,6 @@ def minimum_yaw_angle(self):
 
     @minimum_yaw_angle.setter
     def minimum_yaw_angle(self, value):
-        if not hasattr(self, "maximum_yaw_angle"):
-            self._set_opt_bounds(value, 25.0)
-        else:
-            self._set_opt_bounds(value, self.maximum_yaw_angle)
```
ejsimley, line 391: Removed the calls to `_set_opt_bounds` because this would overwrite bounds that were previously set by the optional `bnds` argument. @bayc please check if this is ok! There might be another way to handle this.

floris/tools/optimization/scipy/yaw_wind_rose.py
```diff
@@ -349,10 +407,26 @@ def reinitialize_opt_wind_rose(
...
+            (
+                self._norm(
+                    self.bnds[i][0], self.minimum_yaw_angle, self.maximum_yaw_angle
+                ),
+                self._norm(
+                    self.bnds[i][1], self.minimum_yaw_angle, self.maximum_yaw_angle
+                ),
+            )
+            for i in range(self.nturbs)
+        ]
```
ejsimley, line 429: Similarly in `yaw_opt_wind_rose`, I changed how the bounds are set and normalized if `bnds` is defined in the arguments. Before, `bnds_norm` was getting set to (0,1) for each turbine, even if `bnds` specified different bounds for each turbine.

bayc: Looks good, thanks!

floris/tools/optimization/scipy/yaw.py
bayc: I think this works.

PR #154, opened 2020-08-03
Feature/performance

paulf81:
**Complete this sentence**
THIS PULL REQUEST __IS NOT__ READY TO MERGE

This branch is to be used in assessing the overall speed performance of FLORIS, it includes a set of new examples to check the speed of certain functions, and methods to assess the relative change in performance.  Running with the recalculate_baseline flag on resets the baseline result compared to, to the current setting


floris/utilities.py
```diff
@@ -168,6 +177,41 @@ def tand(angle):
     return np.tan(np.radians(angle))
 
 
+# @jit(nopython=True)
```
rafmudaf, line 180: Do we want to keep these two functions (`wrap_180` and `wrap_180_fast`) if we aren't doing jit?

floris/simulation/wake_velocity/gaussianModels/gauss_legacy.py
```diff
@@ -17,6 +17,95 @@
...
+    sigma *= 0.501 * D * np.sqrt(Ct / 2.0)
+    sigma += (diff2 / diff1) * sigma_0
+    return sigma
+
+
+# @njit
+def compute_r_C(
+    veer, sigma_y, sigma_z, y_locations, z_locations, delta, x2, yaw, D, HH, Ct
+):
+    """Computes r and C.
```
rafmudaf, line 74: What are `r` and `C`?

floris/utilities.py
```diff
@@ -90,6 +92,10 @@ def rotate_on_x3(self, theta, center_of_rotation=None):
         )
         self.x3prime = self.x3
 
+    @property
+    def coords(self):
```
rafmudaf, line 96: Could this be renamed? The `Vec3` class is often used for coordinates, but its meant to be a general purpose class for any 3-component vector. For example, this could represent the velocity components in a field.

Another option is to implement to `TODO` in the docstring.

rafmudaf:
Is this meant to be merged as is? If so, there's a bit of cleanup that should occur before merging.

- [ ] General clean up of commented code that was replaced with other functionality - without the context of this performance improvement effort, commented code blocks will quickly become noise.
- [ ] Resolve merge conflicts
- [ ] Enable tests
- [ ] Ensure tests pass
- [ ] We did some validation along the way... is there any more robust proof that we haven't changed the results?

Alternatively, we could try to take a subset of the changes and create new commits, but I'm not sure how difficult that would be. @RHammond2 @paulf81 @bayc do you have any insight there?

floris/utilities.py
paulf81: I removed them, good catch

floris/utilities.py
paulf81: This one I might leave up to you, sorry, don't have a strong opinion...

paulf81:
I can see the logic of a clean commit we can start from develop and apply the subset of changes we intend to keep by hand, it might go something like:

1). Bring over 1-2 of the new performance testing examples so we can both confirm speed up and minimal change
   - Note GCH results will change and that's ok
2) Then start copying over the changes while re-running the code?

A little time consuming, but maybe faster than cleaning up conflicts and cleaning and back testing?

floris/simulation/wake_velocity/gaussianModels/gauss_legacy.py
RHammond2: I know I made this change, but I don't know, but those were the parameter calculations I was pulling out of the class. This piece, I believe, offered a decent amount of speedup because of how much it's called.

floris/utilities.py
RHammond2: I did this one also, it could easily be renamed. I assumed it was coordinates because it's frequently used for them; it was a convenience feature instead of calling all the items in a number of spots throughout the code.

RHammond2:
@rafmudaf I'm happy to go through with someone who knows a little bit more about what code I was modifying and fill in the incomplete docstrings, update function names to be more explanatory/appropriate to their use case, and anything else. Also, if this functionality is coming from various papers, I could also use those to go through and do a first pass. There may also be some commented out code from when I was pulling functionality out of the classes; I know I did some of it along the way, but I likely still left spots. Happy to take a first pass at cleaning those spots up as well (some commented out functionality did already exist, so this could be an opportunity to also address that).

Due to their not being tests for everything I can't speak to every change, but I did bring my results in line with the expected results of each case when doing validation along the way.

PR #156, opened 2020-09-16
Correcting balanced energy ratio weights

ejsimley:
This is a minor change to correct the weights used to balance wind speed distributions for baseline and controlled data in the `energy_ratio` and `energy_ratio_ws` methods in `energy_ratio.py`. The new weights should balance between baseline and controlled cases while making the relative weights placed on different wind speed bins based on the combined (baseline + controlled) wind speed distribution.


paulf81:
Looks great!  I say we merge, objections?

PR #166, opened 2020-11-11
Hotfix: Add lower bound to Ct interpolation

bayc:
**Complete this sentence**

Adds a lower bound to the Ct interpolation, preventing zero or negative values of Ct. Also bounds the calculation of `C` in the `gauss_legacy` model to prevent complex wake deficits.

None.

`turbine.py`
`gauss_legacy.py`

The Ct error can occur at high wind speeds due to the interpolation used for determining a turbine's Ct value, especially if the thrust table in the input file is not defined above the cut-out wind speed. Complex values of C have occurred in the extreme near wake of turbines in the `gauss_legacy` model due to near wake approximations. The formatting of the return statement was done by the pre-commit actions.

None.


rafmudaf (commented):
Is this also a possible problem for Cp?

bayc:
@rafmudaf It shouldn't be an issue, but the code would be more robust if we added similar limits to Cp. See the additional commit.

PR #170, opened 2020-11-18
Bugfix: specify shading for visualization and masking for zero wind speeds

paulf81:
**Complete this sentence**
Specifies color mesh shading rather than default to get rid of warning message


paulf81 (commented):
This passes all checks I made

bayc (approved):

PR #210, opened 2021-06-17
Remove "gauss_legacy" wake model

paulf81:
Removing the `gauss_legacy` wake model from FLORIS.


rafmudaf:
@paulf81 I see there's a new folder `historic_versions` containing a couple of Gaussian wake models including `gauss_legacy`. These will always be available in older versions of FLORIS via GitHub and git. Is there a good reason to keep these in the source code? My thought is that this doesn't add much in functionality but it does takeaway clarity in the repository. For example, whats the difference in these two models of the same name:
- https://github.com/NREL/floris/blob/develop/floris/simulation/wake_velocity/gaussianModels/gauss.py
- https://github.com/NREL/floris/blob/develop/floris/simulation/wake_velocity/gaussianModels/historic_versions/gauss.py

rafmudaf (approved):

paulf81:
@rafmudaf I have had second thoughts on this branch and might close it, we can discuss

rafmudaf:
Ok I'll make it a draft PR for now

rafmudaf:
@paulf81 Is it ok to close this pull request without merging?

PR #259, opened 2021-08-19
Use the Cp/Ct table from example input in tests

rafmudaf:
This pull request updates the regression tests so that they use the power and thrust coefficient look up tables from the example input file. This makes the test results match the examples and allows for simpler testing.

For reference, here are the curves compared.
Cp
Ct

None

Regression tests only


tests/conftest.py
```diff
@@ -54,93 +54,153 @@ def __init__(self):
...
+                        0.07448356,
+                        0.07050731,
+                        0.06684119,
+                        0.06345518,
+                        0.06032267,
+                        0.05741999,
+                        0.05472609
                     ],
                     "wind_speed": [
-                        0.0,
```
bayc, line 116: Is there a reason to change this from 0.0 to 2.0? I feel like 0.0 covers more cases that people may try, or at least would cover more cases for the interpolation.

bayc (approved):

PR #303, opened 2022-01-27
Support heterogeneous wind speed inputs

bayc:
This pull request adds heterogeneous inflow capabilities and an example to FLORIS v3. This includes both 2- and 3-dimensional flow definitions. The heterogeneous flows are defined as an array of speed ups which are multipliers of the ambient wind speed. The dimensions of this speed up array are the number of wind directions by the number of defined speed up points. In addition to the speed ups, the locations of the speed ups are required (for 2d, x and y locations, and for 3d, x, y, and z locations). Note: the example uses the `gch.yaml` input file, as it won't work with the `cc.yaml` input file until plotting is implemented for the CC model.

#282 

Changes are limited to `flow_field.py` and `floris_interface.py`.

If FLORIS is asked to compute a speed up outside of the defined area of the heterogeneous interpolant map, the speed up that is returned is equal to 1.0, so the user must make sure that they have defined inflows over the full area that they want heterogeneous flow for.

Also, the heterogeneous map is stored on the `FlorisInterface` object as well as the `FlowField` object. This is because the map is used in the flow field, when the flow field is initialized with the specified grid. However, upon reinitialization (e.g. a `fi.reinitialize()` call), the map is lost from the flow field object, and thus requires re-assigning from the stored version on the `FlorisInterface` object. The reason to store the map on the `FlowField` object was made to prevent having to pass the map through the calculation call each time (starting in `FlorisInterface`, the map would have to be passed to every call of a solver, which would then pass it to `initialize_flow_field`).

The example produces the following figures. The 1st figure shows a 2d inflow definition where the flow varies in the x direction only. The 2nd figure shows a 3d inflow definition where the flow varies only in the z direction.

2d_het_flow

3d_het_flow

### DIFF
--- a/src/floris/simulation/flow_field.py
+++ b/src/floris/simulation/flow_field.py
@@ -46,14 +46,14 @@ class FlowField(FromDictMixin):
     v: NDArrayFloat = field(init=False, default=np.array([]))
     w: NDArrayFloat = field(init=False, default=np.array([]))
     speed_ups: NDArrayFloat = field(init=False, default=np.array([]))
-    het_map: list = field(init=False, default=[])
+    het_map: list = field(init=False, default=None)
 
     turbulence_intensity_field: NDArrayFloat = field(init=False, default=np.array([]))
 
     def __attrs_post_init__(self) -> None:
         # If no hetergeneous inflow defined, then set all speeds ups to 1.0
         if np.size(self.speed_ups) == 0:
-            self.speed_ups = np.ones_like(self.wind_directions)
+            self.speed_ups = np.ones_like(self.wind_directions)[:, None, None, None, None]
 
     @wind_speeds.validator
     def wind_speeds_validator(self, instance: attrs.Attribute, value: NDArrayFloat) -> None:
--- a/src/floris/simulation/flow_field.py
+++ b/src/floris/simulation/flow_field.py
@@ -45,15 +45,9 @@ class FlowField(FromDictMixin):
     u: NDArrayFloat = field(init=False, default=np.array([]))
     v: NDArrayFloat = field(init=False, default=np.array([]))
     w: NDArrayFloat = field(init=False, default=np.array([]))
-    speed_ups: NDArrayFloat = field(init=False, default=np.array([]))
     het_map: list = field(init=False, default=None)
 
     turbulence_intensity_field: NDArrayFloat = field(init=False, default=np.array([]))
-
-    def __attrs_post_init__(self) -> None:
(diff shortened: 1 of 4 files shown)


src/floris/simulation/flow_field.py
```diff
@@ -89,3 +102,18 @@ def finalize(self, unsorted_indices):
         self.u = np.take_along_axis(self.u, unsorted_indices, axis=2)
         self.v = np.take_along_axis(self.v, unsorted_indices, axis=2)
         self.w = np.take_along_axis(self.w, unsorted_indices, axis=2)
+
+    def calculate_speed_ups(self, het_map, x, y, z=None):
```
paulf81, line 106: A couple of questions about this function: 1) How often is called?  2) It has a list comprehension, is that fast enough?

paulf81:
This looks really good Chris.  I pulled the branch down and confirm the examples worked well on my mac.  I put a specific comment on one of the functions but then I think more generally my concerns are just on speed.  Maybe these are simple to answer: 
1) I confirmed the older examples all work fine, and I'm assuming that since they don't pass in a het_map, there is no speed impact on these?  Could you confirm?
2) I just wanted to see what the speed impact is when this is applied?  If it's minimal all good, but if bigger maybe some pre-allocation or some numpy to replace the list comprehension (or maybe my understanding of python is out of date and that's fine)

Otherwise super super cool!!! Thanks @bayc !

src/floris/simulation/flow_field.py
bayc: Good question, so this function gets called once per solve execution, at the beginning when the flow field is initialized based on the specified grid. And for a "full load" calculation of 100 turbines, 72 wind directions, and 20 wind speeds, the `calculate_speed_ups` function call takes approx. 0.01 seconds. So in my opinion, it is fast enough.

bayc:
> * I confirmed the older examples all work fine, and I'm assuming that since they don't pass in a het_map, there is no speed impact on these?  Could you confirm?

That's correct, there is not speed impact on the case that there is no `het_map` supplied. In this case, the speed up array is just set to 1.0 for the same size as the flow field.

> * I just wanted to see what the speed impact is when this is applied?  If it's minimal all good, but if bigger maybe some pre-allocation or some numpy to replace the list comprehension (or maybe my understanding of python is out of date and that's fine)

I think I addressed this in my reply to your comment above, but in brief, the function takes about 0.01 seconds for 100 turbines, 72 wind directions, and 20 wind speeds.

rafmudaf (approved):

PR #306, opened 2022-02-01
Add yaw optimization with SciPy

Bartdoekemeijer:
This PR reintroduces the SciPy yaw optimizer class, now built on top of the new yaw optimization base class. It thereby can also leverage the functions like exploiting symmetry in the wind farm and automated removal of downstream turbines. Upon implementing, I have moved a couple of lines of code from the Serial-Refine optimizer class to the general base yaw optimization class, since all optimizers (including SciPy) will likely use them.

Practically, the SciPy optimizer deals with the optimizations by going through all wind directions one by one. This therefore does not exploit the vectorization of FLORIS v3. It does still benefit from other performance and code improvements.

N/A.

Yaw optimization module.

N/A

I have added an example, `10_compare_yaw_optimizers.py`. Running that script yields:
```
Performing optimizations with SciPy...
  NIT    FC           OBJFUN            GNORM
    1     3    -1.000000E+00     7.944927E-03
Optimization terminated successfully    (Exit mode 0)
...
Optimization terminated successfully    (Exit mode 0)
            Current function value: -1.0
            Iterations: 1
            Function evaluations: 3
            Gradient evaluations: 1

Performing optimizations with Serial Refine...
[Serial Refine] Processing pass=0, turbine_depth=0 (0.0 %)
[Serial Refine] Processing pass=0, turbine_depth=1 (16.7 %)
[Serial Refine] Processing pass=0, turbine_depth=2 (33.3 %)
[Serial Refine] Processing pass=1, turbine_depth=0 (50.0 %)
[Serial Refine] Processing pass=1, turbine_depth=1 (66.7 %)
[Serial Refine] Processing pass=1, turbine_depth=2 (83.3 %)

 Time spent, Serial Refine: 0.30 s.
 Time spent, SciPy (SLSQP): 18.57 s.
```

t0
t1
t2
power


paulf81:
This looks great Bart, thanks for putting together!  Just for testing I checked out your branch and ran example 10, which worked fine on my mac and had final plots like you show above and timings:
 Time spent, Serial Refine: 0.19 s.
 Time spent, SciPy (SLSQP): 12.22 s.

examples/10_compare_yaw_optimizers.py
```diff
@@ -0,0 +1,86 @@
...
+# Load the default example floris object
+fi = FlorisInterface("inputs/gch.yaml") # GCH model matched to the default "legacy_gauss" of V2
+# fi = FlorisInterface("inputs/cc.yaml") # New CumulativeCurl model
+
+# Reinitialize as a 3-turbine farm with range of WDs and 1 WS
+D = 126.0 # Rotor diameter for the NREL 5 MW
+fi.reinitialize(
+    layout=[[0.0, 5 * D, 10 * D], [0.0, 0.0, 0.0]],
+    wind_directions=np.arange(0.0, 360.0, 3.0), 
+    wind_speeds=[8.0],
```
bayc, line 42: @Bartdoekemeijer does the `YawOptimizationScipy` class handle a multiple wind speed scenario? From looking at the code, I can see where it loops over the wind directions individually, but what happens if the user defines multiple wind speeds as well? Seems like a nested loop over `wd_array` and a `ws_array` would be the most obvious solution, but perhaps not the most efficient.

Bartdoekemeijer: Good point, it doesn't support multiple wind speeds currently. I wasn't sure if we were going to change the way we insert wind directions/wind speeds on the `floris` side, so I've left it at that. The base yaw optimizer code currently has [these lines](https://github.com/Bartdoekemeijer/floris/blob/cf3d7c413bd72f3b22b4512b0f2118add1fcc2d9/src/floris/tools/optimization/yaw_optimization/yaw_optimization_base.py#L142) in the initialization step:

```
        # Check floris options
        if self.fi.floris.flow_field.n_wind_speeds > 1:
            raise NotImplementedError(
                "Optimizer currently does not support more than one wind" +
                " speed. Please assign FLORIS a single wind speed."
            )
```

It should be pretty easy to expand over a range of wind speeds, but there are some subtleties depending on whether we assume `n_wind_directions == n_wind_speeds` (timeseries playthrough, kind of style), or a grid style like currently implemented in `floris`. Let me know what you think would be the best way to proceed!

bayc: Hey @Bartdoekemeijer , getting back to this, I think keeping it in the grid style currently implemented would be best for now. How difficult would that be to support?

Bartdoekemeijer: Should not be too difficult. I'll take a crack at that and let you know if there are any potential barriers!

Bartdoekemeijer:
@bayc  I've added compatibility with multiple wind speeds as specified in the floris object. Feel free to give it a spin. It works for me for all the examples. Do we want to update or add an example to demonstrate this use case, @paulf81 ?

Also, speed is still great for Serial Refine. SciPy quickly becomes too much to handle (300x slower than SR) for the 3-turbine example case when considering a wind speed range of 4 through 20 m/s in steps of 1 m/s on my laptop.

paulf81:
I think a separate example for this would be welcome if you have the time Bart

Bartdoekemeijer:
Just pushed up an example, @paulf81. I also found something strange in the results, which is most likely because of numerical inaccuracies in the cp curve. Will open a separate issue for that. Anyhow, please check out the example and see if you like it!

Issue #287, opened 2022-01-25
Include uncertainty options in the input file and fi.reinitialize()


rafmudaf:

Bartdoekemeijer:
This has been pulled in with #308. We decided not to include uncertainty in the input file, but rather design a wrapper around the `FlorisInterface` object that adds uncertainty. That feature is called `UncertaintyInterface`. Note that this is only one method for including uncertainty with FLORIS, and therefore has not been made part of the core code.

PR #308, opened 2022-02-01
Feature/include unc

Bartdoekemeijer:
**Pull request status**
This PR is ready to be merged.

This PR reintroduces the `include_unc` option to calculate the wind farm power productions with a probability distribution for the inflow wind directions and turbine yaw angles.

Issue #287 

Core floris usage, simulation.

The `include_unc` functionality in `v2.4` looped over different wind directions and wind speeds to calculate the floris solution. If the probability distribution on the wind direction was discretized over 27 points, and the probability distribution for the yaw angle was discretized over 11 points, we would call the `calculate_wake` function 27x11=297 times vs. a single time for the nominal case without uncertainty. Practically, for `v3`, I calculate all the possible combinations of wind directions, wind speeds and yaw angles a priori and then evaluate them in vectorized style using a single `calculate_farm_power` call. I also check for uniqueness so that we are not evaluating identical situations multiple times (which yields about a 50% reduction in number of cases to evaluate in my simple example case). All of this is very favorable for the computational efficiency of floris.

A number of things we still need to do before we can merge this PR:

- Can we lift the `include_unc` functionality in `calculate_farm_power` to a more generic location, so that it will also work for `calculate_turbine_power`? This question is probably best answered by @rafmudaf and @bayc.
- Allow the user to define `include_unc` and the probability distributions in the input `.yaml` file, rather than when calling the calculate_farm_power function. This question is probably best answered by @rafmudaf and @bayc.
- Add documentation and expand the example case.
- Validate solutions in comparison to `v2.4`.

I have added an example called `11_calculate_farm_power_with_uncertainty.py`. This script compares the `floris` wind farm power results with and without uncertainty. It produces the following figure:
image


Bartdoekemeijer:
OK, I completely lifted the uncertainty part out of `FlorisInterface` and created a separate class for it.
- I added the `UncertaintyInterface` class in `uncertainty_interface.py`. This class can be either initialized in the exact same way as the regular `FlorisInterface` class (by providing the `configuration` dictionary or the path to it), or just be passed the already instantiated `FlorisInterface`. It additionally takes `unc_options` or `unc_pmfs`, which specify the probability distributions of `wind_direction` and `yaw_angle`.
- `UncertaintyInterface` is interacted with in the same way `FlorisInterface` is interacted with. It includes a minimum set of essential functions like `reinitialize`, `calculate_wake`, `calculate_turbine_powers`, `calculate_farm_power` to allow basic operations and yaw optimizations. It can be expanded to include the calculation and visualization of probablistic flow fields, but currently does not have such features.
- To enable compatibility with the yaw optimizer, I had to generalize how we copy a `FlorisInterface` object. I added a public function called `copy` in `FlorisInterface` and in `UncertaintyInterface` to address that. This also makes it easy for an user to copy a `FlorisInterface` object just by `fi_copy = fi.copy()`.

Now, if one wants to optimize the yaw angles with uncertainty, all you do is, for example:
```
fi_unc = UncertaintyInterface("inputs/gch.yaml")

fi.reinitialize(
    layout=[[0.0, 600, 1200], [0, 600, 1200]],
    wind_directions=np.arange(0.0, 360.0, 3.0), 
    wind_speeds=[8.0],
)
fi_unc.reinitialize(
    layout=[[0.0, 600, 1200], [0, 600, 1200]],
    wind_directions=np.arange(0.0, 360.0, 3.0), 
    wind_speeds=[8.0],
)

yaw_opt = YawOptimizationSR(fi)
df_opt = yaw_opt.optimize()

yaw_opt_unc = YawOptimizationSR(fi_unc)
df_opt_unc = yaw_opt_unc.optimize()

print("Optimization results (no uncertainty):\n {}\n".format(df_opt))
print("Optimization results (with uncertainty):\n {}".format(df_opt_unc))

```

I haven't changed the way uncertainty is actually calculated. @paulf81, if you like the way `UncertaintyInterface` works, I can look at that next.

paulf81:
I really like this @Bartdoekemeijer !  I think it really nicely organizes the nominal/uncertain calculations and lets us keep the base code more trim.  I downloaded and read the new files, and also ran the example which went fine.

My main comment, should we have a flag for when we use wd uncertainty that determines whether:

a) Wind direction variation is matched by yaw angle variations (such that we assume the turbines are aligned to the flow but that data point at a particular condition is a mix of nearby wind directions)
b) Yaw stays fixed to the nominal wind direction (like in robust yaw control design, to plan for the cases especially near big wrap points)

What do you think of this?  @ejsimley also curious your opinion

Bartdoekemeijer:
@paulf81 this is something easy to implement. So essentially the two situations you describe are:

a) The relative yaw angle is fixed and always has the nominal value (e.g., 0 deg) when evaluating uncertainty. Evaluating wind direction uncertainty like this would essentially come down to a Gaussian smoothing of FLORIS solutions over the wind directions. This can also be really fast, since it would not require any additional calculations compared to the non-uncertainty FLORIS evaluation.

b) The absolute yaw angle (in compass reference frame) is fixed when evaluating uncertainty, so when evaluating a probablistic wind direction 3 deg higher than nominal, the relative yaw angle is 3 deg lower than its nominal value. This requires many FLORIS evaluations but theoretically may be more often correct.

I think having both options is very useful, even if it just is for having an option to include uncertainty without increasing the computational cost.

A flag could be, `fix_yaw_in_relative_frame`, which toggles between the two options, (a) corresponding to `fix_yaw_in_relative_frame=True` and (b) corresponding to `fix_yaw_in_relative_frame=False`.

paulf81:
> @paulf81 this is something easy to implement. So essentially the two situations you describe are:
> 
> a) The relative yaw angle is fixed and always has the nominal value (e.g., 0 deg) when evaluating uncertainty. Evaluating wind direction uncertainty like this would essentially come down to a Gaussian smoothing of FLORIS solutions over the wind directions. This can also be really fast, since it would not require any additional calculations compared to the non-uncertainty FLORIS evaluation.
> 
> b) The absolute yaw angle (in compass reference frame) is fixed when evaluating uncertainty, so when evaluating a probablistic wind direction 3 deg higher than nominal, the relative yaw angle is 3 deg lower than its nominal value. This requires many FLORIS evaluations but theoretically may be more often correct.
> 
> I think having both options is very useful, even if it just is for having an option to include uncertainty without increasing the computational cost.
> 
> A flag could be, `fix_yaw_in_relative_frame`, which toggles between the two options, (a) corresponding to `fix_yaw_in_relative_frame=True` and (b) corresponding to `fix_yaw_in_relative_frame=False`.

Yes, I think that is right.  I think (a) is what you should use when comparing FLORIS to SCADA data and the data consists of 10 minute averages.  This represents the probable distribution of points which get assigned to a particular wind direction bin.  Then (b) is what you should use when designing a wake steering controller (or predicting uplift), as it representing likely distributions of tracking error, wind direction measurement error.

ejsimley:
@Bartdoekemeijer and @paulf81 I'm not very familiar with v3 yet, so don't think I can provide the best code review (although everything looked correct and nicely organized reading through the files!)

But I totally agree that it would be nice to include the two options with a) fixed yaw misalignment, and b) fixed absolute yaw position (option b being what we'd use for robust yaw offset optimization). Also, great point Paul that option a) probably makes more sense when comparing to 10-minute SCADA data. 

Another idea for the flag name is `fix_yaw_offset` or `fix_yaw_misalignment`, but I know different people use different terms to describe absolute vs. relative yaw position. From my perspective, setting the default value to False would make more sense since I use FLORIS for yaw optimizations a lot, but depending on what you think the most popular use case will be, maybe True is a better choice?

Bartdoekemeijer:
Thanks Eric, Paul! Just pushed up the latest with:

- `yaw_std` removed because it wasn't used very much and added complexity.
- Added option `fix_yaw_in_relative_frame` 
- Changed default option to `False`, making the default valid for robust wake steering optimization
- Updated docstring

Could you @paulf81 please check the latest version? After that, I think it's ready to be pulled in!

paulf81:
hi @Bartdoekemeijer  this looks really good to me.  I wanted to fully test my understanding by making an example of yaw control design as a new example.  I'm not sure it shows here but if you receive it could you tell me if that all looks right?  Maybe we should exaggerate the effect?  I tried to make the yaw optimizer use neg/pos yaw but was getting errors so thought I might pause here and wait for you to come back so we can chat

Bartdoekemeijer:
@paulf81 the following seems to work fine for me:

```
import numpy as np
from floris.tools import FlorisInterface, UncertaintyInterface
from floris.tools.optimization.yaw_optimization.yaw_optimizer_sr import (
    YawOptimizationSR,
)

fi = FlorisInterface("inputs/gch.yaml") # GCH model matched to the default "legacy_gauss" of V2
D = 126.0 # Rotor diameter for the NREL 5 MW
fi.reinitialize(
    layout=[[0.0, 5 * D, 10 * D], [0.0, 0.0, 0.0]],
    wind_directions=np.arange(0.0, 360.0, 3.0), 
    wind_speeds=[8.0],
    turbulence_intensity=0.06,
)

# Initialize uncertainty FLORIS object as copy of nominal object
fi_unc = UncertaintyInterface(fi)  

# Initialize optimizer object and run deterministic optimization
yaw_opt = YawOptimizationSR(
    fi,
    minimum_yaw_angle=-25.0,
    maximum_yaw_angle=25.0,
)
df_opt = yaw_opt.optimize()

# Initialize optimizer object and run stochastic optimization
yaw_opt_unc = YawOptimizationSR(
    fi_unc,
    minimum_yaw_angle=-25.0,
    maximum_yaw_angle=25.0,
)
df_opt_unc = yaw_opt_unc.optimize()

# Split out the turbine results
yaw_angles_opt_deterministic = np.vstack(df_opt["yaw_angles_opt"])
yaw_angles_opt_stochastic = np.vstack(df_opt_unc["yaw_angles_opt"])

# Show the results
fig, ax = plt.subplots(4, 1, sharex=True, figsize=(8,8))

# Yaw results
for ti in range(3):
    ax[ti].plot(
        df_opt["wind_direction"],
        yaw_angles_opt_deterministic[:, ti],
        label="Deterministic",
    )
    ax[ti].plot(
        df_opt_unc["wind_direction"],
        yaw_angles_opt_stochastic[:, ti],
        label="Stochastic"
    )
    ax[ti].set_ylabel("Optimal yaw \n angle T{:d} (deg)".format(ti))
    ax[ti].grid(True)
    ax[ti].legend()

# Power results
ax[3].plot(
    df_opt["wind_direction"],
    df_opt["farm_power_baseline"],
    "-",
    color="tab:blue",
    label="Baseline Farm Power (deterministic)"
)
ax[3].plot(
    df_opt["wind_direction"],
    df_opt["farm_power_opt"],
    "--",
    color="tab:blue",
    label="Optimized Farm Power (deterministic)"
)
ax[3].plot(
    df_opt_unc["wind_direction"],
    df_opt_unc["farm_power_baseline"],
    "-",
    color="tab:orange",
    label="Baseline Farm Power (stochastic)"
)
ax[3].plot(
    df_opt_unc["wind_direction"],
    df_opt_unc["farm_power_opt"],
    "--",
    color="tab:orange",
    label="Optimized Farm Power (stochastic)"
)
ax[3].set_ylabel("Power (W)")
ax[3].set_xlabel("Wind Direction (deg)")
ax[3].legend()
ax[3].grid(True)

plt.show()
```

and produces:
image

paulf81:
This is strange!!  I copied in your code and it works, but mine still doesn't and I can't spot the difference, would you mind running example 08b and 08c in the include_unc_add_yaw_example branch?

Bartdoekemeijer:
@paulf81 Found the issue! The yaw optimizer didn't handle `minimum_yaw_angle` being an `int` well. In my code, they were floats so the issue did not arise. I fixed the compatibility with ints with the latest commit. Your examples should work now!

paulf81:
> @paulf81 Found the issue! The yaw optimizer didn't handle `minimum_yaw_angle` being an `int` well. In my code, they were floats so the issue did not arise. I fixed the compatibility with ints with the latest commit. Your examples should work now!

subtle!  nice catch!!

paulf81:
I think this pull is good to go now from my perspective, my examples were only to test my understanding and don't need to be pulled in, going to delete my branch of this branch now and close the pull request

Bartdoekemeijer:
@paulf81 Oops, I had already pulled in your examples/PR. I reverted that now. Is there anything else we want to do in this PR? In terms of adding examples that use the uncertainty functionality, I'd say let's pull in this feature now and we can maybe add a robust optimization example later. To me, this PR seems ready to be merged in. @paulf81 if you agree, I'll get this squashed and merged.

Bart

paulf81:
Yes I agree, this is looking good, I say squash and merge!

PR #317, opened 2022-02-15
Increase performance in Jensen-Jimenez models with Numexpr

rafmudaf:
This pull request changes the Jensen and Jimenez wake models to use Numexpr in their mathematical operations resulting in about a 25% speedup.


paulf81:
This is really neat

PR #323, opened 2022-02-17
Simplify rotor plotting tool and add to example

rafmudaf:
Adds an example for using the rotor-grid plotting function. Also, simplifies this function to make it more general and allow for expansion in the future.

#285


paulf81:
This is nice, I pulled it down and tested and works great!  Since in this example 1, we don't use fi to do any power calculations, maybe it's consistent to just call it fi_viz from the start?

Issue #224, opened 2021-06-29
Feature request: Allowing complex topology


Bartdoekemeijer:
**Is your feature request related to a problem? Please describe.**
Whenever a turbine has a different relative rotor position (vertically) due to a complex topology, it feels like I am hacking the FLORIS object. Namely, what I end up doing is incrementing each turbine's hub height with the ground height w.r.t. sea level or other reference value. Something along these lines:

```
# Define complex topology
layout_z = np.array(...)

# Define a zero point, e.g., min. of all turbines (z0)
if normalize_terrain:
    z0 = np.min(layout_z)
else:
    z0 = 0.

# Shift all the turbine hub heights accordingly
for ti in range(len(fi.layout_x)):
    hub_height_z0 = fi.floris.farm.turbines[ti].hub_height
    hub_height_z = layout_z[ti] - z0 + hub_height_z0
    fi.change_turbine([ti], {'hub_height': hub_height_z})
```

The problem is that this is a somewhat roundabout process, that we lose track of what `specified_wind_height` corresponds to if its value was set to `-1`, and that `layout_z` is not included in the .json file and therefore needs to be defined somewhere externally.

**Describe the solution you'd like**
Similar to in #223, I would like `layout_z` to be defined in the .json file. If there is a single turbine type and `layout_z` is an array with all the same value, then the definition of `specified_wind_height` is not ambiguous. In any other situation, it must force the user to specify its value manually.

**Describe alternatives you've considered**
I now have a practice where I create a class for these types of wind farms, wherein I contain all these operations. That way, I can still simply call a wind farm using a single command and get a floris object in return.

**Additional context**
This issue is somewhat related to Issue #223.

bayc:
I believe we can address this in v3.1, and want to add a comment here to note that. Thanks Bart!

rafmudaf:
@bayc I've created a v3.1 milestone and added this to it

misi9170:
Closing as we now simply use wind field heterogeneity to model complex terrain/topology (see also #851). If someone is interested in contributing a complex terrain model to FLORIS, I'd be happy to support!

Issue #225, opened 2021-06-29
Feature request: Turbine identifiers


Bartdoekemeijer:
**Is your feature request related to a problem? Please describe.**
Often, a FLORIS model is an abstraction of a real wind farm. Within those wind farms, turbines have typical identifiers such as A1-A9, B1-B5 and C14-C22, rather than integers from `0` to `num_turbs-1`. I often find myself generating a `mapping.csv` file that shows what turbine number in FLORIS correspond to what turbine in the actual farm. I believe that being able to keep track of the turbine names and generate visualizations with the right turbine names would be a valuable addition to floris.

**Describe the solution you'd like**
An (optional) additional input in the floris input .jsons to specify the turbine names. These turbine identifiers are never used in the actual core code, but rather only used in visualizations and as a useful way for the user to keep track of the turbines.

**Describe alternatives you've considered**
I now have a practice where I create a wrapper class for floris wherein I save a variable called "turbine_names" that can be returned to the user upon request.

**Additional context**
This is a minor suggestion, and by no means urgent. I'd like to hear whether others think this is useful.

bayc:
Thanks for this suggestion Bart! I agree, we have a need for turbine identifiers for many reasons (I find myself needing them in layout/yaw optimization). This is currently on my list of things to get included in FLORIS 3.0, so if nothing else, it should be in the next version. But know that it is being worked on and should be available in the near future.

Issue #223, opened 2021-06-29
Feature request: Allowing multiple turbine types


Bartdoekemeijer:
**Is your feature request related to a problem? Please describe.**
It is not particularly related to a problem, but rather a nuisance. Whenever I want to set up a wind farm with more than one turbine type, I feel like I am "hacking" the FLORIS object. Currently, the process goes something along these lines:

- I create one FLORIS .json file with the locations and turbine properties for every wind turbine type
- I load all .json files and combine the floris objects by updating the layout and then the turbine properties:

```
layout_x = []
layout_y = []
turbine_cpct = []
turbine_hh = []
turbine_D = []

for fi in fi_list:
    # Wind farm layout
    layout_x.extend(fi.layout_x)
    layout_y.extend(fi.layout_y)

    # Turbine properties
    num_turbs = len(fi.layout_x)
    t0 = fi.floris.farm.turbines[0]
    turbine_cpct.extend([t0.power_thrust_table] * num_turbs)
    turbine_hh.extend([t0.hub_height] * num_turbs)
    turbine_D.extend([t0.rotor_diameter] * num_turbs)

fi = deepcopy(fi_list[0])

# Copy default turbines to new layout locations
fi.reinitialize_flow_field(layout_array=[layout_x, layout_y])

# Update turbine properties
for ti in range(len(fi.layout_x)):
    fi.change_turbine(
        [ti],
        {
            "power_thrust_table": turbine_cpct[ti],
            "rotor_diameter": turbine_D[ti],
            "hub_height": turbine_hh[ti]
        }
    )
```

This also makes it more difficult to track the `specified_wind_height`. Ideally, FLORIS could simply tell the user to specify an explicit `specified_wind_height` whenever more than one turbine type is used.

**Describe the solution you'd like**
Ideally, the .json would simply allow a variable number of turbine types. Then, one would then need another input in the json in either each `turbine` field or `farm` specifying which turbines belong to which turbine type.

**Describe alternatives you've considered**
I now have a practice where I create a class for these types of wind farms, wherein I contain all these operations. That way, I can still simply call a wind farm using a single command and get a floris object in return.

bayc:
Thanks for this feature request! I have a feature in the works for allowing better definition/changing of turbine Cp/Ct curves. When I submit that pull request, I will be sure to add you as a reviewer and we can make sure that it has the functionality you are looking for.

PR #325, opened 2022-02-18
Support multiple turbine types

bayc:
This pull request _is not_ ready to be merged.

This PR adds the capability to define multiple turbine types within a FLORIS simulation.

#283, #225, #223. Note: the turbine identifier functionality still needs to be added.

Nearly all of the `simulation` directory.

I have added an example, `10_multiple_turbine_types.py`, that showcases the new features. The user is now required to define the turbine type within the input file, at which point, multiple turbine types can be defined. The `turbine_type` array must have the same length as the `layout_x/layout_y` arrays. The turbine definitions can be included directly in the main input file, or referenced with the added yaml handler `!include`. An example of such a reference is given below.

image

Also, the user can make use of the new turbine definition library by providing a valid `turbine_type` with not details (see below), which will use the static turbine definition in `examples/inputs/turbine_definitions`.

image

Running the new example gives the plots below, where the larger rotor size and wake can be seen on the third turbine.

image


examples/10_multiple_turbine_types.py
```diff
@@ -0,0 +1,45 @@
...
+# Initialize FLORIS with the given input file via FlorisInterface.
+# For basic usage, FlorisInterface provides a simplified and expressive
+# entry point to the simulation routines.
+fi = FlorisInterface("inputs/gch_multiple_turbine_types.yaml")
+
+# Using the FlorisInterface functions for generating plots, run FLORIS
+# and extract 2D planes of data.
+horizontal_plane = fi.get_hor_plane(x_resolution=200, y_resolution=100, height=90)
+y_plane = fi.get_y_plane(x_resolution=200, z_resolution=100)
+cross_plane = fi.get_cross_plane(y_resolution=100, z_resolution=100)
```
Bartdoekemeijer, line 36: Initially, I wasn't sure where this cross-plane was located. I see that the default option is 5 * 126 meters downstream. Perhaps this default option is unnecessary (even potentially confusing) and can just become a required user input? Also, it would be nice to show both the wake behind a 5MW turbine, and behind the 10MW turbine, to clearly show that the wake is significantly larger with the larger turbine. Something like:

```
...
cross_plane1 = fi.get_cross_plane(downstream_dist=50, y_resolution=100, z_resolution=100)
cross_plane2 = fi.get_cross_plane(downstream_dist=5*126 + 10, y_resolution=100, z_resolution=100)
cross_plane3 = fi.get_cross_plane(downstream_dist=10*126 + 10, y_resolution=100, z_resolution=100)

# Create the plots
fig, ax_list = plt.subplots(5, 1, figsize=(10, 8))
ax_list = ax_list.flatten()
visualize_cut_plane(horizontal_plane, ax=ax_list[0], title="Horizontal")
visualize_cut_plane(y_plane, ax=ax_list[1], title="Streamwise profile")
visualize_cut_plane(cross_plane1, ax=ax_list[2], title="Spanwise profile")
visualize_cut_plane(cross_plane2, ax=ax_list[3], title="Spanwise profile")
visualize_cut_plane(cross_plane3, ax=ax_list[4], title="Spanwise profile")
plt.tight_layout()
plt.show()
```

examples/inputs/gch_multiple_turbine_types.yaml
```diff
@@ -0,0 +1,94 @@
...
+  layout_y:
+  - 0.0
+  - 0.0
+  - 0.0
+  turbine_type:
+  - nrel_5MW
+  - nrel_5MW
+  - iea_10MW
+  turbine_definitions:
+    nrel_5MW: !include turbine_definitions/nrel_5MW.yaml
```
Bartdoekemeijer, line 32: I really like this! I think this makes it easier for users to see the different turbines implemented in `floris` and what is needed to implement their own turbine.

My only thought is whether something like `layout_z` or `turbine_height` should be added to `gch_multiple_turbine_types.yaml` to allow users to specify the center point of the turbine rotor. For example:

```
farm:
  layout_x:
  - 0.0
  - 630.0
  - 1260.0
  layout_y:
  - 0.0
  - 0.0
  - 0.0
  layout_z:
  - 90.0
  - 90.0
  - 119.0
  turbine_type:
  - nrel_5MW
  - nrel_5MW
  - iea_10MW
  turbine_definitions:
    nrel_5MW: !include turbine_definitions/nrel_5MW.yaml
    iea_10MW: !include turbine_definitions/iea_10MW.yaml
```

and then remove `hub_height: 90.0` and `hub_height: 119.0` from the `nrel_5MW.yaml` and `iea_10MW.yaml` files. Namely, I know working with TU Munich on the MATLAB version of `floris` that they specifically requested this. I think they were working with farms with the same turbine types but different tower lengths, and this would allow them to easily implement this without having to make multiple copies of the same turbine definition. Additionally, it would resolve issue #224.

examples/inputs/turbine_definitions/iea_10MW.yaml
```diff
@@ -0,0 +1,164 @@
+turbine_type: 'iea_10MW'
+generator_efficiency: 1.0
+hub_height: 119.0
```
Bartdoekemeijer, line 3: Could potentially move this to the main input file.

examples/inputs/turbine_definitions/nrel_5MW.yaml
```diff
@@ -0,0 +1,155 @@
+turbine_type: 'nrel_5MW'
+generator_efficiency: 1.0
+hub_height: 90.0
```
Bartdoekemeijer, line 3: Could potentially move this to the main input file (see comment above).

Bartdoekemeijer (commented):
This looks really good, Chris! I didn't look into the underlying code in detail, but I ran the example codes and looked at the input files. The example case clearly demonstrates the functionality and the results seem sensible to me. Thanks for doing this!

paulf81:
hi @bayc  thanks for doing this!  I pulled the latest version, probably you know this but just to capture, example 10 worked well but the early examples are crashing for me.

Also, could we make it that if you just put one turbine type, it assumes they are all the same?  It seems like for a 100-turbine farm copying the turbine name 100 times would be annoying?

rafmudaf:
What do you think of an input structure like this:

```
layout_x:
- 1
- 2
- 3
layout_y:
- 4
- 5
- 6
turbine:
- nrel_5mw
- !include a_turbine.yaml
- iea_10mw
```

It would simplify the input block by having only one section for specifying the turbine type. `nrel_5mw` and `iea_10mw` would be retrieved from the Floris turbine library and `a_turbine.yaml` would arrive at `Floris()` as the contents of the file. I think this would require that the turbine definition be defined outside of this file, though, but that may be a natural progression of our input file anyway.

examples/inputs/turbine_definitions/iea_10MW.yaml
```diff
@@ -0,0 +1,164 @@
+turbine_type: 'iea_10MW'
```
rafmudaf, line 1: Is this used? I think we are mapping this definition to the name in the main input file, so could possibly be removed.

paulf81:
@rafmudaf I think your input looks good, to couple in my suggestion with yours, if it could be that:
```
layout_x:
- 1
- 2
- 3
layout_y:
- 4
- 5
- 6
turbine_type:
- nrel_5mw
```

Would be 3 nrel_5mw turbines, I think a single turbine is used more often than multiple, so would like this to be acceptable

bayc:
I've implemented the one turbine input, and will work to get these other changes integrated. Thanks for the feedback!

examples/inputs/gch_multiple_turbine_types.yaml
bayc: Thanks for the feedback Bart! I've been thinking about this more, and I think that the most probable use case by our users would be to use the hub height defined in the turbine definition, with the varying hub heights being more of a special case for optimization or trying to represent complex terrain. It is something I definitely want to support! But don't want to require the normal user to have to also enter the same hub height value for a homogenous turbine-type farm (e.g. having 100 entries of `layout_z` for a 100 turbine farm would add clutter to the input file). As such, what I am thinking is to allow the user to define `layout_z`, but not require it, and in the cases where it is defined, it will overwrite the hub height values in the turbine definition files. This has some implications on the back end where we use the hub heights in the wake calculations, and isn't immediately trivial to implement. Thus I propose addressing this in v3.1, and have made such a comment on the related issue you shared.

examples/inputs/turbine_definitions/iea_10MW.yaml
bayc: See above comment.

examples/inputs/turbine_definitions/nrel_5MW.yaml
bayc: See above comment.

bayc:
@rafmudaf  @paulf81 @Bartdoekemeijer I have implemented support for specifying just one turbine type, collapsed the turbien type definitions in the input file, and explanded multiple turbine type support to the CC and Jensen wake models. I've also decided it would be best to implement the `layout_z` capability in v3.1. Let me know what you think of the updates!

examples/inputs/turbine_definitions/iea_10MW.yaml
bayc: It is, in that it makes it easier to do some of the farm expansion functions, but wasn't completely necessary with the previous code. Now that the turbine definition has been collapsed in the input file, I do believe it is necessary.

examples/inputs/gch_multiple_turbine_types.yaml
Bartdoekemeijer: That makes a lot of sense. I agree with you that it'll be a niche use case and like your proposed solution. Thanks Chris!

bayc:
> This looks excellent Chris, I like a lot the input formatting. I pulled the code and ran it, there's a small bug in the example which I hope it's ok I'm committing the fix (updating the slice calls)

That's great, thanks for catching that Paul!

rafmudaf:
@bayc What do you think about making `gch.yaml` use multiple turbine types instead of having two GCH input files? `cc.yaml` would show how to use a single turbine type and `gch.yaml` would show how to use multiple.

src/floris/simulation/wake_deflection/jimenez.py
```diff
@@ -63,9 +61,9 @@ def function(
         yaw_i: np.ndarray,
         turbulence_intensity_i: np.ndarray,
         ct_i: np.ndarray,
+        rotor_diameter: np.ndarray,
```
rafmudaf, line 64: Appropriate to use `_i` here?

rafmudaf: Ah I see you did in Gauss. I'll make that change.

paulf81:
> @bayc What do you think about making `gch.yaml` use multiple turbine types instead of having two GCH input files? `cc.yaml` would show how to use a single turbine type and `gch.yaml` would show how to use multiple.
> 
> See https://github.com/NREL/floris/tree/main/examples/other_jsons for an example of the slippery slope.

I take your point on the slippery slope, but I think I prefer to have a clean gch/cc whose only difference is the modeling setup, and one extra, of any model, showing multi-turbine type.  Put another way, I want people to be able to grab the simplest form of cc/gch, and multi-turbine should be an intentional selection, not something that inadvertantly comes along if you start your template copy from cc

src/floris/simulation/wake_deflection/jimenez.py
bayc: Thanks for doing that!

Issue #315, opened 2022-02-14
Need to reload input file when updating/plotting in a loop?


paulf81:
**Bug description**
Had a code set up to change FLORIS settings and plot in a loop and discovered it only worked if I reopened the input file (or if don't plot).  Is this expected operation

**To Reproduce**
Made this code example, it doesn't work for me unless I:
1). Uncomment the code which reopens the model from input yaml
2) Dont plot
```
import matplotlib.pyplot as plt
import numpy as np

from floris.tools import FlorisInterface
from floris.tools.visualization import visualize_cut_plane


# Inputs
X = [0, 500]
Y = [0, 0]
yaw_angles = np.zeros((1, 1, 2)) # 1 wd/ 1ws/ N*N turbines
fi_gch = FlorisInterface("inputs/gch.yaml")


for wd_idx, wind_direction in enumerate([270.,  280.]):

    # Code works if this lined is uncommented
    # fi_gch = FlorisInterface("inputs/gch.yaml") 


    fi_gch.reinitialize(  layout=( X, Y ), wind_speeds=[8.0], wind_directions=[wind_direction], turbulence_intensity=0.06 )
    fi_gch.calculate_wake(yaw_angles=yaw_angles)
    turbine_powers_gch = fi_gch.get_turbine_powers()
    turbine_powers_gch = np.sort(turbine_powers_gch.flatten())[::-1]/1000.

    # Code also works if comment out this block
    fig, ax = plt.subplots()
    horizontal_plane_gch = fi_gch.get_hor_plane(x_resolution=100, y_resolution=100)
    visualize_cut_plane(horizontal_plane_gch, ax=ax, title="Wind Direction = %.1f" % wind_direction)


plt.show()
```

**Expected behavior**
Was expecting the code to work without the call back to the yaml, is something in plotting "sticking"?

**Screenshots, if applicable**
If the code works plots:
![image](https://user-images.githubusercontent.com/6117702/153953706-4ed0fa86-4791-4100-a57b-cd37c79f9a02.png)

**Floris Version**
V3

bayc:
Yea, I noticed this too. My thought was that something was sticking from the grid used for the flow field visualizations, but haven't been able to dig into it yet.

rafmudaf:
@paulf81 can you confirm that this is fixed by #327?

paulf81:
[...]

PR #327, opened 2022-02-18
Bugfix/plotting

bayc:
This PR automatically sets the `FlorisInterface` object's solver settings back to the original settings after using any of the `calculate_horizontal_plane`, `calculate_cross_plane`, and `calculate_y_plane` calls.

#315 

`floris_interface.py`, examples, and any code using the updated methods within `floris_interface.py`

Now when calling either `fi.calculate_wake()` or any call where you would expect to get turbine values after doing a visualization, the code works as intended and does not error.

### DIFF
--- a/examples/01_plot_wakes.py
+++ b/examples/01_plot_wakes.py
@@ -38,12 +38,12 @@ fi = FlorisInterface("inputs/gch.yaml")
 # placed on each rotor plane with 9 points in a 3x3 layout. This can
 # be plotted to show the wind conditions that each turbine is experiencing.
 # However, 9 points is very coarse for visualization so let's first
-# increase the rotor grid to 20x20 points.
+# increase the rotor grid to 10x10 points.
 
 # Create a solver settings dictionary with the new number of points
 solver_settings = {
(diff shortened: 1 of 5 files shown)


examples/01_plot_wakes.py
```diff
@@ -64,9 +65,11 @@
...
 
 # Using the FlorisInterface functions, get 2D slices.
-horizontal_plane = fi.get_hor_plane(x_resolution=200, y_resolution=100)
-y_plane = fi.get_y_plane(x_resolution=200, z_resolution=100)
-cross_plane = fi.get_cross_plane(y_resolution=100, z_resolution=100)
+horizontal_plane = fi.calculate_horizontal_plane(x_resolution=200, y_resolution=100)
+y_plane = fi.calculate_y_plane(x_resolution=200, z_resolution=100)
+cross_plane = fi.calculate_cross_plane(y_resolution=100, z_resolution=100)
+
+fi.calculate_wake()
```
rafmudaf, line 72: @bayc what does this do here?

PR #332, opened 2022-02-23
Resolve errors from divide-by-zero

rafmudaf:
In some models, particular instances arise where based on the location of wind turbines within the farm a 0 denominator may occur. To avoid this, I've added a `NUM_EPS` attribute on the `BaseModel` class and use it when doing division where this may happen.


src/floris/simulation/wake_velocity/jensen.py
```diff
@@ -110,13 +110,13 @@ def function(
...
         # TODO: why do we need to slice with i:i+1 below? This became a problem when adding the wind direction dimension. Prior to that, the dimensions worked out simply with i
-        c = ( reference_rotor_radius / ( reference_rotor_radius + self.we * dx ) ) ** 2
+        c = ( reference_rotor_radius / ( reference_rotor_radius + self.we * dx + self.NUM_EPS ) ) ** 2
         # c *= ~(np.array(x - x[:, :, i:i+1] <= 0.0))  # using this causes nan's in the upstream turbine because it negates the mask rather than setting it to 0. When self.we * (x - x[:, :, i:i+1]) ) == the radius, c goes to infinity and then this line flips it to Nans rather than setting to 0.
         # c *= ~(((y - y_center) ** 2 + (z - z_center) ** 2) > (boundary_line ** 2))
         # np.nan_to_num
 
         # C should be 0 at the current turbine and everywhere in front of it
-        downstream_mask = np.array(dx > 0.0, dtype=int)
+        downstream_mask = np.array(dx > 0.1, dtype=int)
```
bayc, line 119: Should this also be `NUM_EPS`? The 0.1?

rafmudaf: Yes good catch

Issue #331, opened 2022-02-22
Seeing NaN in output for simulations at 5m/s


paulf81:
**Bug description**
Noticed get_farm_power() can return nans for simulations at 5m/s for gch and cc models

**To Reproduce**
```
import matplotlib.pyplot as plt
import numpy as np

from floris.tools import FlorisInterface

# Instantiate FLORIS using either the GCH or CC model
fi = FlorisInterface("inputs/gch.yaml") # GCH model matched to the default "legacy_gauss" of V2
# fi = FlorisInterface("inputs/cc.yaml") # New CumulativeCurl model

# Define a 5 turbine farm
D = 126.
layout_x = np.array([0, D*6, D*12, D*18,D*24])
layout_y = [0, 0, 0, 0, 0]
fi.reinitialize(layout = [layout_x, layout_y])

# Single wind speed
ws_array = [5]

# Output array does contain nan?
print('Full array shows nan in output')
wd_array = np.arange(0,105,5.)
fi.reinitialize(wind_speeds=ws_array, wind_directions=wd_array)
fi.calculate_wake()
farm_power = fi.get_farm_power() / 1E3 # In kW
print(farm_power)

# Output array doesn't contain nan?
print('Sub array (including value at 90 deg where nan was?) doesnt show nan in output')
wd_array = np.arange(80,105,5.)
fi.reinitialize(wind_speeds=ws_array, wind_directions=wd_array)
fi.calculate_wake()
farm_power = fi.get_farm_power() / 1E3 # In kW
print(farm_power)
```

**Expected behavior**
```
Full array shows nan in output
[[1918.47467001]
 [1918.47467001]
 [1918.47467001]
 [1918.47467001]
 [1918.47467001]
 [1918.47467001]
 [1918.47467001]
 [1918.47467001]
 [1918.47467001]
 [1918.47467001]
 [1918.47467001]
 [1918.47467001]
 [1918.47467   ]
 [1918.47462839]
 [1918.44502477]
 [1914.27189111]
 [1775.24422589]
 [1255.26527951]
 [          nan]
 [1275.12982915]
 [1786.34788498]]
Sub array (including value at 90 deg where nan was?) doesnt show nan in output
[[1775.24422589]
 [1255.26527951]
 [ 900.52999968]
 [1275.12982915]
 [1786.34788498]]
```

rafmudaf:
Here's the output of this script after #333 

```
Full array shows nan in output
[[1918.47467001]
 [1918.47467001]
[...]

PR #333, opened 2022-02-23
Enforce min/max bounds on Gauss velocity deficit

rafmudaf:
This pull request enforces realistic bounds on the Gauss velocity deficit. In deep-wake scenarios, a term with the velocity deficit calculation was exceeding the realistic limit and causing a negative term inside a square root.

#331 

Gauss velocity deficit


paulf81:
Pulled this down and made some checks, really looks fixed, thanks!!!

PR #334, opened 2022-02-23
Bound Ct to [0.0001, 0.9999]

rafmudaf:
The thrust interpolation function was bound to [0.0001, 0.9999] but the Ct calculated from the velocities was bound to [0, 1]. In some models, the upper limit of 1 causes a divide-by-zero error.


paulf81:
Looks dope!

PR #335, opened 2022-02-24
Bugfix/cc model plotting

### LINKED ISSUE #298: Support visualization for CC model

bayc:
This is a bugfix to enable plotting when using the CC model.

Resolves #298 

`floris.py` and `solver.py`

This adds a separate full flow field solver, `full_flow_cc_solver`, that follows the same structure as `full_flow_sequential_solver`, except using the CC model solver where appropriate, and having chnages for the flow field solve that match the CC model requirements. Currently, there is a `vel_model` check in `floris_interface.py` to check which solver to use (for both calculations and visualizations). I expect this will eventually be updated to have a more general "solver" class that will enable switching of solvers similar to how we switch velocity and deflection model functions, but this can be addressed in v3.1.

Running example `01_plot_wakes.py` with the `cc.yaml` input file produces the following plots.

image

### DIFF
--- a/src/floris/simulation/__init__.py
+++ b/src/floris/simulation/__init__.py
@@ -39,7 +39,7 @@ from .farm import Farm
 from .grid import Grid, TurbineGrid, FlowFieldGrid, FlowFieldPlanarGrid
 from .flow_field import FlowField
 from .wake import WakeModelManager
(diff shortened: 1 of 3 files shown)


paulf81:
Hi Chris,

  I pulled this down and tried the example.  I also played around with adding yaw and that all checked out.  So this looks great from my perspective!

PR #338, opened 2022-02-24
Save and reset the state of Floris during plotting

rafmudaf:
In the `FlorisInterface.calculate_XX_plane()` functions, save the current state of `Floris()` and reset to this after doing the wake calculation for plotting. This allows the user to continue to use the existing `FlorisInterface` object for further analysis.

#315

`FlorisInterface.calculate_horizontal_plane`
`FlorisInterface.calculate_cross_plane`
`FlorisInterface.calculate_y_plane`


rafmudaf:
Only added and testing with `calculate_horizontal_plane` but I'll add it to the rest if this looks good.

bayc:
Seems ok to me for now. I think we can eventually implement something that doesn't require the re-computing of wakes.

rafmudaf:
Agreed - this is a quick fix (read *hack*)

Issue #329, opened 2022-02-22
PyOptSparse + SNOPT example 


MYMahfouz:
**I am trying to reproduce the work done in this paper "Wind plant system engineering through optimization
of layout and yaw control" https://doi.org/10.1002/we.1836**

I am using PyOptSparse with SNOPT for  wind farm layout optimization. At the beginning of the optimization, I tried to verify that my gradients are ok, and well calculated, but I get an error that some of my Jacobian matrix elements and objectives derivatives are not well calculated. 
If I turn off the "Verify level" option and run the optimization, SNOPT terminates after numerical difficulties, so I am not sure I can trust my outputs

**Describe the solution you'd like**

I would like to know if there is an existing example using SNOPT, or if there are specific options I need to modify. I have currently modified the tolerance as specified in the paper and scaled my variables and constraints.

I would really appreciate any input and help. Thanks.

MYMahfouz:
First of all thanks for adding the layout optimization to FLORIS v3. 

I tried to use the new version with SNOPT, but it doesn't work and I get that the problem is infeasible.

In FLORIS v2.4, I amanged to make SNOPT work. I was only getting that the gradients are calculated wrong if I activate the option "Verify level" and set it to 3. But when I set it to zero I got a result that was very reasonable.

With the new update I am not getting a result, even when i deactivate the "Verify level" option. I didnt check yet what might be causing the problem but wanted to share what I found.

Thanks again for all the effort.

bayc:
Did the pull request #367 fix this new issue you commented above?

MYMahfouz:
Hi, I was just about to send another update.

It solves part of the issue but not completely. Before the pull request, I was getting no answer at all.

Now I am getting and output, but still not as good as the results I obtained using v2.4. 

With v3, the layout configuration is not reaching the same optimum layout and producing much less AEP. The turbines now seem to be allowed to relocate only slightly. 
I tried decreasing the tolerance value, but still not working. I am unable to tell for sure, but I think there is still something that needs modifying.

PR #339, opened 2022-02-25
Enable layout optimization

bayc:
This PR updates the layout optimization tools to work with v3.

#329 While the example is setup to run with the "SLSQP" optimizer, the user can change that option to "SNOPT" to use the SNOPT optimizer (which requires a license and the user to properly install SNOPT within their pyOptSparese installation).

Optimization tools.

This enables basic layout optimization. More options and features will be added in upcoming FLORIS versions. Note that this optimization framework will be ported over to utilize the same framework as the yaw optimization code.

Running the example 10_layout_optimization.py produces the below plot and results output.

image

image


paulf81:
This pull looks good, I reviewed the example and like its simplicity.  I couldn't pyoptspare installed on my mac but that could be my problem.  Recommend we include this ahead of release

Issue #211, opened 2021-06-17
Suggestion to use air density-corrected wind speed to determine power and Ct


ejsimley:
**Bug description**
Currently changing the air density parameter scales the entire power curve up or down. A more realistic approach would be to use an air density-corrected wind speed to determine power. This would shift the wind speed at which rated power is reached, but the turbine would still achieve the same rated power.  

**To Reproduce**
This code produces the following plot, illustrating how the entire power curve is scaled
```
import floris as fl
import numpy as np
from matplotlib import pyplot as plt

fi = fl.tools.floris_interface.FlorisInterface("example_input.json")
ws = np.arange(0.,21.,1.)
pc = fi.get_power_curve(ws)
fi.reinitialize_flow_field(air_density=1.)
pc1 = fi.get_power_curve(ws)

plt.figure()
plt.plot(ws,pc,label='Air Density = 1.225')
plt.plot(ws,pc,label='Air Density = 1.0')
```

<img width="619" alt="image" src="https://user-images.githubusercontent.com/40040961/122478477-b1c55900-cf86-11eb-8ce7-5abeb4ae693d.png">

**Expected behavior**
The change in the power curve expected from the reduction in air density is as follows. Rated power is still achieved, but at a higher wind speed. 

<img width="622" alt="image" src="https://user-images.githubusercontent.com/40040961/122478619-f9e47b80-cf86-11eb-9077-da811d13071c.png">

This can be achieved by scaling the wind speed to create an air-density corrected wind speed used to determine power. A similar approach could be used to determine Ct. This implies that the Cp and Ct curves that are defined should be the standard sea-level values.

```
# Compute the air density-corrected yaw effective velocity
        pW = self.pP / 3.0  # Convert from pP to w
        yaw_effective_velocity = ((self.air_density/1.225)**(1/3)) * self.average_velocity * cosd(self.yaw_angle) ** pW

        # Now compute the power
        return (
            1.225
            * self.powInterp(yaw_effective_velocity)
            * self.turbulence_parameter
        )
```

paulf81:
Hi @ejsimley , the correction to power based on air density is included in V3.  For Ct, I'm not sure if the same fix is appropriate.  Here is the power curve for the NREL 5MW:

![image](https://user-images.githubusercontent.com/6117702/156252712-7c507986-14e7-459a-b953-ad8b2c08b8b6.png)

[...]

Issue #340, opened 2022-02-25
Bug: numerical precision in farm power at high wind speeds


Bartdoekemeijer:
**Bug description**
At high wind speeds, the farm power productions are inconsistent, e.g. for a three 5MW-turbine farm the rated power is often not 15.00 MW but closer to 14.99MW or 14.98 MW. Perhaps this has to do with the precision of the ``cp`` table in the input file. Since the curve does not uniformly increase over a range of wind speeds.

**To Reproduce**
Run `05_sweep_wind_speeds.py` or run `08_opt_yaw_single_ws.py` with a `wind_speeds = [15.0]`.

**Expected behavior**
The power production should be flat for region III operation.

**Screenshots, if applicable**
Zooming in on the region III for the example `05_sweep_wind_speeds.py`, we find:

![image](https://user-images.githubusercontent.com/22119448/155747242-00a42ae3-521d-4e26-b253-01915a3a4146.png)

I also ran a yaw optimization, `08_opt_yaw_single_ws.py`, with `wind_speeds=[15.0]`, and found the following results:
![image](https://user-images.githubusercontent.com/22119448/155746192-9fc7b910-5f6d-41a5-a5e6-0cfc1e059bef.png)

It seems that the yaw angles help change the local wind speed, which can then lead to a power increase due to the wiggles in the power curves.

**Floris Version**
<Please provide as much detail as possible including git commit or version from pip.>
v3.0rc3

**System Information (please complete the following information):** 
 - OS: Ubuntu 18.04 LTS
 - Library versions
   * matplotlib==3.4.3
   * numpy==1.21.0
   * pytest==6.2.5
   * scipy==1.7.1
   * Sphinx==4.4.0

**Additional context**
<Add any other context about the problem here.>

Bartdoekemeijer:
Issue solved by @paulf81 in #341.

PR #341, opened 2022-02-25
Feature/correct turbine power

paulf81:
I think this one will be easier to pull in once the multi-turbine branch is merged in since it starts from there

This pull request corrects a few issues related to getting power given average wind speed.  First it resolves issue #211 by taking the suggested code solution from @ejsimley 

Second, it recomputes cp for the two current turbine model inputs to be smoother above rated (Issue #340)

Finally it adds an example that, for each turbine model in the model library folder, make a small checkout plot to confirm the power curve and power loss to yaw are reasonable and smooth 

Issue #211 
Issue #340


rafmudaf:
@paulf81 Could you review this unit test and update as needed? I think I formulated it incorrectly to begin with but you probably have a better idea of what we should be testing with the `Turbine.power` function. As is, it's effectively just testing the interpolation.

https://github.com/NREL/floris/blob/27dcb5155523f23948cdd6f9d1a211fb5cc8445f/tests/turbine_unit_test.py#L249

ejsimley:
The air density-corrected power calculation looks good. But should the density-correction be applied when calculating Ct as well? I would think so, but could use a second opinion. This could probably be handled in the turbine class `Ct` method.

Bartdoekemeijer:
I think @ejsimley is correct: we would have to apply the same correction for the `Ct` calculation. @paulf81, would you want to do that (or at least open an Issue and put that on the milestones list for v3.1)?

PR #358, opened 2022-03-02
Feature: add IEA 15MW  turbine

paulf81:
95% ready to be merged, just a few last questions

Adds an input model for the IEA 15 MW reference turbine to the library.

Further, adds a little more to the example script 15_check_turbine.py to enable comparison / sanity check on cp/ct surfaces to catch clear errrors.

Only question I had is would we like to revisit the ct values of the NREL 5MW at very low winds?  Seems to rise above 1.0:

image


Bartdoekemeijer:
The NREL 5MW thrust curve is interesting. Even though the values are above `1.0`, they just get clipped by FLORIS in the code:

https://github.com/NREL/floris/blob/95f2aefb05efcaffea8e3a9f63bc4043c0f20595/floris/simulation/turbine.py#L195

We could maybe just clip all the values in the input YAMLs ourselves to `0.9999`, and also assumes values of `0.0001` for below-cut-in wind speeds? I think that would increase the code transparency since then at least there are no hidden operations on the `cp` and `ct` values. It also wouldn't change any of the regression tests.

bayc (approved):

bayc:
The code looks good! I agree with Bart, that perhaps clipping the Ct values in the input file would be more transparent. Would it be worth re-calculating the Ct/Cp values for the 5MW from the latest version of OpenFAST?

paulf81:
> The code looks good! I agree with Bart, that perhaps clipping the Ct values in the input file would be more transparent. Would it be worth re-calculating the Ct/Cp values for the 5MW from the latest version of OpenFAST?

I'm not sure we'd get a different answer if we did recompute, I found this online showing the same high ct values:
https://github.com/NREL/turbine-models/blob/master/Offshore/NREL_5MW_126_RWT.csv

Bartdoekemeijer:
> > The code looks good! I agree with Bart, that perhaps clipping the Ct values in the input file would be more transparent. Would it be worth re-calculating the Ct/Cp values for the 5MW from the latest version of OpenFAST?
> 
> I'm not sure we'd get a different answer if we did recompute, I found this online showing the same high ct values: https://github.com/NREL/turbine-models/blob/master/Offshore/NREL_5MW_126_RWT.csv

Okay, so I'd say let's clip the thrust coefficient values in the input files and get this merged in!

paulf81:
Did it!  I set two values to 0.99. and then for two values where cp was 0 made ct 0 as well

rafmudaf:
Here's the new Cp and Ct curves, for reference
5mw power thrust

rafmudaf (approved):

Bartdoekemeijer (approved):

PR #366, opened 2022-03-04
Bug/calc aep

paulf81:
This is a somewhat fake pull request, I wanted to share the code:

examples/07x_calc_aep_from_rose_diagnose.py

That shows the main issues, it includes an option that lets the calculate wake be called a number of times, yielding unexpectedly different results.

Proposals for resolution:
1) Replace calc_aep with features according to example 7 + @Bartdoekemeijer 's recommendations on teams
2) Not sure what is up with calculate wake thing but seems to be a bug

PR #369, opened 2022-03-04
Redesign AEP calculation function

paulf81:
No, meant to show one way of doing calc farm AEP while avoiding 0 m/s issue

Introduces a new (or replacement) function in FLORIS INTERFACE get_farm_AEP_from_rose, that when passed in a wind rose data frame (columns wd, ws, freq) computes the expected AEP.  Note that this will change the wind speed/ direction to match what is in wind rose (minus the 0m/s case)

Gives a temp example 07b showing that it matches how we compute in the example

Issue #366 

**Impacted areas


Bartdoekemeijer:
Hi @paulf81. I see you implemented a temporary no_wake solution. It currently seems to rely on the assumption that all turbines are identical, and it still runs the wake calculations. I'll see if I can turn this into a more generic `no_wake` option!

paulf81:
ok @Bartdoekemeijer , @bayc  and @rafmudaf this is now updated to use the no_wake option of calculate_wake, my remaining doubt is do I want to insist on the dataFrame input, or allow a second option where 3 seperate arrays are passed in and the dataframe is made internally,

paulf81:
@Bartdoekemeijer  I should've read your comments above more carefully, you make some further good suggestions, would you like to pass on including your suggestions?  

Alternatively, @Bartdoekemeijer and @ejsimley maybe we just wanted to go ahead and get windrose and powerrose back in the mix for V3?  I was thinking they could provide a lot of supporting functions back and forth between dataframe and vectorized representations, csv readers and visualziers?

Bartdoekemeijer:
No worries -- I'd be happy to take a stab at adding my suggestions! I do think there's real value to adding a separate `windrose` class in the floris tools that deals with more complicated AEP calculations. I'll start with the regular calculate_AEP() function for now.

Bartdoekemeijer:
@paulf81 can you check out the latest example and let me know what you think?

Bartdoekemeijer:
OK cleaned things up. I also removed the function `calculate_AEP_wind_limit` which seems to have been some sort of iterative function that finds at what wind speed the AEP of the farm stops increasing. The farm layout, among others, were hardcoded in there, so it seems to me that that function can be removed.

paulf81:
Thanks @Bartdoekemeijer  lets pull it!

jaredthomas68:
> @Bartdoekemeijer I should've read your comments above more carefully, you make some further good suggestions, would you like to pass on including your suggestions?
> 
> Alternatively, @Bartdoekemeijer and @ejsimley maybe we just wanted to go ahead and get windrose and powerrose back in the mix for V3? I was thinking they could provide a lot of supporting functions back and forth between dataframe and vectorized representations, csv readers and visualziers?

I noticed that PowerRose is still not in v4. Are there any plans for getting it back in?

paulf81:
Maybe, but also the WindData objects (including WindRose) are a lot more powerful and integrated then they used to be, and so could be simpler to add missing PowerRose capabilities into them, do you have specific functionality in mind?

PR #376, opened 2022-03-07
More robust file path creation

ElieKadoche:
Using `os.path.join` is more robust than the `+` operator.
It will maybe solve the issue of @econdon16 (related issue #345).

### DIFF
--- a/floris/simulation/farm.py
+++ b/floris/simulation/farm.py
@@ -74,7 +74,9 @@ class Farm(BaseClass):
         for i, val in enumerate(value):
             if type(val) is str:
                 _floris_dir = Path(__file__).parent.parent
-                fname = _floris_dir / "turbine_library" / f"{val}.yaml"
+                fname = os.path.join(_floris_dir,
+                                     "turbine_library",
+                                     "{}.yaml".format(val))
                 if not os.path.isfile(fname):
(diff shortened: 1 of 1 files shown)


bayc:
Thank you for the pull request! This is a good call, and should be more robust. It is still curious for the issue #345 that the standard turbine definitions work, as they use the same code to get those files. I am hoping that they can check the path that is being used and see if that exists or not.

RHammond2:
Hi @ElieKadoche, thanks for this pull request. Based on commentary in #345 it seems like what is ultimately needed is the capability to link to an external turbine library, but the changes proposed here are functionally the same as the original code. For a bit of background, the `os` and `pathlib` libraries are highly complementary, with [`Path` objects](https://docs.python.org/3/library/pathlib.html) offering a simpler API for general path manipulations, and `os` being more useful for lower level string manipulations. Because most of the path manipulations in FLORIS are straightforward, we adopted `Path` in place of `os` for much of the repository ([a bit more on the os-Path mapping](https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module)). In addition, around the same time that `Path` was brought to the forefront in the Python community, the `f-string` was also introduced, to continue towards more user-readable features in Python. The end result is in redesigning  FLORIS, we moved in the direction that Python was moving in improving the readability of the code.

ElieKadoche:
Thanks for all the explanations! So I will close this pull request.

Issue #357, opened 2022-03-01
Allow for horizontal plots to be shown with fixed orientation


paulf81:
**Is your feature request related to a problem? Please describe.**
Currently the horizontal plots are shown rotated such that wind is always left to right.  New feature would allow fixed orientation (such as north as up) or perhaps arbitrary specified orientation.

**Describe the solution you'd like**
Maybe include a boolean argument to visualize_cut_plane or get_x_plane something like fix_north_up

**Additional context**
Issue is responding to discussion #348

PR #377, opened 2022-03-08
Fixed orientation for plots

ElieKadoche:
> NOT READY TO BE MERGED

## What

I try to implement the option to allow plots with fixed orientation.
Related issue: #357.

## Done

I added a boolean argument `north_up` to the `calculate_horizontal_plane`.
I created a related example `17_fixed_orientation.py`.
Plots can now have a fixed orientation.

## TODO

An image is better than words here.
So I am not sure about what to do to have a valid flow field on all the figure.
tmp

### DIFF
--- a/examples/17_fixed_orientation.py
+++ b/examples/17_fixed_orientation.py
@@ -34,8 +34,6 @@ if __name__ == "__main__":
         # Get horizontal plane
         hor_plane = fi.calculate_horizontal_plane(
             height=90.0,
-            # x_bounds=[250.0, 1750.0],
-            # y_bounds=[-250.0, 250.0],
             yaw_angles=None,
             north_up=True,
         )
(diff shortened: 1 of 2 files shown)


rafmudaf:
The flow field domain has to be redefined based on the rotated coordinates. Check out these lines of code:
- https://github.com/NREL/floris/blob/main/floris/simulation/floris.py#L93
- https://github.com/NREL/floris/blob/main/floris/simulation/grid.py#L335

rafmudaf:
Hey @ElieKadoche I've converted this to draft but feel free to mark it "ready for review" when you're ready. In the meantime, we can continue to discuss it here.

ElieKadoche:
I struggle to redefine the flow field.

- Option 1: is the `floris.solve_for_viz()` function enough to reinitialize the flow field?
- Option 2: or do we need to call the `calculate_wake()` function before?

In the case of option 1, I think I miss some important concepts in the FLORIS implementation.
In the case of option 2, I think the required modifications are bigger than I expected.

In both cases, I do not want to take too much of your time.
If I manage to implement this feature, it would be great; otherwise I will close this pull request.

PR #381, opened 2022-03-09
Fix for repeated calls to calculate_wake returning incorrect values.

### LINKED ISSUE #378: Repeated calls to calculate wake can result in wrong wake calculations
**Bug description**
When placed in a for loop with certain turbine layouts, `calculate_wake` does not return the expected values. **Note: this can currently be avoided by calling `fi.reinitialize()` before each repeated calculate wake call, but that adds unnecessary computation.**

**To Reproduce**
1. Run the below code in the examples folder of FLORIS.
2. See the error in the displayed plot, where the total farm power varies with each calculate wake call.

```
import numpy as np


D = fi.floris.farm.rotor_diameters[0][0][0]
layout_x = np.array([0.0, 5.0*D, 10.0*D, 0.0, 5.0*D, 10.0*D])
layout_y = np.array([0.0, 0.0, 0.0, 5.0*D, 5.0*D, 5.0*D])
fi.reinitialize(layout=(layout_x,layout_y))

n = 10
wf_pow = np.zeros(n)
for k in range(n):
    fi.calculate_wake()
    wf_pow[k] = fi.get_farm_power()

plt.figure(1)
plt.plot(wf_pow)
plt.xlabel("calculate wake call #")
plt.ylabel("farm power")
plt.show()
```

image

**Expected behavior**
Expected behavior would be that calculate wake would return the same power with repeated calls where no inputs have changed (example shown below).

image

**Floris Version**
Release v3.0.1

**System Information (please complete the following information):** 
 - OS: Ubuntu 20.04
 - Library versions
   * matplotlib - 3.5.1
   * numpy - 1.22.1
   * pytest - 6.2.5
   * scipy - 1.7.3
   * pandas - 1.4.0

**Additional details**
This is do to the current inclusion of the `finalize` call within the solve wrapper, which unsorts the turbines. The turbines are then not re-sorted from upstream to downstream (which occurs in the initialization step) before the next calculate wake call, which results in incorrect wake calculations.

bayc:
**Not ready to be merged**

This bugfix adds `_sorted` versions of the respective variables that get sorted based on wind direction during the initialization step for use within the solver. In the finalize step after the solve completes, the sorted variables are unsorted and stored in a new variable, therefore allowing repeated calls to `calculate_wake` and maintaining the sorting of the required variables.

Closes #378 

Most of the simulation software, though I am not 100% that I have properly covered all the initialization of the variables (@rafmudaf could use your eyes here).

The code in issue #378 now produces the expected result of:

image

### DIFF
--- a/examples/10_optimize_yaw.py
+++ b/examples/10_optimize_yaw.py
@@ -42,8 +42,8 @@ def load_floris():
     # Specify wind farm layout and update in the floris object
     N = 5  # number of turbines per row and per column
     X, Y = np.meshgrid(
-        5.0 * fi.floris.farm.rotor_diameters[0][0][0] * np.arange(0, N, 1),
-        5.0 * fi.floris.farm.rotor_diameters[0][0][0] * np.arange(0, N, 1),
+        5.0 * fi.floris.farm.rotor_diameters_sorted[0][0][0] * np.arange(0, N, 1),
+        5.0 * fi.floris.farm.rotor_diameters_sorted[0][0][0] * np.arange(0, N, 1),
     )
     fi.reinitialize(layout=(X.flatten(), Y.flatten()))
 
--- a/floris/simulation/farm.py
+++ b/floris/simulation/farm.py
@@ -21,6 +21,7 @@ import os
 import copy
 
 from floris.type_dec import (
+    NDArrayObject,
     floris_array_converter,
     NDArrayFloat
 )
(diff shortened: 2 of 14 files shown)


paulf81:
thanks @bayc !  I made the following checks:

-example 1: Results are unchanged compared to when run on main branch
- example 7: AEP result is unchanged when I add additional unnecessary calls to calculate wake
-example 2: Visualizations all look reasonable
-example 14: Visualization looks reasonable
- other examples: Work as expected


Looks good!

paulf81 (approved):

bayc:
@rafmudaf I updated the tests here to use the `_sorted` versions of u, v, w and x, y, z. Let me know what you think!

Issue #379, opened 2022-03-08
Wind turbine power being extrapolated past rated wind speed


Bartdoekemeijer:
The current implementation for the turbine power interpolant extrapolates the power production to high wind speeds past the upper bound of the defined curve. This means that FLORIS will predict rated wind farm power production even at wind speeds of 100 m/s.

This has to do with this line, where the power production values are extrapolated to the right bound of the specified wind speeds:
https://github.com/NREL/floris/blob/df4622fab9547433f94ff13380456fa3360b73b1/floris/simulation/turbine.py#L372

This may cause unpredictable behavior. Perhaps we should just throw an error when wind speeds are evaluated that are not included in the turbine power table? It should be easy enough for users to fix such errors by simply including a zero point at cut-out and a zero point at a very high wind speed.

**Example code**
This issue can easily be demonstrated using the following code:
```
# Initialize FLORIS object
fi = FlorisInterface("inputs/gch.yaml") # Initialize GCH model
wind_speeds = np.arange(1.0, 50, 0.50)
fi.reinitialize(layout=[[0, 600], [0, 0]], wind_speeds=wind_speeds)

# Calculate solutions
fi.calculate_wake()
farm_power = fi.get_farm_power() / 1.0e6 # In MW

# Produce wind-speed power curve
fig, ax = plt.subplots()
ax.plot(wind_speeds, farm_power.flatten())
ax.set_xlabel("Wind speed (m/s)")
ax.set_ylabel("Farm power production (MW)")
ax.grid(True)
plt.show()
```

which returns
![image](https://user-images.githubusercontent.com/22119448/157339428-f9ce01de-96ae-494f-9a75-7e5847559762.png)

Issue #375, opened 2022-03-07
Add a default option to reference wind height


paulf81:
With the latest input format, the reference wind height and turbine hub heights are not specified in the same file.  To avoid accidentally changing one without the other, recommend having a "default" option to reference for the reference wind height

**Describe the solution you'd like**
Instead of a number, reference_wind_height accepts the term "hub_height" or similar.  If only one turbine_type is identified, the turbine hub_height is now the reference wind height.  If multiple turbines are specified, an error is thrown informing the user that if multiple turbine types are specified, it is necessary to manually set the reference wind height.

An additional concept could be that in the event:
1) only one turbine type is specfied
2) the reference wind height does not equal the turbine hub height

Have FLORIS produce a warning as it could be the turbine type was changed without realizing the reference wind height is fixed to a manual setting

PR #382, opened 2022-03-09
Add cut out and default reference option

paulf81:
Ready to be merged

This code adds a few features:
- All example turbine inputs have a defined cut-out region
- YAML inputs can specify -1 as reference wind height to indicate that turbine hub height should be used
- Provides a function to set the reference wind height to the turbine height in case the turbine model is changed via reinitialize (as in example 15)

Issue #379 
Issue #375 

**Last doubt**
One change I considered was changing reference_wind_height to be defined as a float rather than an int as it is currently.  Is there a reason it should be an int?


Bartdoekemeijer:
Thanks for getting this arranged so quickly, Paul! Last couple things:

- I agree that `reference_wind_height` should be a float
- Could you expand the NREL 5MW power table down to 0 m/s?
- Now that the turbine power and thrust are defined over such a wide range of wind speeds, shall we disable extrapolation in `turbine.py`?:

https://github.com/NREL/floris/blob/df4622fab9547433f94ff13380456fa3360b73b1/floris/simulation/turbine.py#L379

such that an error is thrown if we try to interpolate values outside of the power/thrust tables.

paulf81:
Thank you for your suggesitons @Bartdoekemeijer  I have adopted all of them

Bartdoekemeijer:
@paulf81 looks good to me! Also just confirmed that I'm getting errors when trying to evaluate negative and very high wind speeds (past 50 m/s). This PR is ready to be pulled in, in my opinion!

On a separate note, we're starting to get close to 20 example scripts. Maybe we should split the examples into different subfolders by topic/application? Anyhow, not very important for the current pull request but something to think about.

Bartdoekemeijer (approved):

PR #384, opened 2022-03-09
Selectively invert plot axis via normal vector in CutPlane

rafmudaf:
This pull request adds an option to invert the flow field and rotor plane plots. Specifically, floris currently plots y-planes from the perspective of downstream of the turbine and looking upstream. This pull request inverts the axes when we are plotting a `CutPlane` with x normal vector.

Below are plots of three turbines in a row with the leading turbine yawed with the latest `develop` and this pull request. The spanwise plots and the rotor plane plots are inverted.

None

Visualization tools and `CutPlane` class

## Current `develop` branch
up1
up2

## This pull request
down
down2

### DIFF
--- a/floris/tools/cut_plane.py
+++ b/floris/tools/cut_plane.py
@@ -105,7 +105,7 @@ class CutPlane:
     FLORIS simulation, or other such as SOWFA result.
     """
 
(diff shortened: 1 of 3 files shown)


paulf81:
this is great, thanks @rafmudaf

Issue #383, opened 2022-03-09
Bug report: verify_convergence not working in yaw optimization, example 07


Bartdoekemeijer:
**Bug description**
<A clear and concise description of the bug.>
The option `verify_convergence` appears to be broken for yaw optimizations with `n_wind_speeds > 1`.
 
**To Reproduce**
<Update the following list with your specific information.>
1. Replace line 42 in `09_opt_yaw_multiple_ws.py` with `yaw_opt = YawOptimizationSR(fi, verify_convergence=True)`
2. Run the script
3. See the error
```
Exception has occurred: IndexError
index 814 is out of bounds for axis 0 with size 814
```

**Expected behavior**
Successful completion of the code, and filter out potential issues at high wind speeds where large yaw offsets achieve marginal improvements on the farm power (in the order of < 1 W).

**Floris Version**
The develop branch of `v3.0`, commit version `6f035d`.

Bartdoekemeijer:
Solved with #385

PR #385, opened 2022-03-09
Bug fix: revise verify_convergence to support multiple wind speeds

Bartdoekemeijer:

This feature fixes bug #383 where `verify_convergence` does not work if more than one wind speed is defined in FLORIS. 

Issue #383 

Yaw optimization.

In the past, we would first locate which turbines have a nonzero yaw angle and at what wind directions and wind speeds. We would only check the wind farm's power production for those turbines and conditions with and without yaw misalignment. With the grid-style way FLORIS is calculated, it's not possible to isolate exactly which wind direction/speed combinations we need to evaluate without doing a one-by-one loop, which is slow. Hence, the current method is simplified and just evaluates all atmospheric conditions num_turbs times, and for each atmospheric condition set we reset one of the turbine's yaw angle to its baseline value and calculate the power production. This means we may do many more calculations than necessary, but it does keep things in gridded format.

Running example `09_opt_yaw_multiple_ws.py` with `verify_convergence=False` yields:
image

In which we see nonzero yaw angles at high wind speeds of 15-17 m/s, due to numerical precision. These yaw angles just shift the `cp` value ever so slightly that the wind farm power is slightly higher. With `verify_convergence=True`, it yields:
image

Though, I have seen that the computation time can increase by 10-20% using `verify_convergence=True`. Hence, it still defaults to `False` and it's up to the user to decide whether one would like to include this functionality. Perhaps once the timeseries calculation functionality for FLORIS is implemented (#299) I can restructure this code to more efficiently evaluate a minimal set of conditions.

### DIFF
--- a/examples/09_opt_yaw_multiple_ws.py
+++ b/examples/09_opt_yaw_multiple_ws.py
@@ -38,7 +38,17 @@ fi.reinitialize(
 )
 
 # Initialize optimizer object and run optimization using the Serial-Refine method
-yaw_opt = YawOptimizationSR(fi)
+# Now, we enable the verify_convergence option. This function is useful to prevent
+# yaw misalignment that increases the wind farm power production by a negligible
+# amount. For example, at high wind speeds (e.g., 16 m/s), a turbine might yaw
+# by a substantial amount to increase the power production by less than 1 W. This
+# is typically the result of numerical inprecision of the power coefficient curve,
+# which slightly differs for different above-rated wind speeds. The option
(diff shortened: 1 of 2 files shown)


Bartdoekemeijer:
Also fixed the example case and added explanation on how `verify_convergence` should be used. Once @paulf81 approves, I think this is ready to be merged in.

paulf81 (approved):
This looks good to me @Bartdoekemeijer , I tested example 09 and looks great

Issue #380, opened 2022-03-08
Bug report: legacy input reader not working


Bartdoekemeijer:
**Bug description**
The legacy input reader in `floris_interface.py` called `FlorisInterface_legacy_v24` does no longer work. This has to do with a recent PR #325 that changed the way turbines are defined in `floris`.

**To Reproduce**
1. Copy the old example input file from FLORIS `v2.4`
2. Try reading that file with:
```
from floris.tools import FlorisInterface_legacy_v24
fi = FlorisInterface_legacy_v24("inputs/gch_legacy_format_v24.json")
```
3. See the error: `The initialization for Floris was given extraneous inputs: ['turbine']`

**Expected behavior**
Successfully load the legacy .JSON file.

**Screenshots, if applicable**
<Add screenshots to help explain your problem.>

**Floris Version**
`v3.0`

Bartdoekemeijer:
Yes!

PR #387, opened 2022-03-11
Bug fix and formatting for legacy reader for FLORIS V2.4 input files

Bartdoekemeijer:
This pull request is ready to be merged.

Issue #380 explains that the legacy input reader, which would allow users to read V2.4 FLORIS input files directly into the floris v3.0 `FlorisInterface` object, was broken due to recent changes in how turbines are assigned. This PR addresses that bug. Additionally, I moved the entire legacy input reader to its own file in `floris.tools` to reduce clutter in `floris_interface.py`.

Issue #380.

Legacy input reader that allows users to directly read FLORIS v2.4 input `.json` files into `v3.0`.

I have now placed the legacy input reader into a file called `floris_interface_legacy_reader.py`. Is this OK or would you say this can become confusing for new users with having a name similar to `floris_interface.py`?

I've added an example that demonstrates the usage. Essentially, it is as simple as:

```
from floris.tools import FlorisInterfaceLegacyV2
fi = FlorisInterfaceLegacyV2("inputs/legacy_example_input.json")
```
Which translates the legacy input file in an appropriate `configuration` dictionary and a `turbine_type` dictionary, which are then loaded into a v3.0 `FlorisInterface` object.


examples/inputs/legacy_example_input.json
```diff
@@ -0,0 +1,258 @@
+{
```
rafmudaf, line 1: I'm wondering if there's a way to demo this without including a legacy input file in the example inputs. In v2 we had a lot of [input files](https://github.com/NREL/floris/tree/v2/examples/other_jsons) and the differences between them were lost due to noise, so I'd like to only add another one here when we definitely need it. For this use case, would it be enough to include a script in the docstring for the new class?

floris/tools/floris_interface_legacy_reader.py
```diff
@@ -0,0 +1,184 @@
...
+
+from __future__ import annotations
+
+import copy
+from pathlib import Path
+
+from floris.tools import FlorisInterface
+
+
+class FlorisInterfaceLegacyV2(FlorisInterface):
```
rafmudaf, line 23: I could see this file also including a `if __name__=="__main__"` so that it could be executed directly to convert a file from v2 to v3 without running the simulation.

floris/tools/floris_interface_legacy_reader.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+    FlorisInterface_legacy_v24 provides a wrapper around FlorisInterface
+    which enables compatibility of the class with legacy floris v2.4 input
+    files. The user can simply pass this class the path to a legacy v2.4
+    floris input file to this class and it'll convert it to a v3.0-compatible
+    input dictionary and load the floris v3.0 object.
+
+    After successfully loading the v3.0 Floris object, you can export the
+    input file using: fi.floris.to_file("converted_input_file_v3.yaml").
```
rafmudaf, line 32: Nice use of the `.to_file` method. This worked out really well.

rafmudaf:
@Bartdoekemeijer I've left a few review comments as feedback but overall nicely done and looks good to me

examples/inputs/legacy_example_input.json
Bartdoekemeijer: Yeah sounds good to me! I'll work on getting that in.

Bartdoekemeijer:
@rafmudaf Thanks for the review! I added the `if __name__ == __main__` idea and removed the example files for clarity. What do you think?

bayc:
Looks good to me! Good to merge @rafmudaf ?

PR #388, opened 2022-03-11
Save yaw angles for reset in plane visualizations

rafmudaf:
Currently, the state of `floris.farm.yaw_angles` is not saved during the `floris`-object export in `FlorisInterface.calculate_XX_plane()` functions. While we generally expect that yaw angles will have to be reset for additional calculations, the intention for resetting the `floris`-object to its previous state after the visualization calculation was to allow for additional post processing in the script. Not saving the yaw angles means that any additional analysis will be different before and after `FlorisInterface.calculate_XX_plane()`.

None

`FlorisInterface.calculate_horizontal_plane()`
`FlorisInterface.calculate_y_plane()`
`FlorisInterface.calculate_cross_plane()`

Generally, these API's will will be improved in the future along with the rest of the `floris.tools` package.

### DIFF
--- a/floris/tools/floris_interface.py
+++ b/floris/tools/floris_interface.py
@@ -335,10 +335,9 @@ class FlorisInterface(LoggerBase):
         # Reset the fi object back to the turbine grid configuration
         self.floris = Floris.from_dict(floris_dict)
         self.floris.flow_field.het_map = self.het_map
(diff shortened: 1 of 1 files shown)


paulf81 (approved):
This seems good to me!

PR #392, opened 2022-03-15
Update wind rose and power rose

paulf81:
Would like to update and reincorporate the wind rose and power rose tools.  There are some points I have open questions:
 - @ejsimley : is wind toolkit stuff still current / need an update?
 - @Bartdoekemeijer : do you have preferences on putting things in current/matching style of wd/ws/turbine matrices?


paulf81:
Maybe this one is getting closer but I had a couple of questions:

@ejsimley : at the bottom of the wind rose file there are a number of tools related to getting data from the wind toolkit, do you know, do these still work?  Or has the recommended usage changed?  Do we still want this all in, or does this now belong in flasc or pruf or something?  What do you think?

@Bartdoekemeijer  and @ejsimley : similar question, scrolling through power rose, I was wondering if this isn't doing things we handle more intrinsically in V3 since the results come back vectorized?

floris/tools/floris_interface.py
```diff
@@ -722,6 +722,111 @@ def get_farm_AEP(
...
+        # over to those values in the wind rose class
+        wind_speeds_wind_rose = wind_rose.df.ws.unique()
+        wind_directions_wind_rose = wind_rose.df.wd.unique()
+        self.reinitialize(wind_speeds=wind_speeds_wind_rose, wind_directions=wind_directions_wind_rose)
+
+        # Build the frequency matrix from wind rose
+        freq = wind_rose.df.set_index(['wd','ws']).unstack().values
+
+
+        # Verify dimensions of the variable "freq"
```
Bartdoekemeijer, line 777: After this point, could you just use `fi.get_farm_AEP`?

floris/simulation/wake_velocity/cumulative_gauss_curl.py
```diff
@@ -175,21 +175,24 @@ def function(
...
-                (n * turbine_Ct[:,:,ii:ii+1])
-                * cosd(turbine_yaw)
-                / (
-                    16.0
-                    * gamma(2 / n)
-                    * np.sign(sigma_n)
-                    * (np.abs(sigma_n) ** (4 / n))
-                    * (1 - sum_lbda) ** 2
-                )
+        tmp = a2 - (
```
Bartdoekemeijer, line 178: Should this fix go in a separate Pull Request? Is this generic enough to be merged back into `develop` or is this just a temporary fix for something?

Bartdoekemeijer (commented):
Just a quick couple comments! I agree that power_rose has been pretty much replaced by FLORIS v3. `power_rose` does have a couple functions like plotting by direction and additionally calculating a baseline scenario and a `no_wake` scenario, but that's very easy with FLORIS v3 at this point, too. I vote that we just remove `power_rose`, and perhaps only carry over the visualization function to `floris.tools.visualization`.

floris/tools/floris_interface.py
paulf81: that is a good suggestion @Bartdoekemeijer , just commited a version that does this

rafmudaf:
> 2) It also includes a fix to the cumulative curl model which in theory could go in separately, @rafmudaf what do you prefer?

I'd say the best is to include that in a separate pull request and make it to the `main` branch. However, if that's a pain, at least it would be good to describe that change in the original post of this pull request.

paulf81:
> 

done!  see pull request #438

Bartdoekemeijer:
I think it looks good, but the additional commits surrounding layout optimization makes it hard to identify whether each change is really required. I think I'd like to have #429 merged before we merge this one. That'll make things clearer.

paulf81:
> I think it looks good, but the additional commits surrounding layout optimization makes it hard to identify whether each change is really required. I think I'd like to have #429 merged before we merge this one. That'll make things clearer.

Agreed!!

Issue #394, opened 2022-03-17
Feature request: ability to disable wake deflection submodel to speed up calculations


Bartdoekemeijer:
**Is your feature request related to a problem? Please describe.**
#270 suggests the option to assign "None" to the wake deflection model choice, which then skips those calculations. This can further speed up computations for when we are not interested in wake deflection.

**Describe the solution you'd like**
Ability to assign `None` to the wake deflection submodel. Perhaps something similar for other submodel types, such as the local turbulence model and perhaps even the wake deficit model, assuming it is simple to do.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

paulf81:
This seems like a good idea to me too

rafmudaf:
One idea for implementing this is to have a "no_deflection" wake deflection model that simply returns zero. Something like this:

```Python
@define
class NoDeflection(BaseModel):
    """
    Don't do anything
    """

    model_string = "none"

    def prepare_function(
        self,
        grid: Grid,
        flow_field: FlowField,
    ) -> Dict[str, Any]:
        return dict()

    # @profile
    def function(
        self,
        x_i: np.ndarray,
        y_i: np.ndarray,
        yaw_i: np.ndarray,
        turbulence_intensity_i: np.ndarray,
        ct_i: np.ndarray,
        rotor_diameter_i: np.ndarray,
        *,
        x: np.ndarray,
    ):
        return 0.0
```

I think a scalar 0.0 will work but that should be double checked. If not, you could get `grid.x` in the prepare function and use that to get the right shape for `np.zeros()`.

Bartdoekemeijer:
@rafmudaf this is exactly what I have implemented. I was just struggling with GitHub and pushing up that branch to my fork, but essentially that is what I did for the deflection, deficit and added turbulence models.

PR #395, opened 2022-03-17
Add 'none' submodel option for wake deflection, deficit and turbulence

Bartdoekemeijer:
Adds the option to use `none` or `None` for the wake deficit, wake deflection and added-turbulence submodels.

#394 

FLORIS simulation.

N/A

Example 17 has been added that compares the farm power with and without a deflection model. It shows that the non-deflection model is notably faster (20%+) while yielding pretty much identical results.

```
Time spent (with defl. model): 1.576 s.
Time spent (without defl. model): 1.321 s.
```
image

### DIFF
--- a/examples/17_compare_models_nodeflection.py
+++ b/examples/17_compare_models_nodeflection.py
@@ -1,0 +1,78 @@
+# Copyright 2022 NREL
+
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+
+# See https://floris.readthedocs.io for documentation
+
+
+from time import perf_counter as timerpc
+import matplotlib.pyplot as plt
+import numpy as np
+
+from floris.tools import FlorisInterface
+
+
+"""
+04_sweep_wind_directions
+
+This example demonstrates vectorization of wind direction.  
+A vector of wind directions is passed to the intialize function 
+and the powers of the two simulated turbines is computed for all
+wind directions in one call
+
+The power of both turbines for each wind direction is then plotted
+
+"""
+
+# Instantiate FLORIS using either the GCH or CC model
+fi = FlorisInterface("inputs/gch.yaml") # GCH model matched to the default "legacy_gauss" of V2
+fi_nodeflection = FlorisInterface("inputs/gch_nodeflection.yaml")
+
+# Define a two turbine farm
+wd_array = np.arange(0.0, 360.0, 1.0)
+layout_x = 5 * 126.0 * np.arange(30)  # 30 turbines
+layout_y = np.zeros_like(layout_x)
+
+fi.reinitialize(
+    layout = [layout_x, layout_y],
+    wind_directions=wd_array
+)
+fi_nodeflection.reinitialize(
+    layout=[layout_x, layout_y],
+    wind_directions=wd_array,
+)
+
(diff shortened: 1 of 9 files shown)


paulf81:
Hi @Bartdoekemeijer this is really cool, on my machine the speed-up is about 22%, cool beans man!

A few questions:

1. Should this go into 3.1, and the 3.1 announcement?  I think a fair number of users don't use deflection
2. What should the behavior be if a yaw angle is specified and the deflection model is off?  When I checked it computes a result but I wasn't sure if that should be allowed since with the deflection model off it is a little suspect?
3. Similar question for the turbulence model, should this be only allowed to be off for certain wake models?

Bartdoekemeijer:
Thanks @paulf81! Glad to see the timing improvement works for you too.
1. I think we can include this in 3.1. It's a small piece of code but very useful for the wider public. Should be easy to get merged in before our announcement.
2. What will happen is that turbines may lose power from being misaligned with the wind yet the wakes are not deflected. Essentially, I think this aligns with what I as a user would expect to happen. As in, I don't think this is ambiguous and this is predictable behavior of the code given the inputs. I do think we can print a warning message to make the users aware that they don't have a wake deflection model enabled.
3. Good point. I guess the same question can arise with the lack of a wake deficit model.
 
So should we raise a warning in the following scenarios?
- No wake deflection model, `yaw_angles` not all zeros
- No wake deficit model, all scenarios
- No turbulence model, specific submodels (Gauss, CC, any others @paulf81?)

What do you think? I can get this implemented if you agree.

paulf81:
Hi @Bartdoekemeijer  this warning approach sounds great, you're right that it least it behaves as expected.  I like the warning list you give, I don't think TurboPark will use a TI model, right @bayc ?

rafmudaf:
In my opinion, this sort of input checking should happen in the `FlorisInterface`. Other similar input validation checks have been talked about, so it would worthwhile to have a single function that does this validation and raises errors.

Bartdoekemeijer:
OK @paulf81, I added those warnings. I couldn't specifically add a warning for the wake-turbulence model that would raise a warning depending on the wake model used, because I don't think the wake model choice is available at that level. So now,

- An INFO is raised when the wake deflection or wake-turbulence submodels is set to `none`.
- A WARNING is raised when any of the `np.abs(yaw_angles)` are larger than 1 deg yet the deflection model is disabled
- A WARNING is raised when the wake deficit model is disabled.

rafmudaf:
Just saw your commit @Bartdoekemeijer and looks good. I missed that it was going to be a warning and not an error.

bayc:
@Bartdoekemeijer Is there a situation where you would want to run a calculation with non-zero yaws but not calculate deflection? That kind of behavior is initially concerning to me, as a user could forget to turn deflection back on and take the results as yawed and accurate. I agree, the code behaves as intended, but this seems like it could cause some headaches.

Bartdoekemeijer:
Good point @bayc. My initial reasoning was based on the fact that the yaw angles passed to `fi.calculate_wake` are not always equal to the yaw angles passed to the deflection model `none.py` when `enable_secondary_steering=True` because of the [wake-added yaw functionality](https://github.com/NREL/floris/blob/df4622fab9547433f94ff13380456fa3360b73b1/floris/simulation/solver.py#L107). Hence, even if the user specifies all yaw angles to be zero, the yaw angles passed to the deflection model can actually be several degrees positive or negative. Thinking about this more, I agree with your suggestion on throwing an error. How about we indeed throw an error when any of the effective yaw angles are nonzero, and add a message that the user must disable `secondary_steering` if they wish to use the `none` deflection model?

bayc:
@Bartdoekemeijer  I see, that makes sense. And that sounds like a good solution to me!

Bartdoekemeijer:
Done! Shall I also delete the example script to declutter things a bit, @bayc?

bayc:
@Bartdoekemeijer Yea, that would be great to remove the example script and the example input file, thanks!

Bartdoekemeijer:
Deleted the example files. Ready to be merged, I think!

bayc (approved):

PR #396, opened 2022-03-17
Adding the TurbOPark wake velocity deficit model

### LINKED ISSUE #297: Add TurbOPark model

bayc:
This pull request add the TurbOPark wake velocity deficit model, respective solver, and tests to the FLORIS package. For more details on the TurbOPark model, see: https://github.com/OrstedRD/TurbOPark.

Closes #297 

Most of the `simulation` directory and parts of `FlorisInterface`.

This is an initial implementation of the TurbOPark model. Due to the nature of the model, it does not currently support plotting of the wakes (flow field points are assumed to be at turbine locations). Additionally, the current implementation does not support wake steering. Either plotting or setting of yaw angles will cause an error to be issued when using this model. Also, the model is currently only meant to be used with the linear (FLS) wake combination method.


floris/simulation/grid.py
```diff
@@ -197,13 +197,16 @@ def set_grid(self) -> None:
...
         )
         # Calculate the radial distance from the center of the turbine rotor
-        disc_grid = np.linspace(
-            -1 * disc_area_radius,
-            disc_area_radius,
-            self.grid_resolution,
-            dtype=floris_float_type,
-            axis=1
-        )
+        if self.grid_resolution == 1:
```
rafmudaf, line 200: @bayc could you describe this change in a code comment?

rafmudaf:
@bayc we also need to update the TurbOPark solver to use the `_sorted` variables where necessary

floris/simulation/grid.py
bayc: Added! I realized this was another bug if a grid resolution of 1 was selected, as the one point would be placed below the rotor center.

rafmudaf (approved):

Issue #401, opened 2022-03-30
Feature request: Code parallelization


Bartdoekemeijer:
**Is your feature request related to a problem? Please describe.**
I would like to have the option to parallelize the floris calculations that require high computational resources, such as `calculate_wake`, `get_farm_AEP` and `YawOptimizationSR`.

**Describe the solution you'd like**
I'm not sure at what level we should implement parallelization. If we can implement parallelization on the `simulation` level, and the overhead is sufficiently small, that would be the most generic and would also suffice for yaw optimizations, the uncertainty interface, and other applications. However, if the overhead is significant, (additionally) parallelizing the code at a higher level might make more sense.

Also, parallelization should be compatible both with `multiprocessing` and with `mpi`.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

rafmudaf:
Is this suggestion around splitting up a single wind condition (finding the wake of individual turbines concurrently) or parallelizing the calculation of multiple wind conditions (find the wake of the full farm for multiple wind conditions concurrently)?

Bartdoekemeijer:
@rafmudaf parallelizing the calculation of multiple wind conditions

Bartdoekemeijer:
Here's an example demonstrating wake steering optimizations and AEP calculations (assuming uniform distribution) parallelized at the user level. Would it make sense to include this as an example script for users, either as a temporary or permanent solution? It also contains optimizations to include uncertainty in the optimization and/or AEP optimization, making it generally a good starting point for practical usage.

Note that parallel performance depends a lot on the problem size (number of turbines and wind conditions) and on the resources available. My 8-core laptop with 64GB of memory sees a factor 5-6x improvement from parallelization in a 20-turbine application.

This function also calls upon some functionality within [FLASC](https://github.com/NREL/flasc/). Would be worth considering if that's worth keeping and in what manner.

Anyhow, please feel free to ignore/copy/modify/implement this code as you see fit. Thanks!


```
from time import perf_counter as timerpc
import matplotlib.pyplot as plt
import multiprocessing as mp
import numpy as np
import pandas as pd
from scipy.interpolate import NearestNDInterpolator

from floris.tools import UncertaintyInterface, FlorisInterface
from floris.tools.optimization.yaw_optimization.yaw_optimizer_sr import (
    YawOptimizationSR,
)
from flasc.wake_steering import lookup_table_tools as lut_tools
[...]

PR #403, opened 2022-03-30
Feature: parallelization in the floris interface

Bartdoekemeijer:
This PR is **not** ready to be merged.

This feature adds a parallelization option to `calculate_wake` in `FlorisInterface`.

Issue #401.

Simulation core of FLORIS.

This PR adds two options to `calculate_wake`, being `num_tasks=1` and `use_mpi=False`. If `num_tasks > 1`, the code is simulated using parallel processing. If `use_mpi=False` (default), it'll use Python's `multiprocessing` library for that. If `use_mpi=True`, it uses the `mpi4py.futures` module for parallelization.

This PR is one way I envision is useful to speed up computations, particularly for large farms. The `calculate_wake` parallelization function can easily reflect in other functions such as `get_farm_AEP` by passing the `num_tasks` and `use_mpi` arguments. Additionally, it'll also speed up computations for `UncertaintyInterface`, since that directly uses `calculate_wake`. For yaw optimizations, this PR may yield some speedup, but it is probably better to parallelize the code at a higher level for better performance.

Some open questions:
1. Do we want to have parallelization on the `floris_interface` level, like I propose it to be now? Or should it reside as a separate class on top of `floris_interface`, perhaps?
2. Are all relevant variables gathered after processing the tasks? I am currently mainly collecting the flow information.

I ran the following code on Europa, NREL Eagle's Jupyter Notebook interface. I believe Jupyter has 32 cores readily available for parallelization. Here, I am benchmarking the parallelization performance using `multiprocessing` for various farm sizes.

```
import numpy as np

from floris.tools.visualization import visualize_cut_plane


def load_floris(nx=10):
    # Initialize FLORIS object
    X, Y = np.meshgrid(np.arange(nx) * 5 * 126.4, np.arange(10) * 3 * 126.4)
    fi.reinitialize(
        wind_directions=np.arange(0.0, 360.0, 3.0),
        wind_speeds=np.arange(4.0, 25.0, 1.0),
        layout=[X.flatten(), Y.flatten()]
    )
    return fi


for nx in range(1, 10):
    # Calculate with parallelization
    num_task_list = [1, 2, 3, 4, 8, 12, 18, 24, 30]
    timings = np.zeros(len(num_task_list))
    for ii, num_tasks in enumerate(num_task_list):
        fi = load_floris(nx=nx)
        t0 = timerpc()
        fi.calculate_wake(num_tasks=num_tasks)
        timings[ii] = timerpc() - t0
    
    fig, ax = plt.subplots()
    ax.plot(num_task_list, timings, "-o")
    ax.set_title("Farm with {:d} turbines".format(nx * 10))
    ax.set_xlabel("Num of tasks in parallelization")
    ax.set_ylabel("Time to completion (seconds)")
    ax.grid(True)  
```
download
download
download
download
download
download
download
download
download

Parallelizing the code adds notable overhead and often leads to slow-downs in small farms. Also, FLORIS is particularly fast because the code is vectorized, and splitting it up into multiple problems will reduce the benefit of that. Parallelization is mostly useful for large wind farms and many atmospheric conditions, and with the right computational resources (e.g., a high performance computing facility).

### DIFF
--- a/examples/04_sweep_wind_directions.py
+++ b/examples/04_sweep_wind_directions.py
@@ -13,7 +13,6 @@
 # See https://floris.readthedocs.io for documentation
 
 
-from time import perf_counter as timerpc
 import matplotlib.pyplot as plt
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


Bartdoekemeijer:
This has been superseded by #555 so happy to close this one.

PR #407, opened 2022-04-04
Bugfix: Resolve issues with no_wake=True

Bartdoekemeijer:
This PR **is** ready to be merged.

Resolves the issue with `no_wake=True` in #405.

See thread in #405.

FLORIS simulation.

N/A

Example 07 now runs without issues and with results as expected.


floris/simulation/flow_field.py
```diff
@@ -101,6 +101,10 @@ def initialize_velocity_field(self, grid: Grid) -> None:
         self.v_sorted = self.v_initial_sorted.copy()
         self.w_sorted = self.w_initial_sorted.copy()
 
+        self.u = self.u_sorted.copy()
```
rafmudaf, line 104: @bayc Is this valid? I think `flow_field.u` and etc are meant to be the unsorted versions.

paulf81:
this looks good to me @Bartdoekemeijer  but want to check @bayc 's response to @rafmudaf 's comment

floris/simulation/flow_field.py
Bartdoekemeijer: Ah my bad! @bayc the `steady_state_atmospheric_solver` is not called when `no_wake=True`, and so the `finalize()` commands are never called. That then gives an error because `.u`, `.v` and `.w` are undefined.

bayc: I see! I'll dig into this a bit and see if I can come up with a solution.

Bartdoekemeijer: Thank you @bayc. Another idea I had is to have `no_wake` rely on the `None` submodels. It'll be slower, but still very manageable I would think.

bayc:
@Bartdoekemeijer @rafmudaf @paulf81 I moved the no_wake check down to the steady_state_atmospheric_condition level so that is no_wake is true, only the wake solver call is bypassed. This was the easiest fix I could identify in the current structure, and doesn't require us initializing the unsorted values to anything.

Another option I had mind was to create a `steady_state_no_wake_condition` method, which would be the same as `steady_state_atmospheric_condition` but without the call to the wake solver. This method would live in `floris.py` and be wrapped by a method in `floris_interface` that would be called something like `calculate_no_wake` which would eliminate the flag altogether. Not sure which one is better, but would be interested in your feedback.

bayc:
Alright, talked with Raf, changes incoming shortly, hold please.

bayc:
Ok, we thought that a dedicated no wake calculation method would be the better way to go, maintaining the "library" like nature of the simulation directory. @Bartdoekemeijer @paulf81 Please check out to see if you agree with this functionality, where if you want to calculate a no wake condition, you would call `fi.calculate_no_wake()`.

Note: there are areas in `tools/scipy/optimization/yaw_optimization` that require further updating to use the latest uncertainty interface, thus I did not make the necessary no wake changes there. This would best be captured in an additional pull request.

Bartdoekemeijer:
@bayc @rafmudaf I think the separate `calculate_no_wake` function is a good solution! Let's merge this in!

@bayc can you message me separately with more information on the work that's needed for `UncertaintyInterface` and `tools/optimization/yaw_optimization`?

rafmudaf (approved):

rafmudaf:
The failing test here is the Jupiter notebook in the new example-check action. This pull request fixes another test that was previously failing and is now passing.

Issue #422, opened 2022-04-16
Documentation on how to get the wind speed at each turbine


carlorop:
**Is your feature request related to a problem? Please describe.**
The documentation has diminished in FLORIS v 3.0. Some features can be understood easily reading the code or going through the examples, however it would be convenient to expand the documentation. For instance, explaining hot to get the wind speed at each turbine 

**Additional context**
In FLORIS 2.4. the code for getting such velocity would be:
```python
velocity_at_each_turbine = []
for i in range(len(self.fi.floris.farm.flow_field.turbine_map.turbines)):
    velocity_at_each_turbine.append(self.fi.floris.farm.turbines[i].average_velocity)
```

paulf81:
Thanks very much for this feedback!  I've created pull request #423 which adds a new example for getting wind speeds.  Does this work?  More generally we're hoping to add much more documentation going forward.

PR #423, opened 2022-04-18
Add example of getting wind speeds at turbines

paulf81:
Yes

This adds a new example file that demonstrates two ways of getting the wind speeds at each turbine

Issue #422


Bartdoekemeijer:
The example looks good to me. I just wonder whether we need this to be a separate example. Perhaps we can include this in an existing example, or just limit it to the documentation?

paulf81:
Maybe a separate example for now (so we have something) and then one day this is all in the documentation?  What do you think @rafmudaf and @bayc ?  I had the impression the new documentation is a big lift and will take some time?

bayc (approved):

paulf81:
Thanks @bayc !

Issue #421, opened 2022-04-14
Pyoptspare is difficult to set up on windows


paulf81:
**Bug description**
We got some feedback that pyopstspare is diffucult to setup in Windows.  Catching that feedback into an issue report for tracking.  There is some info about the issue:

Note about the issue:  https://mdolab-pyoptsparse.readthedocs-hosted.com/en/latest/install.html
Proposed workaround: https://github.com/mdolab/pyoptsparse/issues/273

**Proposal**
We're currently considering modularizing layout optimization as yaw optimization is modularized and bringing back support for scipy (more easily available) as an alternative to pyoptspare.  nl-opt also being considered

paulf81:
I believe the new modularization of optimizations will address this, but another idea I have @bayc  and @rafmudaf  is to have recommended install instructions separate for linux/mac/windows, does that make sense?

Gcstk:
I think it's very necessary to provide a pyoptsparse installation tutorial for Windows users. After a day's work, it still hasn't been installed successfully.

gbarter:
If it helps, on the WISDEM/WEIS projects we used to only encourage installation of `pyoptsparse` on Linux/Mac.  Now, it is conda-installable on all platforms, a somewhat recent improvement.  Installing from source on Windows is still difficult and not for the faint of heart.

We have also been very happy with the `nlopt` library of optimizations.  There are many new algorithms to explore and even for optimizers that are also found in Scipy or `pyoptsparse`, the `nlopt` implementations are more robust and reliable.

rafmudaf:
@bayc per Garrett's comment above it seems that this is now available on Windows via conda / conda-forge ([see docs](https://mdolab-pyoptsparse.readthedocs-hosted.com/en/latest/install.html#install)). Can you confirm that this works with FLORIS's optimization module? If so, then I think we can close with issue.

misi9170:
I am going to go ahead and close this issue as stale. While we do still have a pyoptsparse-based layout optimization, we have other layout optimization options that are viable alternatives.

PR #429, opened 2022-04-29
Layout optimization refactoring

bayc:
This PR looks to create a standardized interface for layout optimizations, allowing a user to either use a pre-built optimization provided within FLORIS, or implement a custom optimizer.

#421 

Added layout optimization directory to `tools/optimization`.

I am working to include a class for the boundary grid method as well.

None.


floris/tools/optimization/__init__.py
```diff
@@ -1 +1 @@
-from . import other, scipy, pyoptsparse, yaw_optimization
+from . import other, scipy, pyoptsparse, yaw_optimization, layout_optimization
```
Bartdoekemeijer, line 1: Would we want to remove the `scipy` and `pyoptsparse` folder with this pull request, assuming `yaw_optimization` and `layout_optimization` fully replace those functionalities?

floris/tools/optimization/layout_optimization/layout_optimization_base.py
```diff
@@ -0,0 +1,83 @@
...
+
+import numpy as np
+import matplotlib.pyplot as plt
+from shapely.geometry import Polygon, LineString
+
+from ....logging_manager import LoggerBase
+
+class LayoutOptimization(LoggerBase):
+    def __init__(self, fi, boundaries, min_dist=None, freq=None):
+        self.fi = fi
```
Bartdoekemeijer, line 23: Do we want this to be `self.fi = fi.copy()` to be able to mutate the object within the optimizer without affecting the local floris object?

floris/tools/optimization/layout_optimization/layout_optimization_base.py
```diff
@@ -0,0 +1,83 @@
...
+        self.boundaries = boundaries
+
+        self.boundary_polygon = Polygon(self.boundaries)
```
Bartdoekemeijer, line 26: Since these are derivatives of the public `boundaries` variable, should `self.boundary_polygon` be private, something like `self._boundary_polygon`?

floris/tools/optimization/layout_optimization/layout_optimization_base.py
```diff
@@ -0,0 +1,83 @@
...
+        self.boundary_line = LineString(self.boundaries)
```
Bartdoekemeijer, line 27: Same, should this be `self._boundary_line`?

floris/tools/optimization/layout_optimization/layout_optimization_base.py
```diff
@@ -0,0 +1,83 @@
...
+
+        self.xmin = np.min([tup[0] for tup in boundaries])
+        self.xmax = np.max([tup[0] for tup in boundaries])
+        self.ymin = np.min([tup[1] for tup in boundaries])
+        self.ymax = np.max([tup[1] for tup in boundaries])
+        self.x0 = self._norm(self.fi.layout_x, self.xmin, self.xmax)
```
Bartdoekemeijer, line 33: How about having a separate variable for the initial conditions (turbine locations)? Something like `x0`, which would default to `layout_array`? This way, one can easily try different initial conditions without having to manipulate the `floris` object.

floris/tools/optimization/layout_optimization/layout_optimization_base.py
```diff
@@ -0,0 +1,83 @@
...
+        self.y0 = self._norm(self.fi.layout_y, self.ymin, self.ymax)
+
+        if min_dist is None:
+            self.min_dist = 2 * self.rotor_diameter
+        else:
+            self.min_dist = min_dist
+
+        if freq is None:
+            self.freq = 1
```
Bartdoekemeijer, line 42: Perhaps put this into the right dimensions right away? `np.ones(shape) / np.sum(np.ones(shape))`

floris/tools/optimization/layout_optimization/layout_optimization_base.py
```diff
@@ -0,0 +1,83 @@
...
+
+        Returns:
+            nturbs (int): The number of turbines in the FLORIS object.
+        """
+        self._nturbs = self.fi.floris.farm.n_turbines
+        return self._nturbs
+
+    @property
+    def rotor_diameter(self):
+        return self.fi.floris.farm.rotor_diameters[0][0][0]
```
Bartdoekemeijer, line 83: This can potentially be confusing if you have turbines with different rotor diameters. Perhaps omit as a property or generalize?

floris/tools/optimization/layout_optimization/layout_optimization_pyoptsparse.py
```diff
@@ -0,0 +1,200 @@
...
+            raise ImportError(err_msg)
+
+        # Insantiate ptOptSparse optimization object with name and objective function
+        self.optProb = pyoptsparse.Optimization('layout', self._obj_func)
+
+        self.optProb = self.add_var_group(self.optProb)
+        self.optProb = self.add_con_group(self.optProb)
+        self.optProb.addObj("obj")
+
+        if solver is not None:
```
Bartdoekemeijer, line 55: The way it's defined now, the following situation can occur:
1. You initialize the optimization class with `solver="Nelder-Mead"` and certain `optOptions`.
2. You want to try another set of optimization settings, `newOptions`, so you do `opt._reinitialize(optOptions=newOptions)`.
3. With (2), you have unknowingly overwritten the solver to `SLSQP`.

To prevent this, I would just rename `_reinitialize` to `_initialize`, and assume you only call this function once -- when the class is created. If the user wants to try different settings, they just recreate the optimization class from scratch with `opt = LayoutOptimizationPyOptSparse(...)`.

floris/tools/optimization/layout_optimization/layout_optimization_pyoptsparse.py
```diff
@@ -0,0 +1,200 @@
...
+        self.parse_opt_vars(varDict)
+
+        # Update turbine map with turbince locations
+        self.fi.reinitialize(layout=[self.x, self.y])
+        self.fi.calculate_wake()
+
+        # Compute the objective function
+        funcs = {}
+        funcs["obj"] = (
+            -1 * np.sum(self.fi.get_farm_power() * self.freq * 8760) / self.initial_AEP
```
Bartdoekemeijer, line 89: Could you use `fi.get_farm_AEP()` for this operation? This would simplify the calculation. You could even consider adding the `cut_in_wind_speed` and `cut_out_wind_speed` options that are in `fi.get_farm_AEP` to allow the user to constrain the wind speeds that are actually evaluated in the optimization.

floris/tools/optimization/layout_optimization/layout_optimization_pyoptsparse.py
```diff
@@ -0,0 +1,200 @@
...
+        )
+
+        # Compute constraints, if any are defined for the optimization
+        funcs = self.compute_cons(funcs, self.x, self.y)
+
+        fail = False
```
Bartdoekemeijer, line 95: What is the purpose of `fail`? Just a requirement for PyOptSparse?

floris/tools/optimization/layout_optimization/layout_optimization_scipy.py
```diff
@@ -0,0 +1,316 @@
...
+            "fun": lambda x: self._distance_from_boundaries(x),
+        }
+
+        self.cons = [tmp1, tmp2]
+
+    def _set_opt_bounds(self):
+        self.bnds = [(0.0, 1.0) for _ in range(2 * self.nturbs)]
+
+    def _space_constraint(self, x_in):
+        rho=500
```
Bartdoekemeijer, line 105: Why is `rho` hardcoded here?

floris/tools/optimization/layout_optimization/layout_optimization_scipy.py
```diff
@@ -0,0 +1,316 @@
...
+            self.x0,
+            method=self.solver,
+            bounds=self.bnds,
+            constraints=self.cons,
+            options=self.optOptions,
+        )
+
+        return self.residual_plant.x
+
+    def _obj_func(self, locs):
```
Bartdoekemeijer, line 64: Is the objective function different per optimizer? Could (part of) it be moved to the base class?

floris/tools/optimization/layout_optimization/layout_optimization_scipy.py
```diff
@@ -0,0 +1,316 @@
...
+        locs_unnorm = [
+            self._unnorm(valx, self.bndx_min, self.bndx_max)
+            for valx in locs[0 : self.nturbs]
+        ] + [
+            self._unnorm(valy, self.bndy_min, self.bndy_max)
+            for valy in locs[self.nturbs : 2 * self.nturbs]
+        ]
+        self._change_coordinates(locs_unnorm)
+        self.fi.calculate_wake()
+        AEP_sum = np.sum(self.fi.get_farm_power() * self.freq * 8760)
```
Bartdoekemeijer, line 74: see comment above: consider replacing with `fi.get_farm_AEP`?

floris/tools/optimization/layout_optimization/layout_optimization_scipy.py
```diff
@@ -0,0 +1,316 @@
...
+            ],
+            [
+                self._unnorm(valy, self.bndy_min, self.bndy_max)
+                for valy in opt_locs_norm[self.nturbs : 2 * self.nturbs]
+            ],
+        ]
+
+        return opt_locs
+
+    def reinitialize_opt(
```
Bartdoekemeijer, line 189: Not sure if a `reinitialize_opt` function is necessary. Users can just reinitialize the entire class.

floris/tools/optimization/layout_optimization/layout_optimization_scipy.py
```diff
@@ -0,0 +1,316 @@
...
+        else:
+            self._set_opt_bounds()
+        if solver is not None:
+            self.solver = solver
+        if optOptions is not None:
+            self.optOptions = optOptions
+
+        self._generate_constraints()
+
+    def plot_layout_opt_results(self):
```
Bartdoekemeijer, line 267: Can this be moved to base class?

Bartdoekemeijer (commented):
Thanks for doing this, @bayc! I like the structure and simplicity of these new classes. I added comments throughout the code. Also, do you have an example script that I could run with these new classes?

floris/tools/optimization/layout_optimization/layout_optimization_scipy.py
bayc: The objectives are not different in terms of the value they compute, but the format of design variables and what is returned by the objective function is different for pyoptsparse and scipy, so that is the reason for them being specified within the optimizer classes.

floris/tools/optimization/layout_optimization/layout_optimization_base.py
bayc: I spent some time thinking about this, and decided to remove the option to specify x0 altogether and have it just be taken from the floris interface. If we allowed the users to change the initial conditions of the optimization directly within the object, we would need a  `reinitialize_opt` function, which then we get in the same issue of  needing to make sure everything gets updated properly (positions get changed in the `fi` object, the initial AEP is re-calculated, etc.). Also, it doesn't require the user to know whether they should input the `x0` as normalized or unnormalized, and the different optimization packages (Scipy vs pyOptSparse) require `x0` in different formats anyway. We could add this as a feature in the future, but it would require a significnat overhaul of the optimization classes and would need some more thought.

floris/tools/optimization/layout_optimization/layout_optimization_base.py
bayc: Updated!

floris/tools/optimization/layout_optimization/layout_optimization_base.py
bayc: Updated!

floris/tools/optimization/layout_optimization/layout_optimization_base.py
bayc: Updated!

floris/tools/optimization/layout_optimization/layout_optimization_scipy.py
bayc: Updated!

floris/tools/optimization/layout_optimization/layout_optimization_scipy.py
bayc: Updated!

floris/tools/optimization/layout_optimization/layout_optimization_scipy.py
bayc: Agreed, it was more a function from v2, but with the new mantra in v3 of being more explicit, I think a full reinitialization is correct.

floris/tools/optimization/layout_optimization/layout_optimization_scipy.py
bayc: Updated!

floris/tools/optimization/__init__.py
bayc: > @bayc does this pull request work with existing examples or require new ones?

Just added and updated example @paulf81 !

bayc:
> @bayc does this pull request work with existing examples or require new ones?

Just added and updated example @paulf81 !

floris/tools/optimization/__init__.py
bayc: Not all of the functionality has been ported over, so I have moved the folders into a legacy folder until we can finish bringing everything over.

floris/tools/optimization/layout_optimization/layout_optimization_base.py
bayc: Updated!

floris/tools/optimization/layout_optimization/layout_optimization_base.py
bayc: This is only used in the case that a minimum turbine spacing is not specified. If the user is doing layout optimization with multiple turbine types, it will be upon them to explicitly define the minimum spacing constraint.

floris/tools/optimization/layout_optimization/layout_optimization_pyoptsparse.py
bayc: Exactly, and in this case just a pass-through for us.

floris/tools/optimization/layout_optimization/layout_optimization_pyoptsparse.py
bayc: Updated to `fi.get_farm_AEP()`. I will need to think through passing in the cut in and cut out wind speeds more, as they would need defaults, and right now `floris_interface` doesn't handle a `None` for the cut in wind speed.

floris/tools/optimization/layout_optimization/layout_optimization_pyoptsparse.py
bayc: Ah, good catch! I have just removed the `_reinitialize` class altogether and merged the required functions with the `__init__` method.

paulf81:
Hi @bayc , I hope it is ok, I added a little more output to your example, just to check my understanding and also to show how to see power improvement.  Could last additional thoughts:
- Could you step through the changes quickly at today's meeting?
- Do you think it would be good to include a 11b example which does the same thing using pyoptsparse?

paulf81 (approved):

Issue #431, opened 2022-05-12
Feature request: clear error message when calculate_wake is forgotten


Bartdoekemeijer:
**Is your feature request related to a problem? Please describe.**
Feature request.

**Describe the solution you'd like**
Requesting a clear error message when the user forgets to call `calculate_wake` before calling `get_farm_power` or `get_turbine_powers` in the `FlorisInterface` class.

**Describe alternatives you've considered**
N/A

**Additional context**
This the current error message the user gets:
```
Exception has occurred: AttributeError       (note: full exception trace is shown but execution is paused at: <module>)
'Farm' object has no attribute 'turbine_type_map'
  File "/home/bartdoekemeijer/python_scripts/floris/floris/tools/floris_interface.py", line 573, in get_turbine_powers
    turbine_type_map=self.floris.farm.turbine_type_map,
  File "/home/bartdoekemeijer/python_scripts/floris/examples/04_sweep_wind_directions.py", line 61, in <module> (Current frame)
    turbine_powers = fi.get_turbine_powers() / 1E3 # In kW
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
```
Which can take a while to trace back to the actual problem.

paulf81:
I made #432 to take a pass, what do you think?

paulf81:
Clicked the wrong button, I'll try that out thanks Bart!

rafmudaf:
This has been addressed in #432.

Given this script:

```python
from floris.tools import FlorisInterface
fi = FlorisInterface("inputs/gch.yaml")
# fi.calculate_wake()
turbine_powers = fi.get_turbine_powers()
```

The following error message is given:
```
>>$ python 01_opening_floris_computing_power.py 
Traceback (most recent call last):
  File "removed /floris/examples/01_opening_floris_computing_power.py", line 9, in <module>
    turbine_powers = fi.get_turbine_powers()
  File "removed /floris/floris/tools/floris_interface.py", line 594, in get_turbine_powers
    raise RuntimeError(
RuntimeError: Can't run function `FlorisInterface.get_turbine_powers` without first running `FlorisInterface.calculate_wake`.
```

PR #432, opened 2022-05-12
Issue alert if calc wake needs to be run

paulf81:
Just a first pass at an idea to address Issue #431

@Bartdoekemeijer  what do you think?

Check a certain attribute exists to determine if calc wake needs to be run

Issue #431


examples/make_error.py
```diff
@@ -0,0 +1,15 @@
+
```
Bartdoekemeijer, line 1: This is nice to demonstrate, but we should remember to remove this example file before merging

floris/tools/floris_interface.py
```diff
@@ -558,12 +558,20 @@ def check_wind_condition_for_viz(self, wd=None, ws=None):
         if len(ws) > 1 or len(ws) < 1:
             raise ValueError("Wind speed input must be of length 1 for visualization. Current length is {}.".format(len(ws)))
 
+    def check_calc_wake_run(self, func_name):
```
Bartdoekemeijer, line 561: Maybe make this private, `_check_calc_wake_run`?

Bartdoekemeijer, line 561: Maybe make this private, `_check_calc_wake_run`?

floris/tools/floris_interface.py
```diff
@@ -558,12 +558,20 @@ def check_wind_condition_for_viz(self, wd=None, ws=None):
...
+            raise AttributeError('Cant call function %s without first calling calculate_wake' % func_name)
+
     def get_turbine_powers(self) -> NDArrayFloat:
         """Calculates the power at each turbine in the windfarm.
 
         Returns:
             NDArrayFloat: [description]
         """
+
+        self.check_calc_wake_run('get_turbine_powers')
```
Bartdoekemeijer, line 573: Nice! We also need to add it to `get_farm_power`, and maybe some of the visualization functions?

Bartdoekemeijer, line 573: Nice! We also need to add it to `get_farm_power`, and maybe some of the visualization functions?

floris/tools/floris_interface.py
paulf81: Good suggestion, changed it

paulf81: Changed

RHammond2:
@rafmudaf, I'm not totally convinced on it, though not opposed either, but I've also never quite found `Enum`s to be useful in my own use cases, so I could be biased.

rafmudaf:
> I've also never quite found Enums to be useful in my own use cases

I'm not tied to the `Enum` type either, but mostly wondering about the idea of having a `State` attribute

floris/tools/floris_interface.py
```diff
@@ -631,6 +639,10 @@ def get_farm_power(
         # for turbine in self.floris.farm.turbines:
         #     turbine.use_turbulence_correction = use_turbulence_correction
 
+        # Confirm calculate wake has been run
+        if self.floris.state is not State.USED:
```
rafmudaf, line 643: @paulf81 I think we do not need this here since it would be caught by `self.get_turbine_powers()`. Do you agree?

paulf81: that makes sense to me, maybe this could be deleted and then we can make a test we get an error if calc wake not run before get_turbine_powers?

PR #435, opened 2022-05-26
Feature: Add turbine_weights option in get_farm_power and get_farm_aep

Bartdoekemeijer:
This PR **is** ready to be merged.

This PR adds an option called `turbine_weights` with which users can weigh individual turbine's power production when doing AEP and farm power calculations in `floris_interface` and `uncertainty_interface`. This is useful when analyzing multiple wind farms. For example, one may want to model neighboring wind farms in FLORIS but exclude their power production when doing AEP calculations. This is already an available feature in the `yaw_optimization` class but was not yet available more generally.

N/A.

`FlorisInterface`, `UncertaintyInterface`

N/A

A couple quick test cases with `07_calc_aep_from_rose.py` indicates that the code works as intended.


Bartdoekemeijer:
Reminder that I'll add an example for this PR with 2 farms, weighing only one of the two farms. Also show yaw optimization and calculate farm AEP.

Bartdoekemeijer:
I've pushed up a preliminary example. I think it works as expected, but the effect of the neighboring turbines is pretty small, to the point where one may ask whether it's even necessary to include neighboring farms in FLORIS. Perhaps @paulf81 can check it out?

paulf81:
@bayc and @Bartdoekemeijer  this looks great to me, I just added one more example I made.  I made it both so I can make sure I understood how to use everything / debugging, and then also I thought it could be useful to show how to use the new feature outside the yaw opt context

paulf81 (approved):

PR #437, opened 2022-05-27
Clean up and simplify some things

rafmudaf:
The biggest change here is that I've removed the wake model `model_string` attribute. This was left over from v2, and in v3 we map model strings directly to their model definitions in the Wake class. While FLORIS doesn't use this attribute internally, @paulf81 @pjstanle is this something you or others use in your scripts?

Otherwise, this pull request removes the outdated Curl model class and other code and import-statements that aren't used.

None, just spring cleaning.

Wake model definitions, most files in `floris.simulation`, and `floris_interface` imports


paulf81 (approved):
Hi @rafmudaf  this sounds good, I wasn't using the model_string myself

PR #440, opened 2022-06-07
Update 07_calc_aep_from_rose.py

kilojoules:
Get rotor diameter from input file instead of hard coding it


rafmudaf (approved):
Thanks @kilojoules

Issue #441, opened 2022-06-21
Feature request: support heterogeneous wind direction inputs


pjireland:
**Is your feature request related to a problem? Please describe.**
I would like to be able to use FLORIS to simulate the waking behavior where both ambient wind speeds and ambient wind directions have spatial variations over the flow domain. Is this currently supported in FLORIS 3.1?

**Describe the solution you'd like**
With FLORIS 2.4, I used an interface like this:
```
FlorisInterface.reinitialize_flow_field(
    wind_speeds=..., # wind speed measurements for the points in wind_layout
    wind_directions=..., # wind direction measurements for the points in wind_layout
    ...
    wind_layout=..., # specify the locations where the wind measurements are taken
    layout_array=..., # specify the locations where the turbines are located in the flow
)
```
I'm not seeing a similar interface in FLORIS 3.1.

**Describe alternatives you've considered**
1. I looked into the `reinitialize` call, but this only supports the `layout` parameter, which seems to be the coordinate locations of the turbines. In addition, from what I understand, each wind speed and wind direction is assumed to apply to the entire flow domain, and it will run for all possible combinations of these.

2. I saw that you can specify a `het_map` parameter when instantiating a `FlorisInterface` object. However, this seems to apply only to wind speeds and not wind directions. Furthermore, I don't see a way to modify this after an object has been instantiated.

bayc:
You are correct, we don't currently have a similar infeterface in FLORIS 3.1 to what we had in FLORIS 2.4. We do support heterogeneous wind speeds, as you noted, but not currently wind directions. With the vectorization and speed advancements of the code, the heterogeneous wind direction functionality adds additional complexity beyond the heterogeneous wind speeds. Additionally, the development team was aware of users using the het. wind speeds, but hadn't heard of folks using the het. wind directions. For these reasons, we decided to prioritze wind speeds first. Het. wind directions could possibly be added, but we don't currently have a clear road map for that. Is that something you would be interested in helping with?

To confirm/clarify your alternatives:

1. `reinitialize` does support the layout parameter, as well as updates to the wind directions and wind speeds. Currently, the wind directions and wind speeds are expanded with one another to represent a wind rose type of calculation. There is a feature implemented in the develop branch currently that allows the user to set a `time_series` flag in the `reinitialize` call that requires the wind direction and wind speed arrays to be of the same length, in which the code will then compute each wind condition specified without expanded the two arrays. This will be merged into the main branch soon in a v3.2 update.
[...]

PR #453, opened 2022-06-23
Support heterogeneous wind direction definition via FlorisInterface.reinitialize

pjireland:
This PR provides the ability to pass `het_map` to `reinitialize` to allow for different hetereogeneous wind speeds than were used when initializing the `FlorisInterface` object.

https://github.com/NREL/floris/issues/441#issuecomment-1164651453

`floris.tools.floris_interface.reinitialize`

I've tested this on a simple case and it worked for me. One thing that I haven't tested, however is the possible issue raised in [this discussion](https://github.com/NREL/floris/issues/441#issuecomment-1164883144):
> One of the intricacies for `het_map` is that you need to define speed ups for each wind direction, so if you change the number of wind directions, you need to also generate a new `het_map` interpolant that matches that dimension, so that is one catch that we would want to check for potentially.


rafmudaf:
TODO:
- [ ] Merge `develop`
- [ ] Verify that this issue still exists

paulf81:
I added a new (maybe temporary) example that shows how to use the new option to change the number of wind directions in code with a het map

paulf81:
TODO:

- [ ] Add a better error when wd doesn't match het map dimensions
- [ ] Add a test for this

rafmudaf:
Please add this example to the Example listing: https://github.com/NREL/floris/blob/develop/docs/examples.md

floris/tools/floris_interface.py
```diff
@@ -195,6 +195,7 @@ def reinitialize(
         solver_settings: dict | None = None,
         time_series: bool | None = False,
         layout: tuple[list[float], list[float]] | tuple[NDArrayFloat, NDArrayFloat] | None = None,
+        het_map=None,
```
rafmudaf, line 198: It would be ideal to maintain the type hinting here but the heterogeneous map is difficult to express. This is a good candidate for a class defining the heterogeneous map and providing a good interface for interacting with it.

floris/simulation/flow_field.py
```diff
@@ -150,6 +150,12 @@ def finalize(self, unsorted_indices):
         self.w = np.take_along_axis(self.w_sorted, unsorted_indices, axis=2)
 
     def calculate_speed_ups(self, het_map, x, y, z=None):
+
+        if x.shape[0] != np.array(het_map).shape[1]: # Check that the het maps wd dimension matches
```
rafmudaf, line 154: Consider user `self.n_wind_directions` here rather than `x.shape[0]` since it is more explicit.

paulf81: Done!

rafmudaf (approved):

PR #477, opened 2022-08-01
Support controllable turbine Ct

genevievestarke:
This feature allows the turbines to be operated below maximum power through setting the thrust coefficients (similar to setting of yaw) for each turbine

farm.py
floris.py
solver.py
floris_interface.py
turbine.py

An example of running it with adjusted thrust coefficient values:
```
# Convert to a simple two turbine layout
fi.reinitialize( layout=( [0, 500.], [0., 0.] ) )

# Single wind speed and wind direction

# Get the turbine powers assuming 1 wind speed and 1 wind direction
fi.reinitialize(wind_directions=[270.], wind_speeds=[8.0])

# Set the yaw angles and C_T values
yaw_angles = np.zeros([1,1,2]) # 1 wind direction, 1 wind speed, 2 turbines
CT_input = np.zeros([1,1,2])+0.75 # 1 wind direction, 1 wind speed, 2 turbines
fi.calculate_wake(yaw_angles=yaw_angles, CT_inputs=CT_input)
```


Halation666:
Hi,This function is very useful to me. Has he succeeded now?Thx!

aallpp83:
Dear @genevievestarke,

Thank you for your helpful suggestion. I am currently working on simulating a wind farm with variable Cp/Ct characteristics for each wind turbine, and I found your modification to be useful. However, I am struggling to understand the logic behind the power calculation part (def power() in turbine.py). Specifically, I am wondering why we should use Cp = 0.77*ratio*4ai*(1-ai)^2 instead of the simple Cp = 4*ai*(1-ai)^2 calculation (that differs from the predefined table of a specific turbine). Could you provide a reference for your calculation? Additionally, did you account for the effect of air density in the final power relation?

Moreover, I am not familiar with the mechanics of the FLORIS code, and I am curious if your modification takes into account the effect of varying Ct on the wake characteristics.

Thank you in advance for your assistance and answers.

Best regards,

Ali

misi9170:
@genevievestarke @rafmudaf @paulf81 

Based on discussions today, I'm tagging this 4.1. However, we may use derating as an example for the turbine operation module (alongside helix), so this might end up in 4.0. Feel free to change the tag to 4.0 if you'd prefer.

misi9170:
@genevievestarke I believe that this functionality has now been incorporated into v4 with #783. Rather than specifying the thrust coefficient, users specify the maximum power setpoint, and the thrust coefficient and axial induction are back-computed from that (thrust is assumed to be at the same fraction of available as power).

misi9170:
After conferring with @genevievestarke , I'm closing this pull request as equivalent functionality is now available in the `SimpleDeratingTurbine` in FLORIS v4.

PR #493, opened 2022-08-31
adding missing time_series flag to cc flow field solver call

bayc:
This is a small bugfix to add a missing `time_series` flag to the CC model full flow field solver. It is pointed at the develop branch as the time series functionality is not yet in the main branch.

None.

`solver.py`

None.

Can now run example 02 and visualize the CC model, when previously it would error out.


paulf81 (approved):
This change makes sense to me!  Thanks for catching the bug!

PR #494, opened 2022-09-02
Performance Improvements and Code Simplification

RHammond2:
This pull request includes some minor to moderate speedups in the following models, with the likely biggest speedups occurring for a GCH analysis:
- Gauss wake deflection
- Crespo Hernandez wake turbulence
- Gauss wake velocity
- Jensen wake velocity
- TurbOPark wake velocity

The methods used to accomplish this include:
- changing `np.array(boolean_operation)` to `boolean_operation` to reduce the number of array creations
- reducing the number of stored variables, especially for single-use variables
- variable reassignment
- chunking large methods into multiple methods, especially those where the code is repeated (including calculations with minor variations)
- more usage of the NumExpr library for large operations
- list comprehensions in favor of loops and nested loops

This impacts a large portion of the code base, especially given the use of the pre-commit workflow, which seems to have been turned off somewhere along the way.


misi9170 (commented):
Looks good to me. On a small AEP calculation, runs faster than the develop version with no change in the resulting AEP.

paulf81:
Thanks for testing @misi9170 and thanks for this @RHammond2 !  

I also pulled down to my machine and just ran a few examples and didn't observe any issues so good to go from my side!

paulf81 (approved):

RHammond2:
Thanks for giving these a review, @misi9170 and @paulf81 ! @rafmudaf it seems like this is ready for a proper review whenever you have availability.

RHammond2:
@rafmudaf I fixed up a merge conflict, and resolved the following based on our conversation:
- Where `black` changed masking from `variable[i:i+1]` to `variable[i : i + 1]`, I've put it back
- I made all the exponents consistent to ` ** ` throughout the code 
- Any method that `black` put on multiple lines, I've reformatted to be of our previously discussed format of a single parameter per line

This should make the diff much easier to follow and stick with the expected formatting (if only for a handful of undiscussed or missed changes)

RHammond2:
@rafmudaf  this is up to date with the latest formatting changes. The only thing not passing the linter is a file that should be deleted before merging but shows some of the changes.

floris/simulation/wake_deflection/gauss.py
```diff
@@ -267,82 +362,46 @@ def wake_added_yaw(
...
-    # NOTE: this is the top of the grid, not the top of the rotor
-    zT = z_i - (HH + D / 2) + BaseModel.NUM_EPS  # distance from the top of the grid
-    rT = yLocs ** 2 + zT ** 2  # TODO: This is - in the paper
-    # This looks like spanwise decay;
-    # it defines the vortex profile in the spanwise directions
-    core_shape = 1 - np.exp(-rT / (eps ** 2))
-    v_top = (Gamma_top * zT) / (2 * np.pi * rT) * core_shape
-    v_top = np.mean( v_top, axis=(3,4) )
-    # w_top = (-1 * Gamma_top * yLocs) / (2 * np.pi * rT) * core_shape * decay
+    v_top, _ = _calculate_vortex(z_i, HH, D, yLocs, eps, Gamma_top, which="top")
```
rafmudaf, line 390: @RHammond2 I see and appreciate your intent here. In the original code, there's a lot of repeat and it can clearly be consolidated into a function. However, I find that readability is reduced, in this particular case. In comparing to the paper, you can follow along with what the original code is calculating, but as a reader you have to parse the new function and map it back to the variables in the calling code. Additionally, it was really helpful to print and plot different terms from each vortex calculation during development, so it would be good to preserve that functionality for future debugging. Instead of consolidation to a function, I propose to incorporate numexpr into all of the terms. I'm happy to make those changes, but do you agree with that suggestion?

RHammond2: Thanks for the feedback, @rafmudaf! I'm totally fine with switching this back given that it helps align the methods with the paper that most users would be familiar with. If you're feeling particularly motivated, then you're more than welcome to put this back in place, otherwise, I'm ok with switching it back myself.

floris/simulation/wake_deflection/gauss.py
```diff
@@ -373,145 +432,116 @@ def calculate_transverse_velocity(
...
+        yLocs,
+        eps,
+        Gamma_wake_rotation,
+        decay=decay,
+        which="rotation",
+        with_decay=True,
+        ground=True,
+    )
+
+    V = np.add.reduce((V1, V2, V3, V4, V5, V6))
```
rafmudaf, line 506: @RHammond2 Can you explain the benefit us using `np.add.reduce` with the array sequence instead of `np.add` or the np.array `+` operator?

RHammond2: At this point, I can't remember why that's there. My hunch is that it was faster, or more convenient for use case. However, after testing out the snippet on small and large arrays, I found it's quite a bit slower compared to `V1 + V2 + V3 + V4 + V5 + V6`. The only reason to use `np.add.reduce` instead of `np.add` is to ensure the addition is handled consistently.

My vote is to put it back to original arithmetic, and if whatever I was attempting to do at that point comes back to me, it can be handled then.

rafmudaf: Ok cool, thanks for following up. I'll revert that now but we can keep it on our radar for the future

rafmudaf:
@RHammond2 I removed a few more places where `np.array` is redundant. Could you please review this commit: https://github.com/NREL/floris/pull/494/commits/879e613f3fd4c940edb6a4103819f09f831b9015?

rafmudaf (approved):

rafmudaf:
I've also timed a 3-turbine array on Eagle, and we see pretty good results:

| Commit hash        | Jensen      | Gauss  | GCH    | CC     |
| ------------------ | ----------- | ------ | ------ | ------ |
| 16628a0b (v3.3)    | 0.2694      | 0.9297 | 1.4587 | 1.6214 |
| 5e5bb7f4 (v3.4)     | 0.2553      | 0.8701 | 1.4324 | 1.5948 |
| c7a3912e (This PR)  | 0.2915      | 0.8588 | 1.0687 | 1.2973 |

This pull request speeds up GCH by about 25% and CC by about 18%.

RHammond2:
> @RHammond2 I removed a few more places where `np.array` is redundant. Could you please review this commit: [879e613](https://github.com/NREL/floris/commit/879e613f3fd4c940edb6a4103819f09f831b9015)?

@rafmudaf this looks good to me. Basically as long as one of the items is an array, the result is always going to be an array.

rafmudaf:
Ok got ya. I made note of that in the [Proposed Style Guide Discussion](https://github.com/NREL/floris/discussions/292). The other changes here are less syntactic and more algorithmic, but this one seems easy to enforce more broadly.

PR #505, opened 2022-09-27
Speed up angle wrap functions and improve API

paulf81:
Is ready to merge

In some other work, found these implementations of the wrap functions which are 2/3x faster then what we have now.  Useful in codes that call this like flasc.  Also, using modulus allows a broader range of inputs to be wrapped into range

One small change is new version of 180 wrap is [-180,180) instead of (-180,180] as previously implemented.   But this is more consistent with the [0,360) wrap I think.  Corrected the unit test for this and all other tests pass.

Ran timing tests in a notebook

image

### DIFF
--- a/floris/utilities.py
+++ b/floris/utilities.py
@@ -164,7 +164,7 @@ def wrap_180(x):
         x (numeric or np.array): Scalar value or np.array of values to shift.
 
     Returns:
-        np.array: Shifted values.
+        np.ndarray | float | int: Shifted values.
     """
 
     return ((x + 180.0) % 360.0) - 180.0
@@ -178,7 +178,7 @@ def wrap_360(x):
         x (numeric or np.array): Scalar value or np.array of values to shift.
 
     Returns:
-        np.array: Shifted values.
+        np.ndarray | float | int: Shifted values.
     """
     
     return x % 360.0
--- a/floris/utilities.py
+++ b/floris/utilities.py
@@ -166,9 +166,8 @@ def wrap_180(x):
     Returns:
         np.array: Shifted values.
     """
-    x = np.where(x <= -180.0, x + 360.0, x)
-    x = np.where(x > 180.0, x - 360.0, x)
-    return x
+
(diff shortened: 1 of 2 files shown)


floris/utilities.py
```diff
@@ -166,9 +166,8 @@ def wrap_180(x):
     Returns:
         np.array: Shifted values.
     """
-    x = np.where(x <= -180.0, x + 360.0, x)
-    x = np.where(x > 180.0, x - 360.0, x)
-    return x
+
+    return ( (x + 180.) % 360.) - 180.
```
Bartdoekemeijer, line 170: Perhaps update formatting?
`return ( (x + 180.0) % 360.0) - 180.0`

Bartdoekemeijer:
Looks great to me, nice find!

floris/utilities.py
RHammond2: I'd recommend Bart's suggestion of the `180.` -> `180.0` style changes plus dropping the extra space between the two leading parentheses.

floris/utilities.py
```diff
@@ -181,9 +180,8 @@ def wrap_360(x):
     Returns:
         np.array: Shifted values.
```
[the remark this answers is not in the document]
RHammond2: As an FYI, the updated type hint will be `np.ndarray | float | int` for the docstrings that need to be udpated

RHammond2 (approved):
I don't have much new to add here other than just a couple of small extras based on what Raf and Bart have said. I think it makes a lot of sense to have the two functions returning the same range bound types for consistency. The only concern I could see is if others are relying on this working the other way around, but that feels more likely to be an edge case.

paulf81:
Hi all, thanks for the helpful comments!  @rafmudaf and @RHammond2 , do you think it would be more seamless if I didn't change the type returned?  Say by wrapping the return in np.array or something?

paulf81:
Hi all, I think this is now good to go if you'd like to take a last look?

PR #507, opened 2022-09-30
Bug fix: cast reference_wind_height to float

Bartdoekemeijer:
Bug fix: the variable `reference_wind_height` is currently being forced into an integer, causing the warning message `WARNING:floris.tools.floris_interface.FlorisInterface:The only unique hub-height is not the equal to the specified reference wind height.  If this was unintended use -1 as the reference hub height to indicate use of hub-height as reference wind height.` whenever the turbine hub height is not a round number.

N/A

`FlorisInterface`

Note: this PR is directly towards the `main` branch since it's a bug fix. Feel free to change this to `develop` if you think it belongs there.

The error no longer appears with this hotfix.

### DIFF
--- a/floris/simulation/flow_field.py
+++ b/floris/simulation/flow_field.py
@@ -34,7 +34,7 @@ class FlowField(FromDictMixin):
     wind_shear: float = field(converter=float)
     air_density: float = field(converter=float)
     turbulence_intensity: float = field(converter=float)
-    reference_wind_height: float = field(converter=int)
+    reference_wind_height: float = field(converter=float)
     time_series : bool = field(default=False)
 
     n_wind_speeds: int = field(init=False)
--- a/floris/simulation/flow_field.py
+++ b/floris/simulation/flow_field.py
@@ -34,7 +34,7 @@ class FlowField(FromDictMixin):
     wind_shear: float = field(converter=float)
     air_density: float = field(converter=float)
     turbulence_intensity: float = field(converter=float)
-    reference_wind_height: int = field(converter=int)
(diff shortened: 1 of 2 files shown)


paulf81:
Nice catch, thank you @Bartdoekemeijer !

paulf81 (approved):

Bartdoekemeijer:
Made some changes to avoid exact comparisons with floats. Let me know if you agree. Thanks!

Bartdoekemeijer:
> > Made some changes to avoid exact comparisons with floats. Let me know if you agree. Thanks!
> 
> It looks good to me. I see the bug in the typecast in flow_field, but I don't follow why the rounding and comparison with a small value are needed in floris_interface. Comparing a float and int works as expected:
> 
> ```
> >>> 1.0==1
> True
> ```
> 
> And arithmetic between a float and int results in a float:
> 
> ```
> >>> 1.0 + 1
> 2.0
> ```
> 
> In what circumstance do these lines fail incorrectly due to their type or precision?
> 
> * https://github.com/NREL/floris/pull/507/files#diff-424cc5367d5858585a7b09e4542c79863b193114c5797d97170b05520fd85dc9L69
> * https://github.com/NREL/floris/pull/507/files#diff-424cc5367d5858585a7b09e4542c79863b193114c5797d97170b05520fd85dc9L74

I thought it was generally good practice to avoid exact comparisons with floats due to numerical precision issues. For example: https://davidamos.dev/the-right-way-to-compare-floats-in-python/ and also in my Python environment:

image

Issue #517, opened 2022-10-31
implement reference values for turbine/wind parameters


bayc:
Need to implement input structure and processing of reference values and user-defined values. These include: tilt, floater motion, potentially TI? Can implement similar to reference air density value recently added.

- [x] tilt
- [ ] floater motion
- [ ] TI?

Figure out which reference properties should be scalar, or support vectors

misi9170:
@bayc , what is the status of this one? Is this related to the multidimensional Cp/Ct table effort?

rafmudaf:
Bump @bayc

Issue #515, opened 2022-10-26
add tilt correction to Cp/Ct curves


rafmudaf:
This will use the user-defined average tilt value from the turbine definition to alter the given Cp/Ct curves if the tilt angle is not equal to the reference tilt angle.

misi9170:
@rafmudaf @bayc I believe this was addressed in #518 . Closing, but feel free to reopen.

PR #518, opened 2022-10-31
Feature: Floating turbine Cp/Ct/Tilt support

bayc:
This PR adds support for floating turbines in the form of Cp/Ct curve corrections for tilt of the floating platform based on wind speed.

#515
#517 
#526 
#527 

A lot of the simulation code, focusing mostly in `turbine.py` and `farm.py`. Updates to `floris_interface.py` as well as the pre-defined turbines.

Specific tasks involved in this PR:

- [x] Add reference tilt to inputs and `turbine.py`
- [x] Add tilt/wind speed table to inputs and `turbine.py`
- [x] Build out tilt values in `farm.py`
- [x] Update solvers to support tilt values/corrections
- [x] Re-scope power function in `turbine.py`
- [x] Add tilt angle support to `floris_interface.py`
- [x] Update all predefined turbine definitions to include reference tilt
- [x] Update unit tests
- [x] Update regression tests

Currently all tests pass. Unit tests were added for the new/rescoped functions in `turbine.py`, while other unit tests were updated appropriately for tilt inputs. Note the regression test baseline values did not change. This was intentional to show that the tilt correction code does not cause a change in calculated values when the tilt of the turbine is equal to the reference tilt of that turbine.


paulf81:
Yes @bayc I think you've captured it all nicely, this looks good!

.gitignore
```diff
@@ -32,3 +32,9 @@ examples/SLSQP.out
 
 # Log files
 *.log
+
+# User-specific files
```
rafmudaf, line 36: Are these relevant to keep or left over from local changes?

examples/inputs_floating/gch_fixed.yaml
```diff
@@ -0,0 +1,226 @@
+
```
rafmudaf, line 1: The gch.yaml input file also serves as the reference for the documentation. It might be worthwhile to remove all of the comments here for readability.
Same comment for the group of input files added here.

floris/simulation/turbine.py
```diff
@@ -319,6 +489,38 @@ def __attrs_post_init__(self) -> None:
...
+
+        if any(el.ndim > 1 for el in inputs):
+            raise ValueError("tilt and wind_speed inputs must be 1-D.")
+
+        if len({self.tilt.size, self.wind_speeds.size}) > 1:
+            raise ValueError("tilt and wind_speed tables must be the same size.")
+
+        # Remove any duplicate wind speed entries
+        _, duplicate_filter = np.unique(self.wind_speeds, return_index=True)
+        object.__setattr__(self, "tilt", self.tilt[duplicate_filter])
```
rafmudaf, line 520: Just curious about these two lines. What's different between this and the following;
```python
self.tilt = self.tilt[duplicate_filter]
```

tests/turbine_unit_test.py
```diff
@@ -373,4 +576,8 @@ def test_asdict(sample_inputs_fixture: SampleInputs):
     new_turb = Turbine.from_dict(dict1)
     dict2 = new_turb.as_dict()
 
+    print(dict1)
```
rafmudaf, line 579: Still needed?

rafmudaf (changes requested):
I see that this pull request is still a draft, but I looked through the changes and wanted to mentioned a few things. Also, please update the examples listing [here](https://github.com/NREL/floris/blob/develop/docs/examples.md) describing the new example.

.gitignore
bayc: Local changes, good catch! They are removed now.

floris/simulation/turbine.py
bayc: After talking with @RHammond2, my understanding is that this method was used when these tables were initially immutable (I modeled this after the pattern in PowerThrustTable). But sounds like the change was made along the way to have PowerThrustTable to be mutable, so I've updated to your suggestion in both tables.

tests/turbine_unit_test.py
bayc: Nope. Removed!

bayc:
> I see that this pull request is still a draft, but I looked through the changes and wanted to mentioned a few things. Also, please update the examples listing [here](https://github.com/NREL/floris/blob/develop/docs/examples.md) describing the new example.

Updated and added!

floris/simulation/farm.py
```diff
@@ -25,6 +25,7 @@
     State,
     Turbine,
 )
+from floris.simulation.turbine import _compute_tilt_angles_for_floating_turbines
```
rafmudaf, line 28: Since `_compute_tilt_angles_for_floating_turbines` is imported outside of turbine, it should not have a prefixed underscore, or there's a reason it is meant to be private it should not be imported directly here.

floris/simulation/farm.py
```diff
@@ -131,14 +135,27 @@ def construct_rotor_diameters(self):
     def construct_turbine_TSRs(self):
         self.TSRs = np.array([turb['TSR'] for turb in self.turbine_definitions])
 
-    def construc_turbine_pPs(self):
+    def construct_turbine_pPs(self):
         self.pPs = np.array([turb['pP'] for turb in self.turbine_definitions])
 
-    def construc_turbine_ref_density_cp_cts(self):
+    def construct_turbine_pTs(self):
+        self.pTs = np.array([turb['pT'] for turb in self.turbine_definitions])
```
rafmudaf, line 142: This should be listed in the attributes for this class at the top.

floris/simulation/farm.py
```diff
@@ -131,14 +135,27 @@ def construct_rotor_diameters(self):
     def construct_turbine_TSRs(self):
         self.TSRs = np.array([turb['TSR'] for turb in self.turbine_definitions])
 
-    def construc_turbine_pPs(self):
+    def construct_turbine_pPs(self):
         self.pPs = np.array([turb['pP'] for turb in self.turbine_definitions])
```
rafmudaf, line 139: This should be listed in the attributes for this class at the top.

floris/simulation/farm.py
```diff
@@ -131,14 +135,27 @@ def construct_rotor_diameters(self):
...
+    def construct_turbine_pTs(self):
+        self.pTs = np.array([turb['pT'] for turb in self.turbine_definitions])
+
+    def construct_turbine_ref_density_cp_cts(self):
         self.ref_density_cp_cts = np.array([
             turb['ref_density_cp_ct'] for turb in self.turbine_definitions
         ])
 
+    def construct_turbine_ref_tilt_cp_cts(self):
+        self.ref_tilt_cp_cts = np.array(
```
rafmudaf, line 150: This should be listed in the attributes for this class at the top.

floris/simulation/farm.py
```diff
@@ -131,14 +135,27 @@ def construct_rotor_diameters(self):
...
+            [turb['ref_tilt_cp_ct'] for turb in self.turbine_definitions]
+        )
+
+    def construct_turbine_correct_cp_ct_for_tilt(self):
+        self.correct_cp_ct_for_tilt = np.array(
```
rafmudaf, line 155: This should be listed in the attributes for this class at the top.

floris/simulation/farm.py
```diff
@@ -148,6 +165,9 @@ def construct_turbine_fCts(self):
     def construct_turbine_fCps(self):
         self.turbine_fCps = [(turb.turbine_type, turb.fCp_interp) for turb in self.turbine_map]
 
+    def construct_turbine_fTilts(self):
+        self.turbine_fTilts = [(turb.turbine_type, turb.fTilt_interp) for turb in self.turbine_map]
```
rafmudaf, line 169: This should be listed in the attributes for this class at the top.

floris/simulation/farm.py
```diff
@@ -180,11 +200,36 @@ def expand_farm_properties(
             sorted_coord_indices,
             axis=2
         )
+        self.ref_density_cp_cts_sorted = np.take_along_axis(
```
rafmudaf, line 203: Need a similar method for fTilts?

rafmudaf (changes requested):

floris/tools/floris_interface.py
```diff
@@ -630,13 +658,18 @@ def get_turbine_ais(self) -> NDArrayFloat:
...
             fCt=self.floris.farm.turbine_fCts,
+            tilt_interp=self.floris.farm.turbine_fTilts,
+            correct_cp_ct_for_tilt=self.floris.farm.correct_cp_ct_for_tilt,
             turbine_type_map=self.floris.farm.turbine_type_map,
         )
         return turbine_ais
 
     @property
     def turbine_average_velocities(self) -> NDArrayFloat:
+        # TODO: Should we have a `turbine_effective_velocities` function?
```
rafmudaf, line 672: Will this be used in analysis? If so, it would be good to have.

floris/simulation/farm.py
bayc: Updated to be a public method.

floris/simulation/farm.py
bayc: Using the pattern from Ct/power interpolants, an array of just the different interpolators is kept (one for each kind of turbine) and the turbine type is checked to know which interpolant (fTilt) to use, so don't need to sort them.

floris/tools/floris_interface.py
bayc: @rafmudaf Unsure if it would be used or not. It isn't currently, that I know of. The `turbine_average_velocities` are used in a couple of examples, but it doesn't include the effect of yaw or tilt (which would be included in the rotor effective velocity). @paulf81 What do you think?

floris/simulation/turbine.py
```diff
@@ -125,34 +257,21 @@ def power(
     # That said, it adds overhead to convert the floats to 1-D arrays, so I don't
```
rafmudaf, line 257: I think deleting the lines below this means this comment block is no longer relevant and can be deleted

floris/simulation/turbine.py
```diff
@@ -162,7 +281,11 @@ def power(
 def Ct(
     velocities: NDArrayFloat,
     yaw_angle: NDArrayFloat,
+    tilt_angle: NDArrayFloat,
```
rafmudaf, line 284: Need to add these arguments to the docstring below:
tilt_angle
ref_tilt_cp_ct
tilt_interp
correct_cp_ct_for_tilt

floris/simulation/turbine.py
```diff
@@ -210,14 +350,18 @@ def Ct(
...
     thrust_coefficient = np.clip(thrust_coefficient, 0.0001, 0.9999)
-    effective_thrust = thrust_coefficient * cosd(yaw_angle)
+    effective_thrust = thrust_coefficient * cosd(yaw_angle) * cosd(tilt_angle - ref_tilt_cp_ct)
     return effective_thrust
 
 
 def axial_induction(
     velocities: NDArrayFloat,  # (wind directions, wind speeds, turbines, grid, grid)
     yaw_angle: NDArrayFloat,  # (wind directions, wind speeds, turbines)
+    tilt_angle: NDArrayFloat,  # (wind directions, wind speeds, turbines)
```
rafmudaf, line 360: Need to add these arguments to the docstring

floris/simulation/turbine.py
```diff
@@ -359,17 +561,22 @@ class Turbine(BaseClass):
...
+    pP: float = field()
+    pT: float = field()
+    TSR: float = field()
+    generator_efficiency: float = field()
+    ref_density_cp_ct: float = field()
+    ref_tilt_cp_ct: float = field()
     power_thrust_table: PowerThrustTable = field(converter=PowerThrustTable.from_dict)
-
-
+    # floating_tilt_table: NDArrayFloat = field(
```
rafmudaf, line 574: It looks like this is initialized below so this commented code is probably out of date

floris/simulation/farm.py
rafmudaf: I think you may have missed that change in your commit, I still see it with the underscore in the pull request

bayc: Had it locally, didn't push it up. Should be there now.

floris/tools/floris_interface.py
```diff
@@ -642,6 +662,23 @@ def get_turbine_ais(self) -> NDArrayFloat:
     def turbine_average_velocities(self) -> NDArrayFloat:
         return average_velocity(velocities=self.floris.flow_field.u)
 
+    @property
+    def turbine_effective_velocities(self) -> NDArrayFloat:
```
bayc, line 666: @rafmudaf Added `turbine_effective_velocities` at the `floris_interface` level. Should this be an `@property`? Modelled after `turbine_average_velocities`.

paulf81:
The latest commit [adding turbine_effective_velocities to floris_interface](https://github.com/NREL/floris/pull/518/commits/138f56a74367da0325aed6ed5f6470f0668765a3) I think is causing example 24 to break,

bayc:
> The latest commit [adding turbine_effective_velocities to floris_interface](https://github.com/NREL/floris/pull/518/commits/138f56a74367da0325aed6ed5f6470f0668765a3) I think is causing example 24 to break,

Thanks @paulf81, I've updated so that all the examples work now.

paulf81:
Hi @bayc , I was thinking I was going to switch the example to use an external library folder, instead of the !include syntax to reference the floating turbine model, but I think this path would have to be specified so then it seems better actually to leave these as is for now, and maybe we can include syntax on using the external library folder to documentation rather than examples, what do you think @rafmudaf ?

rafmudaf (approved):

PR #521, opened 2022-11-10
Feature: Add external library functionality

RHammond2:
This PR adds support for using external turbine libraries for turbine defintions to `floris.tools.FlorisInterface` and `floris.simulation.Farm`.

This would bump up the priority of #345 and likely close it.

`floris.simulation.Farm` now has a `turbine_library` attribute that defaults to the currently assumed "floris/turbine_library" folder that ensures the input file path exists prior to loading in the data. Additionally this input was added to the `floris.tools.FlorisInterface.reinitialize()` method to ensure usage is consistent with the rest of the inputs. I've also added a few other tests and some test data to ensure this is working as expected.

I saw in #345 that this was marked as v3.3/v3.4 issue, however, in a separate project I'm running into an issue where the current workaround is a little bit clunky in practice, and am proposing the implemented solution here as a starting point.

All previously existing and newly added tests pass.


RHammond2:
The tests are passing locally, but I'm getting the following error when running in the CI Pipeleine. @rafmudaf do you know how to correctly configure the `tests/data` location to work both locally and through the CI pipeline? If not, I can look into it more too, but it seemed like one of those areas where you might already know the answer.
```
FAILED tests/farm_unit_test.py::test_farm_external_library - FileNotFoundError: User-selected turbine definition `nrel_5mw` does not exist in the turbine library: /home/runner/work/floris/floris/tests/data.
```

rafmudaf:
Hey @RHammond2 I see you have "nrel_5mw" in the test case but the file is "nrel_5MW.yaml". Ubuntu is cases sensitive but macOS is not, so that may be the reason for the inconsistency on your computer vs GitHub Actions.

bayc:
@RHammond2 , the change itself looks good to me! Thanks for getting this implemented. I think this is better than having the user specify the full absolute or relative path to the turbine files with the `!include` decorator. I see you have added the `turbine_library` variable as input to `floris_interface`, but I am thinking it would be nice to also have it as an option in the input files themselves. I can put together a commit with this as an optional input if you and @rafmudaf think that would be good.

RHammond2:
@rafmudaf thanks for the pointer, I didn't even catch that I mistyped the "MW" in the `turbine_type` input! I've just committed an update for this, which should take care of the error.

@bayc I'm very ok with that addition, so go for it!

rafmudaf:
Good idea @bayc.

Additionally, I think this implementation requires to use either the user-supplied library or the default library but not both. Is it possible to include support for using both? This would allow using the NREL 5MW as well as a custom turbine definition without having to redefine the NREL 5MW, for example.

rafmudaf:
Also @RHammond2 the codecov upload step has been failing for me, but I have not taken the time to understand why. It seems this happens from time to time, and I'm leaning toward just turning off codecov since I personally never look at it and probably no one else does either.

paulf81:
> 

This same issue has been plaguing FLASC, some googling and it seems to be something between github.com and codecov so also want to remove it from FLASC, so if we take the decision here would like to know how it's done

RHammond2:
@rafmudaf thanks for clarifying the confusion on the code coverage report. If it's only failing sometimes, I think I'm ok with merging if it's only that last step failing and ignoring it, but if it's always failing, then I'd say let's take it out and we can figure this out another way possibly.

On further inspection, in Openoa we do a similar CI step: https://github.com/NREL/OpenOA/blob/main/.github/workflows/ci-tests.yml#L36-L37 with the major difference being that we don't use this line, which is present in FLORIS: https://github.com/NREL/floris/blob/main/.github/workflows/continuous-integration-workflow.yaml#L51. So it seems like we could just remove that line and the CI integration won't fail if it can't do the bonus step of uploading the coverage report.

RHammond2:
@rafmudaf and @bayc this should be good to go now.

rafmudaf:
See #568

PR #535, opened 2022-11-16
Configure isort and apply rules

rafmudaf:
This pull request configures [isort](https://pycqa.github.io/isort/index.html) and applies the rules to all included modules.

Here's my opinions and reasoning for certain configurations.

- `force_grid_wrap = 3`: Forces that any from-imports with more than three symbols are wrapped. This one is important to me because I like the ability to easily edit or comment out items in a list without destroying the syntax of the list. Three is somewhat arbitrary since the default is two, but intuitively it feels right. With two imports, its pretty straightforward to add a comment mid-line without completely destroying the line. It's also worth pointing out that this preference has more to do with Python syntax style like using containers in code rather than import-styles. However, the import statement syntax should match the code syntax.

```python
# For example...
from turbine import Ct, axial_induction, power

# To remove axial_induction requires to also remove power or restructure the list
from turbine import Ct #, axial_induction, power
from turbine import Ct, power #, axial_induction

# Alternatively, wrapping allows to easily remove axial_induction
from turbine import (
    Ct,
#    axial_induction,
    power
)

# 3 vs 2
# This seems like overkill but I would be happy to change to 2 for consistency
from turbine import (
    Ct,
    power
)
```

- `multi_line_output = 3`: This is consistent with the common form for indented containers in Python code.
- `combine_as_imports = true`: I find it helpful to quickly see all of the renamed packages since this provides additional context for the rest of the file. For example, `import floris.tools.visualization as wakeviz` explains quickly what is `wakeviz` in the code, so it is helpful to see all of these up front and easily parsable.

- `include_trailing_comma = false`: Personal preference, I'm not a fan of the trailing comma but this is highly subjective.
- `line_length = 88`: My preference is longer lines like 100, but 88 seems to be the de facto standard.
- `order_by_type = false`: This order imports by the perceived type based on letter case in a symbol's name (i.e. `Turbine` is a class but `turbine` is a module). Since FLORIS does not strictly adhere to this, I've disabled it.

Most files but only the import statements.

This does not yet enforce the ordering via git or GitHub. It is up to the developers to run isort. However, an automated system will be put in place in the future.

### DIFF
--- a/docs/dev_guide.md
+++ b/docs/dev_guide.md
@@ -58,6 +58,31 @@ to fit a specific use case or purpose. If significant development has
 happened locally, then [merge conflicts](https://www.atlassian.com/git/tutorials/using-branches/merge-conflicts)
 are likely and should be resolved as early as possible.
 
+
+## Code quality tools
+
+FLORIS is configured to use tools to automatically check and enforce
+aspects of code quality. In general, these should be adopted by all
+developers and incorporated into the development workflow. Most
+tools are configured in [pyproject.toml](https://github.com/NREL/floris/blob/main/pyproject.toml),
+but some may have a dedicated configuration file.
+
+### isort
+
+Import lines can easily get out of hand and cause unnecessary distraction
+in source code files. [isort](https://pycqa.github.io/isort/index.html)
+it used to automatically manage imports in the source code. It can be run
+directly with the following command:
+
+```bash
+isort <path to file>
+isort dir/*
+```
+
+This tool was initially configured in [#535](https://github.com/NREL/floris/pull/535),
+and additional information on specific decisions can be found there.
+
+
 ## Testing
 
 In order to maintain a level of confidence in the software, FLORIS is expected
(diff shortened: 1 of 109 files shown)


RHammond2 (approved):
Thanks for revisiting the autoformatting! I'm not much a stickler for any of the changes you've made, and therefore have no pushback so long as there is a consistent style. I have a couple of clarifying comments below.

- `force_grid_wrap=3`: I agree with this one quite strongly
- `multi_line_output=3`: This is also my preferred styling to keep things easy to read and consistent with things I like about PEP8
- `combine_as_imports=true`: This is also my preferred styling to keep things as tidy as possible
- `include_trailing_comma=false`: Not my preferred method, but I really don't have strong enough opinions to make a counterargument. This more or less falls in the category of things I like about PEP8, but isn't too high on the priority list.
- `line_length=88`: We could easily do 100, and this setting can modified in all the linters/formatters that we use or are considering, so feel free to bump it to 100. We use 122 for OpenOA, so I'm happy to take your lead on that change (and I've been leaning towards 100 myself recently).
- sorting: I like the cascade of length sort, but it's a highly personal aesthetic choice. That said, the alphabetic sorting makes a lot of sense and am perfectly fine with this change too.

rafmudaf:
Thanks @RHammond2

PR #555, opened 2023-01-09
Add parallel computing interface for farm simulation and yaw optimization

Bartdoekemeijer:
This PR adds a parallel computing interface at the tools level to enable parallel computing of turbine powers, wind farm powers, and yaw optimizations. It does not change anything to the internal FLORIS code, which makes it a much easier integration than #403. I have been using this internally for quick parallel computing solutions using `multiprocessing` for single-node jobs/local parallelization, but this class also supports `mpi4py` for multinode calculations like FLORIS v2 used to.

#401 

FLORIS tools library

The Parallel Computing Interface class is built similarly to the UncertaintyInterface class. It can be loaded up and then interacted with like a regular `FlorisInterface` object, but then the underlying code performs the computations in parallel rather than in serial. In larger problems and especially in yaw optimizations, this can provide significant speed-up.

This PR adds an example file that calculates the optimal yaw angles and the resulting AEP uplift using parallel processing, "12_optimize_yaw_in_parallel.py".

With a maximum of one core, the total time for the yaw angle optimization (using SR) and for the AEP calculation are, respectively:
```
==============================================
[Yaw optimization with SR]
Total time spent for parallel calculation (1 workers): 280.834 s
  Time spent in parallel preprocessing: 0.001 s
  Time spent in parallel loop execution: 280.831 s.
  Time spent in parallel postprocessing: 0.002 s
==============================================
[AEP calculation]
Total time spent for parallel calculation (1 workers): 35.606 s
  Time spent in parallel preprocessing: 0.007 s
  Time spent in parallel loop execution: 35.596 s.
  Time spent in parallel postprocessing: 0.004 s
```

And in parallel with 16 cores on my laptop:
```
==============================================
[Yaw optimization with SR]
Total time spent for parallel calculation (16 workers): 87.727 s
  Time spent in parallel preprocessing: 0.003 s
  Time spent in parallel loop execution: 87.720 s.
  Time spent in parallel postprocessing: 0.004 s
==============================================
[AEP calculation]
Total time spent for parallel calculation (16 workers): 11.781 s
  Time spent in parallel preprocessing: 0.012 s
  Time spent in parallel loop execution: 11.763 s.
  Time spent in parallel postprocessing: 0.006 s
```

And the example produces the results:
image

image

image

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -108,11 +108,6 @@ Construct wind farm yaw settings for a full wind rose based on the
 optimized yaw settings at a single wind speed. Then, compare
 results to the baseline no-yaw configuration.
 
-### 12_optimize_yaw.py
-Construct wind farm yaw settings for a full wind rose based on the
-optimized yaw settings at a single wind speed. Then, compare
-results to the baseline no-yaw configuration.
-
 ### 12_optimize_yaw_in_parallel.py
 Comparable to the above but perform all the computations using
 parallel processing. In the current example, use 16 cores
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -108,6 +108,17 @@ Construct wind farm yaw settings for a full wind rose based on the
 optimized yaw settings at a single wind speed. Then, compare
 results to the baseline no-yaw configuration.
 
+### 12_optimize_yaw.py
+Construct wind farm yaw settings for a full wind rose based on the
+optimized yaw settings at a single wind speed. Then, compare
+results to the baseline no-yaw configuration.
+
+### 12_optimize_yaw_in_parallel.py
+Comparable to the above but perform all the computations using
+parallel processing. In the current example, use 16 cores
+simultaneously to calculate the AEP and perform a wake steering
+yaw angle optimization for multiple wind speeds.
+
 ### 13_optimize_yaw_with_neighboring_farm.py
 Same as above but considering the effects of a nearby wind farm.
 
--- a/examples/12_optimize_yaw_in_parallel.py
+++ b/examples/12_optimize_yaw_in_parallel.py
@@ -50,6 +50,9 @@ def load_windrose():
 
 
 if __name__ == "__main__":
+    # Parallel options
+    max_workers = 16
+
     # Load a dataframe containing the wind rose information
     df_windrose, windrose_interpolant = load_windrose()
 
@@ -66,8 +69,8 @@ if __name__ == "__main__":
     # Pour this into a parallel computing interface
     fi_aep_parallel = ParallelComputingInterface(
         fi=fi_aep,
-        max_workers=16,
-        n_wind_direction_splits=16,
+        max_workers=max_workers,
+        n_wind_direction_splits=max_workers,
         n_wind_speed_splits=1,
         use_mpi4py=False,
         print_timings=True,
@@ -97,7 +100,7 @@ if __name__ == "__main__":
     # Load a FLORIS object for yaw optimization
     fi_opt = load_floris()
     wind_directions = np.arange(0.0, 360.0, 3.0)
-    wind_speeds=np.arange(6.0, 14.0, 2.0)
+    wind_speeds = np.arange(6.0, 14.0, 2.0)
     fi_opt.reinitialize(
         wind_directions=wind_directions,
         wind_speeds=wind_speeds,
@@ -107,8 +110,8 @@ if __name__ == "__main__":
     # Pour this into a parallel computing interface
     fi_opt_parallel = ParallelComputingInterface(
         fi=fi_opt,
-        max_workers=16,
-        n_wind_direction_splits=16,
+        max_workers=max_workers,
+        n_wind_direction_splits=max_workers,
         n_wind_speed_splits=1,
         use_mpi4py=False,
         print_timings=True,
--- a/examples/12_optimize_yaw_in_parallel.py
+++ b/examples/12_optimize_yaw_in_parallel.py
@@ -1,0 +1,285 @@
+# Copyright 2022 NREL
+
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+
+# See https://floris.readthedocs.io for documentation
+
+
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+from scipy.interpolate import LinearNDInterpolator
+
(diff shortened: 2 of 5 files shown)


mr483:
Hi @Bartdoekemeijer, I was testing out this interface and it works nicely! One question: have you done any sensitivity study on performance improvement with number of cores used and/or number of condition splits? I'm finding on my machine (16 cores) that the performance of running fi_parallel.get_turbine_powers() for 400 turbines does not necessarily show the scaling up of improvement as I would expect:

| Cores (max_workers) | Runtime (s) | Improvement |
| :---         |     :---:      |          ---: |
| 1   | 890  | -  |
| 4   |  270 | 3.3     |
| 8   | 245       | 3.63      |
| 12   | 230       | 3.87      |
| 16   | 238       |  3.74     |

In each case, I confirmed that selecting to split by either just WD (e.g., 8 workers, 8 wd splits, 1 ws split), just WS, or some combination of WS/WD which multiplies to the number of workers (e.g., 4 wd splits, 2 ws splits) didn't make a meaningful difference, as expected.

Bartdoekemeijer:
> Hi @Bartdoekemeijer, I was testing out this interface and it works nicely! One question: have you done any sensitivity study on performance improvement with number of cores used and/or number of condition splits? I'm finding on my machine (16 cores) that the performance of running fi_parallel.get_turbine_powers() for 400 turbines, 30 wind speeds, 60 wind directions, does not necessarily show the scaling up of improvement as I would expect: 
> ...

Hi @mr483, thank you for testing out this PR! I have found similar things as you did: performance does not always increase with more cores. I think this is because for large wind farms and many wind directions/speeds, the limiting factor in calculations is the available memory rather than CPU power. Hence, more cores then does not speed up the code, but does add additional overhead, so the net result is a slow-down of the code. This is especially apparent with operations like `get_farm_power`, `get_turbine_power` and `get_farm_AEP`.
If you would use this parallel computing interface for yaw optimization, then the benefit of parallel computing becomes a lot clearer. I think the yaw optimization is much more CPU bottlenecked.

mr483:
> > Hi @Bartdoekemeijer, I was testing out this interface and it works nicely! One question: have you done any sensitivity study on performance improvement with number of cores used and/or number of condition splits? I'm finding on my machine (16 cores) that the performance of running fi_parallel.get_turbine_powers() for 400 turbines, 30 wind speeds, 60 wind directions, does not necessarily show the scaling up of improvement as I would expect:
> > ...
> 
> Hi @mr483, thank you for testing out this PR! I have found similar things as you did: performance does not always increase with more cores. I think this is because for large wind farms and many wind directions/speeds, the limiting factor in calculations is the available memory rather than CPU power. Hence, more cores then does not speed up the code, but does add additional overhead, so the net result is a slow-down of the code. This is especially apparent with operations like `get_farm_power`, `get_turbine_power` and `get_farm_AEP`. If you would use this parallel computing interface for yaw optimization, then the benefit of parallel computing becomes a lot clearer. I think the yaw optimization is much more CPU bottlenecked.

Thanks for your reply. I agree it's related to memory, but perhaps memory bandwidth rather than overall memory availability. I tested on a machine where these calculations used a relatively small percent of available memory, but I've read that the memory bandwidth will still limit performance gains when doing a lot of read/write to memory.

Bartdoekemeijer:
Here are test results for this PR on the Shell high-performance computing facility and on my personal system. These are the timings for a 64-turbine yaw optimization, with `wind_directions = np.arange(0.0, 360.0, 3.0)` and `wind_speeds = np.arange(6.0, 14.0, 2.0)`. The Python script is attached below.

image

This shows that both the `multiprocessing` and `mpi4py` interfaces in `ParallelComputingInterface` work as expected. We also see better performance with more nodes even if the total number of workers is unchanged. I think this confirms our notion that FLORIS can benefit from more memory (i.e., more nodes means more memory available), rather than just more processing power.

Further context: the SLURM jobs are submitted using `mpiexec -n 128 python -m mpi4py.futures yaw_optimization.py`. And here's the Python code I used for the benchmarking:
```
import numpy as np
import pandas as pd

from floris.tools import ParallelComputingInterface


def load_floris():
    fi = FlorisInterface("inputs/gch.yaml") # GCH model matched to the default "legacy_gauss" of V2
    # fi = FlorisInterface("inputs/cc.yaml") # New CumulativeCurl model

    # Specify wind farm layout and update in the floris object
    N = 8  # number of turbines per row and per column
    X, Y = np.meshgrid(
        5.0 * fi.floris.farm.rotor_diameters_sorted[0][0][0] * np.arange(0, N, 1),
        5.0 * fi.floris.farm.rotor_diameters_sorted[0][0][0] * np.arange(0, N, 1),
    )
    fi.reinitialize(layout_x=X.flatten(), layout_y=Y.flatten())

    return fi


if __name__ == "__main__":
    # Parallel options
    max_workers = [128, 64, 32, 8, 1]
    use_mpi4py = True

    # Load a FLORIS object for yaw optimization
    print("Loading the FLORIS model...", flush=True)
    fi_opt = load_floris()
    wind_directions = np.arange(0.0, 360.0, 3.0)
    wind_speeds = np.arange(6.0, 14.0, 2.0)
    fi_opt.reinitialize(
        wind_speeds=wind_speeds,
        turbulence_intensity=0.08  # Assume 8% turbulence intensity
    )


    timings = np.zeros(len(max_workers))
    for mii, mw in enumerate(max_workers):
        print("Running optimizations now with {:d} workers.".format(mw), flush=True)
        t0 = timerpc()  # Start time
        # Pour this into a parallel computing interface
        fi_opt_parallel = ParallelComputingInterface(
            fi=fi_opt,
            max_workers=mw,
            n_wind_direction_splits=mw,
            n_wind_speed_splits=1,
            use_mpi4py=use_mpi4py,
            print_timings=True,
        )

        # Now optimize the yaw angles using the Serial Refine method
        df_opt = fi_opt_parallel.optimize_yaw_angles(
            minimum_yaw_angle=-25.0,
            maximum_yaw_angle=25.0,
            Ny_passes=[5, 4],
            exclude_downstream_turbines=True,
            exploit_layout_symmetry=False,
        )
        t1 = timerpc()  # End time
        timings[mii] = (t1 - t0)  # Save total time

    print(" ", flush=True)
    print("Final timings:")
    for mii in range(len(max_workers)):
        print("  {:d} workers: {:.2f} seconds.".format(max_workers[mii], timings[mii]))
```

Could you review @rafmudaf, @bayc and let me know if any changes should be made before pulling this into `develop`?

paulf81:
hi @Bartdoekemeijer  thanks a lot of this PR!  I was wondering, we're planning to do some focused development of FLORIS next week, do you think this PR is ready to merge?

Bartdoekemeijer:
Yes ready to merge if you ask me! I have been very actively using it on different systems and with and without MPI. It could use some documentation but that can be merged in later.

paulf81:
Excellent, thank you!

rafmudaf:
Please include example in the listing here: https://github.com/NREL/floris/blob/develop/docs/examples.md

Bartdoekemeijer:
> Please include example in the listing here: https://github.com/NREL/floris/blob/develop/docs/examples.md

Done!

rafmudaf:
Thanks @Bartdoekemeijer!

PR #562, opened 2023-02-03
Add Ruff linting and automormatting

rafmudaf:
This pull request adds configuration for [Ruff](https://github.com/charliermarsh/ruff) as a linker and in some cases autoformatter. The decision to use Ruff came about after evaluating a few others tools. The decision and key priorities are described in a discussion, https://github.com/NREL/floris/discussions/561.

The basic configuration is in place here and most errors have been fixed. A few errors are ignored and should be addressed in the future. These are listed in the pyproject.toml ruff configuration under the header "# FIXME"; see https://github.com/rafmudaf/floris/blob/ruff/pyproject.toml#L73.

This pull request causes quite a few changes throughout the project, but they are all stylistic and no algorithmic changes are included.

No issue, but see this discussion: https://github.com/NREL/floris/discussions/561.

Most files are changed to conform to the Ruff rules.


rafmudaf:
@RHammond2 Could you please look over this pull request and especially the `pyproject.toml` configuration? It's basic to begin with, but please let me know if anything stands out.

RHammond2:
@rafmudaf I'm adding some comments for settings we should add or think about adding in the future. Otherwise, I think this is a good setup to get us working towards replacing our current setup, especially while it's still in active development for expanding and refining its capabilities.

- D: pydocstyle
  - This will be a useful catch for technical errors in our docstrings, given it has a google style setting
  - However, I think most of our docstring issues stem from the fact that they don't provide a whole lot of useful information, and this won't remotely help with the real issues we have.
- UP: pyupgrade
  - I've been meaning to use this in other libraries to ensure old code snippets aren't out of date and to find places where some core usage has changed, particularly for targeting specific version of Python we want to support.
  - I think we'll have already addressed much of this in our initial v3 push, but could helpful to ensure we're using modern Python for the lifetime of the project
- FBT: flake8-boolean-trap
  - This could be nice as a safety check, but I don't think we have much in the way of boolean inputs, so could not provide too much value
- B: flake8-bugbear
  - I'd want to look into this more, so I'd definitively put this in the later category, but could be helpful as an additional check on our code style for non PEP8 oddities.
- PT: flake8-pytest-style
  - I could go either way with this one, but I'm wondering if this could help people getting acquainted with writing tests
- RET: flake8-return
  - I haven't used this one, but I think given the issue  we had with porting the old code, we might find this to be helpful in avoiding non-pythonic or odd return patterns.
- SIM: flake8-simplify
  - Seems neat, but could get us down a rabbit hole, so this might be something to consider later to help catch odd code patterns
- PTH: flake8-use-pathlib
  - I'd like to see this one added to help people switch from `os` to `pathlib.Path` and help us identify where to switch things out
- ERA: eradicate
  - We should never use this unless we're happy with removing our old code tracks

rafmudaf:
- [ ] Include in pre-commit workflow
- [ ] Update `setup.py` to include ruff dependency for developer option

RHammond2 (approved):

PR #567, opened 2023-02-06
Enable and fix pre-commit formatting hooks

rafmudaf:
This pull request enables pre-commit hooks for auto formatting and fixes the errors found through the repository. It is entirely whitespace and extra line fixes. This is automatically run in GitHub Actions and will provide feedback to developers via the typical emails and Actions logs.

[Pre-commit](https://pre-commit.com) is a command line utility to run a series of code style checks. It is configured in `.pre-commit-config.yaml`. The configured hooks must be installed into each development environment with the following command:

```bash
cd floris/
pre-commit install
```

Once installed, they are executed during each commit creation directly through `git` for each file that contains changes. Therefore, files that are not changed in a commit will not be modified my pre-commit.


RHammond2 (approved):
I think this looks good, and it's smart to take black out entirely

PR #568, opened 2023-02-06
Feature/turbine library

RHammond2:
This PR is a new copy of #521, with the latest changes on develop, and adds support for using external turbine libraries for turbine defintions to `floris.tools.FlorisInterface` and `floris.simulation.Farm`.

This would bump up the priority of #345 and likely close it.
Closes #521 because it's outdated and got bungled in a merge resolution.

`floris.simulation.Farm` now has a `turbine_library` attribute that defaults to the currently assumed "floris/turbine_library" folder that ensures the input file path exists prior to loading in the data. Additionally this input was added to the `floris.tools.FlorisInterface.reinitialize()` method to ensure usage is consistent with the rest of the inputs. I've also added a few other tests and some test data to ensure this is working as expected.

I saw in #345 that this was marked as v3.3/v3.4 issue, however, in a separate project I'm running into an issue where the current workaround is a little bit clunky in practice, and am proposing the implemented solution here as a starting point.

All previously existing and newly added tests pass.


floris/type_dec.py
```diff
@@ -82,6 +85,28 @@ def iter_validator(iter_type, item_types: Union[Any, Tuple[Any]]) -> Callable:
...
     return validator
 
+
+def convert_to_path(fn: str | Path) -> Path:
+    """Converts an input string or pathlib.Path object to a fully resolved pathlib.Path
+    object.
+    Args:
+        fn (str | Path): The user input file path or file name.
+    Raises:
+        AttributeError: _description_
```
rafmudaf, line 95: Update here to TypeError with description and delete the extra one below

floris/simulation/farm.py
```diff
@@ -13,6 +13,7 @@
 from __future__ import annotations
 
 import copy
+import json
```
rafmudaf, line 16: It may be safer to use `import yaml` here. This works, but since json is a subset of yaml there may be yaml features that aren't handled.

tests/type_dec_unit_test.py
```diff
@@ -114,3 +116,36 @@ def test_attrs_array_converter():
     # Test converstion on reset
     cls.array = array_input
     np.testing.assert_allclose(test_array, cls.array)
+
+
+def test_convert_to_path():
```
rafmudaf, line 121: Well done!

floris/simulation/farm.py
```diff
@@ -50,6 +55,9 @@ class Farm(BaseClass):
     layout_x: NDArrayFloat = field(converter=floris_array_converter)
     layout_y: NDArrayFloat = field(converter=floris_array_converter)
     turbine_type: List = field()
+    turbine_library: Path = field(
```
rafmudaf, line 58: It might be more accurate to call this attribute `turbine_library_path`. In reading the unit test you added, it seems odd to compare `farm.turbine_library` to a string.

rafmudaf:
The last comment I have is a request to add a simple example usage in the pull request description above. This will ultimately be included in the documentation, but it's good enough to include here initially.

Is this a feature that can only be used through a script or is there a way to set an external turbine library through the input file? If so, please provide the relevant blocks of the input file that would be used.

floris/simulation/farm.py
RHammond2: That makes sense, I'll test this out to try and get a clean swap.

floris/simulation/farm.py
RHammond2: Happy to change, but I'm not sure which part of the test you're referring to. It seems like the external_library comparison and the default_library are both `Path` objects, or they'd fail.

RHammond2:
Good call on the example part, I'll add something that calls more attention to it, so users know it's available. As an aside, it was added to the documentation PR in the input reference sheet, so I'll leave that component alone.

floris/simulation/farm.py
rafmudaf: Oh woops - I misread the unit test.
What I meant to get at here is that `turbine_library` sounds like it is the library itself. Without knowing anything about it, I might try to pull a turbine out of it. However, it would be more explicit to indicate that its actually the filepath to the turbine definitions.

rafmudaf (approved):

paulf81 (approved):
Thank you @RHammond2 , I think this is a really nice and general solution

bayc (approved):
Looks great!

rafmudaf:
@scottryn thanks for the info. Could you please open an issue and include a minimal script to reproduce the error?

scottryn:
Hi @rafmudaf, thanks for getting back to me so quickly! I installed in a clean Conda environment and everything works as intended. I'll hold off on raising an issue since it appears I have ghosts in my other environment.

Issue #563, opened 2023-02-06
Actions completion time degraded after #555


rafmudaf:
### Bug description
This isn't exactly a bug, but it is a degradation in performance. Since #555, the [Check Examples API](https://github.com/NREL/floris/actions/workflows/check-working-examples.yaml) workflow in GitHub Actions takes about 11 minutes to complete whereas it finished in around 5 minutes before. This is expected since the new example included in #555 takes a few minutes to run the calculation.

One possible improvement to that workflow is to run the examples in parallel. Another option is to reduce the computation included in the example from #555.

rafmudaf:
@Bartdoekemeijer what do you think about reducing the computation cost in the example through something like increasing the step size for the wind rose?

Bartdoekemeijer:
Good catch! I had no idea this would pose an issue. How about:

1. Set `N=4` or `N=3` in line 35 to go from 25 to 16 or even 9 wind turbines in the `FlorisInterface` object.
2. Set `wind_directions = np.arange(0.0, 360.0, 3.0)` to go from 360 to 120 wind directions.
3. Set `wind_speeds=np.arange(1.0, 25.01, 1.0)` to go from 30 to 25 wind speeds.

On my local machine with one worker, the entire example with `N=4` takes about 3 minutes to complete. With 16 workers, the entire example takes about 30 seconds to complete. With `N=3`, these numbers are 30 seconds and 12 seconds, respectively.

Generally, I'm hesitant to go to a coarser resolution in the wind rose than 3 deg steps. I feel like many users just take an example, modify it to their wind farm, and leave the default settings assuming they are representative of how actual yield calculations should be done. Hence, I'd like to keep a realistic wind rose resolution.

rafmudaf:
@Bartdoekemeijer I change N to 4 for 16 turbines and reduced the max wind speed to 25 in #569. Now the workflow completes in 7-8 minutes which is pretty close to the 5-6 minutes that it was before. Thanks for your input here.

PR #569, opened 2023-02-07
Infrastructure improvements

rafmudaf:
This pull request is a collection of minor infrastructure improvements.

1. In order to preserve isort formatting over ruff or any other tools, the pre-commit configuration swaps the order in which the hooks run and leaves isort last.
2. Due to GitHub rate limiting, the codecov upload step in the "Automated tests & code coverage" workflow can fail unexpectedly. One solution is to simply restart the workflow, but this takes time and manual intervention. Codecov recommend to include an upload token in the uploader. See [here](https://community.codecov.com/t/upload-issues-unable-to-locate-build-via-github-actions-api/3954) for their description of the problem and suggested solutions.
3. PR #555 added a new example that demonstrates a performance improvement in AEP calculations by splitting the atmospheric conditions over multiple threads. The example takes about 5 minutes to run in GitHub Actions, so it noticeably slowed the completion time. Issue #563 discusses some solutions. This commit reduces the number of turbines and wind speeds.
4. An automated workflow to deploy a package to PyPI could trigger for any fork on FLORIS when a release is created. This commit restricts that workflow to only the FLORIS repository under the NREL account.
5. Since #562, the "Automated tests & code coverage" workflow stops and reports failure if the linting step fails. This means that the tests are not run until the code passes the linting check. This commit allows the tests to run regardless of the linting results. The overall workflow is still reported as a failure, but developers get feedback on validity of the results in addition to linting status.

#563

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yaml
+++ b/.github/workflows/continuous-integration-workflow.yaml
@@ -24,8 +24,8 @@ jobs:
         pip install -e ".[develop]"
     - uses: pre-commit/action@v3.0.0
     - name: Run tests
+      if: success() || failure()  # Run this step even if the linting step fails
       run: |
-        # pytest tests/
         # -rA displays the captured output for all tests after they're run
         # See the docs: https://doc.pytest.org/en/latest/reference/reference.html#command-line-flags
         pytest -rA tests/ --ignore tests/timing.py --ignore tests/profiling.py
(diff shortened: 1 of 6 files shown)


.github/workflows/continuous-integration-workflow.yaml
```diff
@@ -24,8 +24,8 @@ jobs:
         pip install -e ".[develop]"
     - uses: pre-commit/action@v3.0.0
     - name: Run tests
+      if: success() || failure()  # Run this step even if the linting step fails
```
RHammond2, line 27: This makes a lot of sense, I wish we thought of that initially!

RHammond2 (approved):
This seems pretty straightforward and agreeable to me.

PR #556, opened 2023-01-09
Feature: Updating "rotate_coordinates_rel_west" function in utilities to fix farm orientation for plots

MYMahfouz:
**Feature or improvement description**
Fixing the farm orientation before plotting, this pull request builds on the work done in an earlier pull request: https://github.com/NREL/floris/pull/533

**Related issue, if one exists**
This pull request builds on the work done in an earlier pull request: https://github.com/NREL/floris/pull/533
Fix orientation: https://github.com/NREL/floris/discussions/348, https://github.com/NREL/floris/issues/357, https://github.com/NREL/floris/issues/372
non-matching plot turbine with fi: https://github.com/NREL/floris/issues/448, https://github.com/NREL/floris/issues/451, https://github.com/NREL/floris/discussions/455, https://github.com/NREL/floris/pull/512

**Impacted areas of the software**
utilities.py: fn rotate_coordinates_rel_west() adding 2 parameters (farm layout_x and layout_y)
florice_interface.py: updated calling rotate_coordinates_rel_west() within get_plane_of_points()
There other changes that I got from https://github.com/NREL/floris/pull/533
fix orientation
v grid.py: class flowfiledplanar Grid method set_grid (Braoden flowfield, rotate mesh grid)
v utilities.py: ftn rotate_coordinates_rel_west() adding 1 parameter (inv_rot)
v floris_interface.py: method get_plane_of_points() (reverse rotating to north up)

**Additional supporting information**

**Test results, if applicable**

PR #578, opened 2023-02-15
Add coordinates from inertial frame to grid class

Bartdoekemeijer:
The pull request **is** ready to be merged, but requires review.

This pull request adds the inertial locations of the flow fields/turbine locations in the `grid` class. With inertial, I mean the location of the turbines without rotating it to the incoming wind direction. These additional variables should also be helpful when working with heterogeneous inflow conditions. The way it seems to be implemented now, the heterogeneous inflow conditions are linearly interpolated using the rotated (wind-aligned) coordinates, meaning their physical location differs for different wind directions. I'm opening a separate issue for that. Anyway, this PR is also particularly helpful when we want to plot horizontal flow slices when fixing the plot to true north and with the turbines plotted in same location for every wind direction -- i.e., the same issue as described in #556.

I have added a couple of handles to the visualization such that it can be used for plotting horizontal flow slices while fixing the orientation to true north. Note that nothing fundamentally has changed to the FLORIS simulation code, besides the exporting/calculation of a couple extra variables, so it's an easy merge.

@MYMahfouz provides a nice overview in #556.

The FLORIS `grid` and `visualization` classes.

These additional variables should also be helpful when working with heterogeneous inflow conditions. The way it seems to be implemented now, the heterogeneous inflow conditions are linearly interpolated using the rotated (wind-aligned) coordinates, meaning their physical location differs for different wind directions. I'm opening a separate issue for that.

A simple test case is:
```
import numpy as np

from floris.tools.visualization import plot_turbines_with_fi

fi.reinitialize(wind_directions=[51.0])

yaw_angles = np.array([[[25.,0.,0.]]])
rotate_to_inertial_frame = True
horizontal_plane = fi.calculate_horizontal_plane(
    x_resolution=200,
    y_resolution=100,
    height=90.0,
    rotate_to_inertial_frame=rotate_to_inertial_frame,
    yaw_angles=yaw_angles,
)

fig, ax = plt.subplots()
wakeviz.visualize_cut_plane(horizontal_plane, ax=ax, title="Horizontal")
plot_turbines_with_fi(
    fi,
    ax=plt.gca(),
    color="black",
    rotate_to_inertial_frame=rotate_to_inertial_frame,
    yaw_angles=yaw_angles,
)

plt.show()
```

which produces
image


Bartdoekemeijer:
Also see my comments in #579.

floris/simulation/grid.py
```diff
@@ -69,6 +72,8 @@ class Grid(ABC):
     wind_speeds: NDArrayFloat = field(converter=floris_array_converter)
     time_series: bool = field()
 
+    _xc_rot: float = field(init=False)
```
rafmudaf, line 75: Are these safe to delete? I only see them written to but not read from.

floris/simulation/grid.py
```diff
@@ -265,6 +276,33 @@ def set_grid(self) -> None:
         self.y = np.take_along_axis(self.y_sorted, self.unsorted_indices, axis=2)
         self.z = np.take_along_axis(self.z_sorted, self.unsorted_indices, axis=2)
 
+        # Now calculate grid coordinates in original frame (from 270 deg perspective)
+        self.x_inertial_frame, self.y_inertial_frame, self.z_inertial_frame = \
```
rafmudaf, line 280: The `TurbineGrid` contains points only on the rotor planes, so this is not used for the horizontal slice visualizations. I think these are not required here since they would not be used.

floris/simulation/grid.py
Bartdoekemeijer: Good catch, I think you are right. These may be fine to delete. It's generally worth thinking through how the heterogeneous map is interpolated and applied in the wind-aligned frame that FLORIS does its calculations in. Though, I think the example in #579 shows that the variable `_xc_rot` is probably not needed.

floris/simulation/grid.py
Bartdoekemeijer: I think you are right. See also my comment above. I think the main variables that we need to properly apply heterogeneous inflow conditions are `grid.x_sorted_inertial_frame`, and `grid.y_sorted_inertial_frame`.

floris/simulation/grid.py
rafmudaf: Ok great I've removed them

floris/simulation/grid.py
bayc: I removed the code not needed in this commit (it also contains some small formatting changes to allow me to commit).

https://github.com/NREL/floris/pull/578/commits/624ca2e489a6fb91fa554a0b0592e9aced77064a

bayc:
@Bartdoekemeijer @rafmudaf @paulf81 

Hey all. I've gone through the PR and made a variety of updates. Thanks for all your work on this Bart! I think it is a really nice addition.

The major changes are that the default plotting behavior is now to display the flow field as rotated to the defined wind direction, instead of the 270 degrees used previously.

Also, I addressed the issue identified by Bart in #579 by making sure the `calculate_speed_ups` function is passed the inertial coordinates (captured here https://github.com/NREL/floris/pull/578/commits/e48237bcea22b37427d227054d240f1f9afd4c53). 

Through this process, I re-worked the heterogeneous workflow so that the user defines an `het_config` dictionary and that information is now stored on `FlorisInterface`. I believe this helps to streamline the interface and not require the user to import a separate function to create the interpolant and pass that in, but leverage this optional parameter. I am curious on y'alls feedback regarding this.

I also made changes to how the heterogeneous flows get plotted, such that the user-defined bounds for the het map are displayed by default on any plot that shows heterogenous flows. Additionally, I have removed the nearest-neighbor interpolation that previously occurred outside the user-defined heterogeneous inflow area and used a fill value of 1.0 for the speed ups such that this region is now set to be equal to freestream. This removed a good chunk of code and simplified the whole calculation.

I have put together an example of this, code and plots below, that demonstrates the new code behavior for 4 different wind directions. Note the option to remove the het map bounding box in the bottom right plot. Please check and provide feedback.


```
import numpy as np
from floris.tools.visualization import plot_turbines_with_fi

speed_ups = [[1.2, 1.2, 1.0, 1.0]]
x_locs = [-300.0, -300.0, 2600.0, 2600.0]
y_locs = [ -300.0, 300.0, -300.0, 300.0]

# Generate the configuration dictionary to be used for the heterogeneous inflow.
het_config = {
    'speed_ups': speed_ups,
    'x_locs': x_locs,
    'y_locs': y_locs,
}

# Initialize FLORIS with a heterogeneous inflow
fi_2d = FlorisInterface("inputs/gch.yaml", het_config=het_config)


# Plot the farm for the wind direction of 270 degrees
fig, ax = plt.subplots(2, 2)
yaw_angles = np.array([[[25.,0.,0.]]])
horizontal_plane_270 = fi_2d.calculate_horizontal_plane(
    wd=[270.],
    x_resolution=200,
    y_resolution=100,
    height=90.0,
    yaw_angles=yaw_angles,
)
wakeviz.visualize_heterogeneous_cut_plane(horizontal_plane_270, fi_2d, ax=ax[0][0], title="Horizontal")
plot_turbines_with_fi(
    fi_2d,
    ax=ax[0][0],
    color="black",
    wd=[270.],
    yaw_angles=yaw_angles,
)


# Plot the farm for the wind direction of 275 degrees
horizontal_plane_275 = fi_2d.calculate_horizontal_plane(
    wd=[275.],
    x_resolution=200,
    y_resolution=100,
    height=90.0,
    yaw_angles=yaw_angles,
)
wakeviz.visualize_heterogeneous_cut_plane(horizontal_plane_275, fi_2d, ax=ax[0][1], title="Horizontal")
plot_turbines_with_fi(
    fi_2d,
    ax=ax[0][1],
    color="black",
    wd=[275.],
    yaw_angles=yaw_angles,
)


# Plot the farm for the wind direction of 315 degrees
horizontal_plane_315 = fi_2d.calculate_horizontal_plane(
    wd=[315.],
    x_resolution=200,
    y_resolution=100,
    height=90.0,
    yaw_angles=yaw_angles,
)
wakeviz.visualize_heterogeneous_cut_plane(horizontal_plane_315, fi_2d, ax=ax[1][0], title="Horizontal")
plot_turbines_with_fi(
    fi_2d,
    ax=ax[1][0],
    color="black",
    wd=[315.],
    yaw_angles=yaw_angles,
)


# Plot the farm for the wind direction of 360 degrees
horizontal_plane_360 = fi_2d.calculate_horizontal_plane(
    wd=[360.],
    x_resolution=200,
    y_resolution=100,
    height=90.0,
    yaw_angles=yaw_angles,
)
wakeviz.visualize_heterogeneous_cut_plane(horizontal_plane_360, fi_2d, ax=ax[1][1], title="Horizontal", plot_het_bounds=False)
plot_turbines_with_fi(
    fi_2d,
    ax=ax[1][1],
    color="black",
    wd=[360.],
    yaw_angles=yaw_angles,
)

# Show the plots
plt.show()
```

image

Bartdoekemeijer:
Thanks Chris, this is great work!! Also, thank you for tackling the heterogeneous inflow function simultaneously!

FYI, the `tricontourf` visualization call does not use the `levels` handle, so it plots a small number of contour fields. This makes it a bit difficult to see the heterogeneous flow fields. Here are the plots again with `levels=100`:

image

I'd like to have a separate handle for the `levels` in `tricontourf`, but not urgent.

A couple thoughts.
1. I really like taking out the nearest neighbor part of the heterogeneous inflows. I like leaving those decisions with the user to avoid ambiguity.
2. I love plotting the heterogeneous bounds. This makes it very clear to the user what is defined and how.
3. The plot above shows a discrete jump in the background flow field on the border of the heterogeneous bounds. I think this is fine and expected behavior, but users may not be aware of this or may not fully understand the consequences. Perhaps the logger can raise a warning to the user that the heterogeneous bounds do not fully cover the simulation domain? If users insist on having `1.0` in the background flow and don't want the warning raised every time, they can just expand their heterogeneous bounds to cover that to avoid the warning, but then it's up to the user to consciously do this.
4. Would it make sense to place `het_map` under `reinitialize()` rather than when initializing the `FlorisInterface` class? This is where the wind directions and wind speeds, wind shear, etc. are also reinitialized, so it makes sense to me to group them together.
5. The speed-ups are now wind-direction and wind-speed agnostic. Would it be possible to incorporate this dependency? For example, if `het_map` is a 1D array, then the same values are applied for all wind directions and wind speeds. If `het_map` is a 3D array, with shape (n_wind_directions, n_wind_speeds, n_points), it is applied per wind direction and wind speed. I think this would further vouch for making `het_map` part of `reinitialize()`, since any change in wind direction or wind speed would warrant an update in `het_map`.

bayc:
Ok, another round of changes after your feedback @Bartdoekemeijer and discussions with @rafmudaf. 

I added the `clevels` functionality to the visualization of the heterogeneous flow fields, good call.

Replying to numbered commentes:
3. Added the warning.
4. After some thinking and talking with Raf, I re-configured the heterogeneous interface again to hopefully simplify the setup for both users and developers. Now the `het_config` dictionary is defined in the input file/dictionary. This, along with some other changes, allowed me to move `het_map` onto `FlowField`, which now nicely contains all the heterogeneous info and calculations, and we don't have to worry about storing and resetting `het_map`, as it is dumped with the dictionary and reinitialized. Also, users can now pass an updated `het_config` dictionary to `fi.reinitialize`, which is handy.
5. I'm not sure I agree with this one, in that I don't think the speed ups became wind direction agnostic. It is still required to define a set of speed ups for each wind direction. We originally adopted this setup as it is 1) more explicit, which has been a direction for v3, and 2) how we were told it is done in industry. I did spend some time trying to allow for the broadcasting of one array of speeds ups across multiple wind directions, but ran into some undetermined difficulties/weird calculations and decided to keep the current functionality. I could be convinced otherwise, but for now opted to keep it so that a speed up array is required for each wind direction.

I also wanted to note with my first round of commits, Raf and I talked about future refactoring of the visualization routines and the current use of `visualize_heterogeneous_cut_plane` is just a step in that direction and will eventually be wrapped in a higher-level user function so that they don't have to worry about which routine to use.

Please check out the new features with the previously posted test code as well as examples 16 and 16b, looking forward to your feedback!

examples/inputs/gch_heterogeneous_inflow.yaml
```diff
@@ -0,0 +1,102 @@
...
+  layout_y:
+  - 0.0
+  - 0.0
+  - 0.0
+  turbine_type:
+  - nrel_5MW
+
+flow_field:
+  air_density: 1.225
+  het_config:
```
bayc, line 30: Note the new input file here with the `het_config` definition.

Bartdoekemeijer:
Thanks for the quick turn-around on this @bayc !! Also thanks for the inclusion of `clevels`. Moving `het_map` to `FlowField` makes a lot of sense to me, and having it saved in the `.yaml` file and as an option in `reinitialize` is going to be super helpful for my work! The examples also look good to me.

> I don't think the speed ups became wind direction agnostic. 

Ahh right, I had forgotten that this is already in there! Great!! Could I persuade you to add another dimension to accommodate wind speed dependency? Namely, I imagine wind speed dependency may perhaps not be very common if heterogeneity comes from terrain effects (guessing that's where the industry recommendation comes from), but wind speed dependency may be more relevant when we're talking about inflow heterogeneity due to the presence of neighboring farms.

floris/utilities.py
```diff
@@ -230,11 +232,8 @@ def rotate_coordinates_rel_west(
         wind_directions (NDArrayFloat): Series of wind directions to base the rotation
         coordinates (NDArrayFloat): Series of coordinates to rotate with shape (N coordinates, 3)
             so that each element of the array coordinates[i] yields a three-component coordinate
```
rafmudaf, line 234: Need to update the docstring to reflect new input arguments and a different output type

floris/utilities.py
```diff
@@ -260,7 +261,50 @@ def rotate_coordinates_rel_west(
         + y_center_of_rotation
     )
     z_coord_rotated = np.ones_like(wind_deviation_from_west) * z_coordinates
-    return x_coord_rotated, y_coord_rotated, z_coord_rotated
+    return x_coord_rotated, y_coord_rotated, z_coord_rotated, x_center_of_rotation, \
+        y_center_of_rotation
+
+
+def reverse_rotate_coordinates_rel_west(
```
rafmudaf, line 268: Please add docstring. It could be similar to the one above.

floris/tools/visualization.py
```diff
@@ -109,12 +124,14 @@ def add_turbine_id_labels(fi: FlorisInterface, ax: plt.Axes, **kwargs):
         fi (FlorisInterface): Simulation object to get the layout and index information.
         ax (plt.Axes): Axes object to add the labels.
     """
+
+    # Rotate layout to intertial frame for plotting turbines relative to wind direction
```
rafmudaf, line 128: intertial -> inertial

floris/utilities.py
bayc: Updated. Not sure how to best post the updates here?

floris/utilities.py
bayc: Updated!

floris/tools/visualization.py
bayc: Fixed, thanks!

bayc:
> > I don't think the speed ups became wind direction agnostic.
> 
> Ahh right, I had forgotten that this is already in there! Great!! Could I persuade you to add another dimension to accommodate wind speed dependency? Namely, I imagine wind speed dependency may perhaps not be very common if heterogeneity comes from terrain effects (guessing that's where the industry recommendation comes from), but wind speed dependency may be more relevant when we're talking about inflow heterogeneity due to the presence of neighboring farms.

Hmm, interesting. Are you thinking that the inflow from upstream wind farms would cause different speed up multipliers in the downstream farm, depending on what the freestream, and thus wake deficits, are? I agree that the inflow would be different, but am unsure if it would cause the speed up multipliers to change over a range of wind speeds.

Bartdoekemeijer:
> Hmm, interesting. Are you thinking that the inflow from upstream wind farms would cause different speed up multipliers in the downstream farm, depending on what the freestream, and thus wake deficits, are? I agree that the inflow would be different, but am unsure if it would cause the speed up multipliers to change over a range of wind speeds.

That's right. The higher the ambient wind speed, the lower the thrust coefficient of neighboring farms, and `ct` is a large driver in the depth of wakes and blockage effects. To demonstrate this, here's a simple example showing the effect of a neighboring wind farm's wakes (11 kilometers away) on the inflow wind speed of 5 turbines, for a wind direction of 270 deg:

```
import numpy as np

def load_floris():
    fi = FlorisInterface("inputs/turbopark.yaml")

    # Specify the full wind farm layout: nominal and neighboring wind farms
    X, Y = np.meshgrid(np.arange(0, 5) * 800.0, np.arange(0, 5) * 800.0)
    X = X.flatten()
    Y = Y.flatten()

    distance = 15.0e3
    Y = np.hstack([Y[X==0], Y + 150 * np.random.rand(*np.shape(X))])
    X = np.hstack([X[X==0], X - distance + 300 * np.random.rand(*np.shape(X))])

    # Turbine weights: we want to only optimize for the first 10 turbines
    turbine_weights = np.zeros(len(X), dtype=int)
    turbine_weights[0:5] = 1.0

    # Now reinitialize FLORIS layout
    fi.reinitialize(layout_x=X, layout_y=Y)

    # And visualize the floris layout
    fig, ax = plt.subplots()
    ax.plot(X[turbine_weights == 0], Y[turbine_weights == 0], 'ro', label="Neighboring farms")
    ax.plot(X[turbine_weights == 1], Y[turbine_weights == 1], 'go', label='Farm subset')
    ax.grid(True)
    ax.set_xlabel("x coordinate (m)")
    ax.set_ylabel("y coordinate (m)")
    ax.axis("equal")
    ax.legend()

    return fi, turbine_weights


if __name__ == "__main__":
    fi, turbine_weights = load_floris()
    fig, ax = plt.subplots()
    for u_inf in range(6, 20):
        fi.reinitialize(
            wind_speeds=[u_inf],
            turbulence_intensity=0.05,
            wind_directions=[270.0],
        )
        fi.calculate_wake()
        
        speedup_map = fi.floris.flow_field.u.mean(axis=3)[0, 0, 0:5, 0] / u_inf
        ax.plot(range(len(speedup_map)), speedup_map, "-o", label=f"Ambient wind speed: {u_inf} m/s")

    ax.set_xlabel("Turbine number (-)")
    ax.set_xticks(range(5))
    ax.set_ylabel("Heterogeneous inflow map (-)")
    ax.grid(True)
    ax.legend()

    plt.show()
```

image

image

This confirms that, the higher the wind speed, the lower `ct` and thus the lower neighboring farm effects. I expect the same correlation for blockage effects. Note that these effects are generally much larger with TurbOPark than with GCH or CC, but the same principles carry over (just to a smaller effect).

bayc:
Wow, thanks for sharing that @Bartdoekemeijer! That is really interesting. I believe the internal consensus over here to is close out this PR where it is, and open up a new issue to tackle adding the heterogeneous input dependency on wind speed.

floris/simulation/flow_field.py
```diff
@@ -69,6 +76,40 @@ def wind_directions_validator(self, instance: attrs.Attribute, value: NDArrayFlo
         """Using the validator method to keep the `n_wind_directions` attribute up to date."""
         self.n_wind_directions = value.size
 
+    @het_config.validator
+    def het_config_validator(self, instance: attrs.Attribute, value: dict | None) -> None:
+        """Using the validator method to check that the het_config dictionary has the correct
+        key-value pairs.
+        """
+        if value is not None:
```
rafmudaf, line 84: @RHammond2 would value ever be None here?

floris/simulation/flow_field.py
```diff
@@ -69,6 +76,40 @@ def wind_directions_validator(self, instance: attrs.Attribute, value: NDArrayFlo
...
+                    raise ValueError(
+                        "het_config must contain entries for 'speeds_ups', 'x_locs', and 'y_locs',"
+                        f" with 'z_locs' as optional. Missing '{k}'."
+                    )
+            if "z_locs" not in value:
+                # If only a 2D case, add "None" for the z locations
+                value["z_locs"] = None
+
+    @het_map.validator
+    def het_map_validator(self, instance: attrs.Attribute, value: list | None) -> None:
```
rafmudaf, line 97: Same as above

floris/simulation/flow_field.py
```diff
@@ -36,6 +42,7 @@ class FlowField(FromDictMixin):
     turbulence_intensity: float = field(converter=float)
     reference_wind_height: float = field(converter=float)
     time_series : bool = field(default=False)
+    het_config: dict = field(default=None)
```
rafmudaf, line 45: I think `het_config` is yet obscure if you don't already know that this is a thing. Instead, I propose to change the name to something like `speed_ups_map`. @bayc @misi9170 @paulf81 what do you think?

paulf81: Agree that this is a more descriptive name and will more clearly indicate its purpose, I say switch it

floris/tools/visualization.py
```diff
@@ -71,27 +73,40 @@ def plot_turbines(
...
+        color (str, optional): Color to plot turbines. Defaults to None.
+        wd (list, optional): The wind direction to plot the turbines relative to. Defaults to None.
+        yaw_angles (NDArray, optional): The yaw angles for the turbines. Defaults to None.
     """
     if not ax:
         fig, ax = plt.subplots()
     if yaw_angles is None:
         yaw_angles = fi.floris.farm.yaw_angles
+    if wd is None:
+        wd = fi.floris.flow_field.wind_directions[0]
```
rafmudaf, line 99: This may be incorrect given the line in `plot_turbines` function call below that also extracts the first wind direction. Double check @bayc

floris/simulation/flow_field.py
misi9170: My only reservation is that we use the term "heterogeneous" pretty regularly throughout the code, so I kind of like having some indication that this is related to heterogeneity in the name. What do you think about `heterogeneous_speed_ups_map`? A little long and clunky...

floris/simulation/grid.py
```diff
@@ -73,12 +77,12 @@ class Grid(ABC):
     n_wind_speeds: int = field(init=False)
     n_wind_directions: int = field(init=False)
     turbine_coordinates_array: NDArrayFloat = field(init=False)
-    x: NDArrayFloat = field(init=False, default=[])
-    y: NDArrayFloat = field(init=False, default=[])
-    z: NDArrayFloat = field(init=False, default=[])
     x_sorted: NDArrayFloat = field(init=False)
     y_sorted: NDArrayFloat = field(init=False)
     z_sorted: NDArrayFloat = field(init=False)
+    x_sorted_inertial_frame: NDArrayFloat = field(init=False)
```
rafmudaf, line 83: These are added and `x` etc are deleted. As I understand it, `x/y/z_sorted_intertial_frame` are the same as `x/y/z` except that they're sorted. Will we need to retain `x/y/z` for postprocessing? For example, if you need to plot the layout with ID's based on your input, do we require to unsort `x/y/z_sorted_intertial_frame`?
2.

floris/simulation/grid.py
```diff
@@ -73,12 +77,12 @@ class Grid(ABC):
     n_wind_speeds: int = field(init=False)
     n_wind_directions: int = field(init=False)
     turbine_coordinates_array: NDArrayFloat = field(init=False)
-    x: NDArrayFloat = field(init=False, default=[])
-    y: NDArrayFloat = field(init=False, default=[])
-    z: NDArrayFloat = field(init=False, default=[])
     x_sorted: NDArrayFloat = field(init=False)
```
rafmudaf, line 80: Given the added complexity in the grids, this is a good time to document the `_sorted` and `_sorted_inertial_frame` variables. The minimum documentation would here in the docstring, and better would be to also add to the Floris_101 notebook in the docs with detailed exaplanation and code or analysis of these variables.

floris/simulation/grid.py
rafmudaf: A follow on question is whether this should be simply `x/y/z` with those attributes being defined (in words) the same as you would define the `x/y/z_sorted_intertial_frame` attributes.

floris/tools/visualization.py
```diff
@@ -71,27 +73,40 @@ def plot_turbines(
...
+    # Rotate yaw angles to inertial frame for plotting turbines relative to wind direction
+    yaw_angles = yaw_angles - wind_delta(np.array(wd))
+
     plot_turbines(
         ax,
         fi.layout_x,
         fi.layout_y,
-        yaw_angles[0, 0],
-        fi.floris.farm.rotor_diameters[0, 0],
+        yaw_angles.flatten(),
```
rafmudaf, line 108: Whats the shape of `yaw_angles` here? Previously, it was assumed that it was 3D (wind direction, wind speed, N turbines). Here, it looks like it assumed this is 1D (N turbines). Same question for `fi.floris.farm.rotor_diameters` below.

floris/tools/visualization.py
rafmudaf: Oh I see now that this is only used to adjust the yaw angles. That's probably good to reflect in the doc string

rafmudaf:
Overall, this looks like it's in pretty good shape!

floris/simulation/flow_field.py
RHammond2: If no `het_config` is passed, then yes, because `None` is the default value. Maybe the question is could the control flow be switch to do the following to clean up the indentation?
```python
if value is None:
    return

# code validation
````

File: floris/simulation/flow_field.py (hunk 8 above)
Comment (RHammond2) [reply floris/simulation/flow_field.py:97]: Same comment. Maybe since this isn't a passed variable you could use the following in the attribute definition, which creates a new list as the default value. However, with `init=False`, and a user not providing an actual value to this, then you'd need to check that it's not an empty list, which I think is preferable.
```python
het_map: list = field(init=False, factory=list, on_setattr=attrs.setters.validate)

@het_map.validator
def het_map_validator(self, instance, value):
    if value == []:
        return
    if self.n_wind_directions!= np.array(value).shape[0]:
        raise ValueError(
            "The het_map's wind direction dimension not equal to number of wind directions."
        )
```

floris/simulation/flow_field.py
rafmudaf: @misi9170 I like it. If we support heterogeneous wind directions, that will logically fit into `heterogenous_inflow_config` also. I think `x/y_loc` should be simply `x/y` since the `_loc` doesn't add much clarity. It's reasonably obvious that these are the locations of the speed multipliers.

Here's where we're at:
```python
heterogenous_inflow_config = {
    "speed_multipliers": [1.0, 2.0, 3.0],
    "x": [ -300.0, 300.0, -300.0, 300.0],
    "y": [ -300.0, 300.0, -300.0, 300.0]
}

File: floris/simulation/flow_field.py (hunk 7 above)
Comment (rafmudaf) [reply floris/simulation/flow_field.py:84]: Thanks for the suggestion to early-exit @RHammond2.

**Comment by rafmudaf** [comment]
For reference, I've updated @bayc's example from https://github.com/NREL/floris/pull/578#issuecomment-1534131964 for the new API in this pull request. The script is below and the plots that it creates are below that.

```python
import numpy as np
from floris.tools.visualization import plot_turbines_with_fi

speed_ups = [[1.2, 1.2, 1.0, 1.0]]
x_locs = [-300.0, -300.0, 2600.0, 2600.0]
y_locs = [ -300.0, 300.0, -300.0, 300.0]

# Generate the configuration dictionary to be used for the heterogeneous inflow.
het_config = {
    'speed_ups': speed_ups,
    'x_locs': x_locs,
    'y_locs': y_locs,
}

# Initialize FLORIS with a heterogeneous inflow
fi_2d = FlorisInterface("inputs/gch_heterogeneous_inflow.yaml")


# Plot the farm for the wind direction of 270 degrees
fig, ax = plt.subplots(2, 2)
yaw_angles = np.array([[[25.,0.,0.]]])
horizontal_plane_270 = fi_2d.calculate_horizontal_plane(
    wd=[270.],
    x_resolution=200,
    y_resolution=100,
    height=90.0,
    yaw_angles=yaw_angles,
)
wakeviz.visualize_heterogeneous_cut_plane(horizontal_plane_270, fi_2d, ax=ax[0][0], title="Horizontal")
plot_turbines_with_fi(
    fi_2d,
    ax=ax[0][0],
    color="black",
    wd=[270.],
    yaw_angles=yaw_angles,
)


# Plot the farm for the wind direction of 275 degrees
horizontal_plane_275 = fi_2d.calculate_horizontal_plane(
    wd=[275.],
    x_resolution=200,
    y_resolution=100,
    height=90.0,
    yaw_angles=yaw_angles,
)
wakeviz.visualize_heterogeneous_cut_plane(horizontal_plane_275, fi_2d, ax=ax[0][1], title="Horizontal")
plot_turbines_with_fi(
    fi_2d,
    ax=ax[0][1],
    color="black",
    wd=[275.],
    yaw_angles=yaw_angles,
)


# Plot the farm for the wind direction of 315 degrees
horizontal_plane_315 = fi_2d.calculate_horizontal_plane(
    wd=[315.],
    x_resolution=200,
    y_resolution=100,
    height=90.0,
    yaw_angles=yaw_angles,
)
wakeviz.visualize_heterogeneous_cut_plane(horizontal_plane_315, fi_2d, ax=ax[1][0], title="Horizontal")
plot_turbines_with_fi(
    fi_2d,
    ax=ax[1][0],
    color="black",
    wd=[315.],
    yaw_angles=yaw_angles,
)


# Plot the farm for the wind direction of 360 degrees
horizontal_plane_360 = fi_2d.calculate_horizontal_plane(
    wd=[360.],
    x_resolution=200,
    y_resolution=100,
    height=90.0,
    yaw_angles=yaw_angles,
)
wakeviz.visualize_heterogeneous_cut_plane(horizontal_plane_360, fi_2d, ax=ax[1][1], title="Horizontal", plot_het_bounds=False)
plot_turbines_with_fi(
    fi_2d,
    ax=ax[1][1],
    color="black",
    wd=[360.],
    yaw_angles=yaw_angles,
)

# Show the plots
plt.show()
```
het

Issue #572, opened 2023-02-09
Bug report - new TI sorting bugfix not compatible with fi.calculate_no_wake


mr483:
**Bug description**
Hi @bayc, I was just testing this bug fix out locally and found that it causes a numpy error when I try running fi.calculate_**no_wake**():

The error occurs at the new line in finalize in flow_field.py: 
```
    def finalize(self, unsorted_indices):
        self.u = np.take_along_axis(self.u_sorted, unsorted_indices, axis=2)
        self.v = np.take_along_axis(self.v_sorted, unsorted_indices, axis=2)
        self.w = np.take_along_axis(self.w_sorted, unsorted_indices, axis=2)

        self.turbulence_intensity_field = np.mean(np.take_along_axis(self.turbulence_intensity_field_sorted, unsorted_indices, axis=2), axis=(3,4))
```

IndexError: index 3 is out of bounds for axis 2 with size 1

The array self.turbulence_intensity_field_sorted attains the correct dimension in the code in the solver, but of course if no wake calculations are performed then this never occurs. 

My ugly fix was just the following (I check the turbine axis and if it still only has a length of 1 I assume a no-wake calculation was performed): 
```
        if self.turbulence_intensity_field_sorted.shape[2]!=1:
            self.turbulence_intensity_field = np.mean(np.take_along_axis(self.turbulence_intensity_field_sorted, unsorted_indices, axis=2), axis=(3,4))
```

rafmudaf:
In addition to addressing this bug, we should add a unit test for `FlorisInterface.calculate_no_wake` to catch these situations in the future.

PR #584, opened 2023-02-16
Bug fix: initialize TI array with correct dimensions

bayc:
This bug fix corrects the shape of the `turbulence_intensity_field` for use with the `calculate_no_wake` function.

#572 

`flow_field.py`

This PR also adds a unit test for `calculate_no_wake`.

The following code runs without error.

```

fi.calculate_no_wake()
TIs = fi.get_turbine_TIs()
```


tests/floris_interface_test.py
```diff
@@ -23,15 +23,90 @@ def test_calculate_wake():
...
+            fi.floris.farm.n_turbines
+        )
+    )
     fi.calculate_wake(yaw_angles=yaw_angles)
     assert fi.floris.farm.yaw_angles == yaw_angles
 
 
+def test_calculate_no_wake():
+    """
+    In FLORIS v3.2, running calculate_no_wake twice incorrectly set the yaw angles when the first
```
rafmudaf, line 49: Heads up to update this docstring

tests/floris_interface_test.py
```diff
@@ -23,15 +23,90 @@ def test_calculate_wake():
...
+        (
+            fi_base.floris.flow_field.n_wind_directions,
+            fi_base.floris.flow_field.n_wind_speeds,
+            fi_base.floris.farm.n_turbines
+        )
+    )
+    fi_base.floris.farm.yaw_angles = yaw_angles
+    fi_base.floris.flow_field.initialize_velocity_field(fi_base.floris.grid)
+    fi_base.floris.farm.initialize(fi_base.floris.grid.sorted_indices)
+    fi_base.floris.state.INITIALIZED
```
rafmudaf, line 86: I don't think we need this here.
Also, this isn't doing anything in `Floris`, so that's a good side effect of including it here! It's accessing the data but not writing it to anything.

tests/floris_interface_test.py
bayc: Updated, but still left a good portion in, as `calculate_no_wake` suffered from the same issue.

tests/floris_interface_test.py
bayc: I removed line 86 above, but the rest is needed to properly setup the `fi_base` object for comparison.

Issue #554, opened 2023-01-06
Turbine power not flat above rated


paulf81:
**Bug description**
Not sure it's an important bug, but it could be a problem for optimizers that the power of turbines is not flat wrt changes in wind speed above rated, but wiggles a little bit.  We tried to cut this down by interpolating on power and not cp, but some non-flatness is still evident

**To Reproduce**
Run the example python 18_check_turbine.py 
For the 5MW plot, zoom in on rated power
![image](https://user-images.githubusercontent.com/6117702/211097479-402061ec-ca6d-4bf6-aa6b-0e4255c6d671.png)

**Expected behavior**
I think this should be perfectly flat.  I can imagine how to do this, but what I have come up with so far would incur some time penalty so I don't have an idea I yet like so just leaving this issue to mark it

**Floris Version**
3.2

paulf81:
I think this issue is leading other problems (see Discussion #583 for example problems), maybe should get advanced toward a solution.  

One thought I had is that if we switched from specifying power in terms of Cp, over to Power, we can guarantee, this flatness.  But that might make scaling by rotor area less convenient.  

Maybe a solution would be to add some sort of check at the point we compute inner power in turbine.py

        inner_power = (
            0.5 * self.rotor_area
            * self.fCp_interp(wind_speeds)
            * self.generator_efficiency
            * wind_speeds ** 3
        )

Maybe to not allow differences below some threshold?  Like start from the highest wind speeds, and insist that if the next lowest wind speed is only different by some negligible amount, then force it to be the same?  I don't think this would impact compute time since it would be done at instantiation of the turbine object?  @Bartdoekemeijer @bayc @rafmudaf any thoughts?  I could test it out if you think it's in the right direction

rafmudaf:
I wonder if this would be better handled by "fixing" the input data rather than adding a check in the computation to handle an edge case. For example, the power curve could be smoothed in a preprocessing step.

paulf81:
I had an idea this morning that we could have the user fix out to 8-9 decimal places and then have FLORIS throw an error if the resulting curve isn't monitonic, but there would be a problem if the power curve really did ramp down at higher wind speeds

paulf81:
Closing this as can say resolved by PR #590

rafmudaf:
After additional discussion, we've concluded that this is *not* resolved by #590, and a more appropriate solution is to require that users define power instead of Cp for the turbine performance curve.

misi9170:
[...]

PR #590, opened 2023-02-28
Add a fix to inner_power that removes tiny wiggles

paulf81:
This could be ready to merge

This fix is meant to address the problem in Issue #554 (and I think along with it discussion #583).  The non-flat region 3 causes odd behaviors in the optimization.  I think one way we can address them is by removing very small changes in the inner_power at the time the turbine class is instantiated.

Issue #554 
Discussion #583 

turbine.py

If you now re-run the example 18_check_turbine.py as in Issue #554 , the small perturbations in power are not there


rafmudaf:
@paulf81 I support including a correction to the power curve such as this, but I'll push back on including this within `floris.simulation`. In general, I'd like to keep these types of changes to user-input at the outer-most level of the code, so I suggest to incorporate this into the `floris.tools` instead. The `simulation` package should strive to remain as general purpose as possible.

That being said, it's not immediately clear where this would fit in the current `floris.tools`, but it might have a role in the `floris.turbine_library`. Overall, I think this should be considered in upcoming work on user interface improvements instead. In that regard, I suggest to reopen #554 and close this pull request.

misi9170:
@paulf81 @rafmudaf An option could be to include it in turbine_utilities.py, which doesn't exist on the main or develop branches but is in the pull request #729 . However, #729 may also not have a place with planned developments of turbine.py.

rafmudaf:
@misi9170 I like that suggestion as it starts to build a library around dealing with turbine inputs.

One workflow might be to do a preprocessing step where the power curve is smoothed and exported either as a new turbine input file or a turbine dictionary for use in the rest of the script. Something along the lines of the code below.

```python
import floris.turbine_library as turbine_library

fi = FlorisInterface(...)

# Show the wiggles
turbine_library.plot_power_curve(fi.turbine[0])

# Smooth the wiggles
corrected_turbine = turbine_library.turbine_utilities.smooth_power_curve(fi.turbine[0])
# Question: what is the type of corrected_turbine - Turbine or dict?

# Show that they're smoothed
turbine_library.plot_power_curve(corrected_turbine)

# Put this back into the simulation object
fi.reinitialize(turbine_type=corrected_turbine)
```

paulf81:
I agree with all of this, and I think putting this correction as something to be more sensibly included in a re-worked turbine workflow, rather than forced into simulation.py will work great for me.  Would you like me then to close this pull request @rafmudaf ?  And re-open #554 ?

Bartdoekemeijer:
To fundamentally resolve this, would it make sense to replace the power coefficient table in the turbine input file by an absolute power table? This would be a cleaner solution than adding some kind of smoothing function.

paulf81:
Great minds think alike!  This is the proposal for 4.0, and agree this is a simpler solution

PR #599, opened 2023-03-03
Use 3x3 turbine grid in regression tests

rafmudaf:
This pull request changes the regression test infrastructure and baseline solutions to use a 3x3 grid on the turbines. The intention is not to reduce the resolution of the calculation, but instead to allowed to test `FlorisInterface` as that enforces a maximum 3x3 grid since https://github.com/NREL/floris/pull/472.

This is required in order to adequately test #584 since that will require testing FlorisInterface through the regression tests.

Regression tests only.

The baseline results for the regression tests do change. However, the only change here is the setting for the turbine grid. While typically moving to a lower resolution will provide a less accurate answer, the consideration for the regression tests is whether the results change from commit to commit. In this case, we are intentionally changing the results with the expectation that subsequent code changes will be able to add more tests but not change the results.


rafmudaf:
@bayc feel free to review this and let me know if you have any concerns. However, I mostly wanted to ping you as a reviewer just so you're aware of this change. For most developers interacting with floris, this will not have much of an impact.

PR #613, opened 2023-03-29
Various changes to code and documentation

rafmudaf:
# Collection of random changes and improvements
This pull request includes a variety of changes and improvements that don't easily fall under other ongoing work. The changes are listed below.

* 25e380da - Replace hard-coded rotor diameter with model value
* 2a309131 - Add all releases to quality metric plots
* dfc00c23 - Remove redundant array slice operations
* 362d97ce - Disable GCH components for Jensen example input
* afd7dfc5 - Simplify time-series selection in FlorisInterface
* bc6a281f - Remove unused turbine fCps array
* a02e68a3 - Use dict over array for turbine property map (see https://github.com/NREL/floris/discussions/594)
* 9dc507f9 - Remove redundant math in wind direction delta
* a888e41b - Fix and improve a few docstrings
* 9b856f70 - git-ignore Ruff cache files
* 27b5ca63 - Add info about GitHub form comments

These changes only impact the quality of the code itself. There are no API or functionality changes.

## Related issue
No issue, but there are some comments:
- https://github.com/NREL/floris/discussions/594
- https://github.com/NREL/floris/commit/e7fe3fde624e92408c37ab11b408f90b9e70639f#r99212494

The code within `FlorisInterface`, the solvers, `Farm`, `FlowField`, and `utilities.wind_delta` are modified, but the functionality and results in all of these modules should not change.

Here's a listing of where the changes are tested.

* 9dc507f9 - Remove redundant math in wind direction delta
  - `utilities_unit_test` already has a test for this
* a02e68a3 - Use dict over array for turbine property map (see https://github.com/NREL/floris/discussions/594)
  - tested in `turbine_unit_test`
* bc6a281f - Remove unused turbine fCps array
  - tested in regression tests; the results are unchanged and the code runs so this is a valid change
* afd7dfc5 - Simplify time-series selection in FlorisInterface
  - See the attached output below from the time series example (21) before and after this change
* dfc00c23 - Remove redundant array slice operations
  - See below
* 25e380da - Replace hard-coded rotor diameter with model value
  - [ ] TODO: Cumulative-Curl regression tests value are unchanged, but this should be tested with a wind farm with turbines of different rotor diameters; @bayc do we have such a test?

### Time series switch

These plots are the results from the time series example (21). The first is before the change to the time series flag and the second is after the change. They are different because that example creates a random-walk for the wind directions. However, this show that the time series flag is enabled correctly, and this verifies the change.

### Redundant array slices

This is verified with the same example as above, 21. The plot below shows that the code runs, as expected, so removing the extra slices did not impact the shape of the arrays.


bayc:
@rafmudaf Thanks for calling these out specifically! Responses are in italics below.

* b4af5943e14d95b3e10dbffe2b0f3eaa5025ea47 - Replace hard-coded rotor diameter with model value
  * Is `rotor_diameter_i` correct? _Yes, that is correct._
* bd626f6cb04e7cda65c4c3dba4ac58e1f08dfa92 - Remove redundant array slice operations
  * Sanity check. _That checks out on my end._
* a84d24eaea9ee65e7b3ed4b08e698cb26efaec6f - Remove unused turbine fCps array
  * Sanity check. _I had added this one because I've needed to get Cp values in the past for certain analyses/customers at each turbine, but didn't add a method to `floris_interface` to get turbine Cps. So I think this would mainly be useful for that._
* 38655da6b70e34fc0467cacef08bb265a5c75ce8 - Use dict over array for turbine property map (see https://github.com/NREL/floris/discussions/594)
  * Sanity check. _Good catch! Seems right to me._
* 1ab795272dffc0501a52769c0335ab2d5f7d937c - Remove redundant math in wind direction delta
  * Sanity check. _Also looks good!_

bayc (approved):

rafmudaf:
Thanks for the review @bayc. Is it worth retaining the `fCp` arrays then?

Issue #609, opened 2023-03-20
Bug report: Incorrect error message raised when non-existent turbine specified


misi9170:
##LOW PRIORITY

`check_turbine_type()` in farm.py raises a confusing error when a turbine model is specified that is not available in the library. The error raised is `ValueError: The turbine type: <non-existent_turbine> exists in both the internal and external turbine library.`, whereas I would expect an error along the lines of `ValueError: The turbine type: <non-existent_turbine> does not exist in the turbine library.`

Note that this bug does not affect running code, and can be treated as low priority.

## Steps to produce
1. Change the `turbine_type` field of examples/inputs/gch.yaml from `nrel_5MW` to `bad_turbine_type`.
2. Run examples/01_opening_floris_computing_power.py

PR #622, opened 2023-04-04
Bug Fix: Correct a misleading error message

RHammond2:
# Bug Fix: Correct a misleading error message for external library checking

This PR is a small correction to an error message that was duplicated, but left unchanged, that incorrectly states an unfound file exists in both locations. The new error is now a `FileNotFoundError` and indicates that the turbine file unable to be found in either the internal or external turbine library. Additionally, I've added a test to ensure the correct error is raised when a turbine can't be located.

## Related issue

N/A: I made the fix instead of submitting an issue and a fix.

- `floris/simulation/farm.py`
  - `Farm.check_turbine_type()`
- `tests/farm_unit_test.py`
  - `test_farm_external_library()`

The duplication of the error is extremely confusing when you're working with an external library that is missing a file, but it indicates that it's located in both libraries. After spending too much time getting to the bottom of it, I figured I'd make someone else's life easier in the process instead of patching it in my own workflow.

All the tests pass.

### DIFF
--- a/floris/simulation/farm.py
+++ b/floris/simulation/farm.py
@@ -133,8 +133,8 @@ class Farm(BaseClass):
                     full_path = external_fn
                 else:
                     raise ValueError(
-                        f"The turbine type: {turbine} exists in both the internal and external"
-                        " turbine library."
+                        f"The turbine type: {turbine} does not exist in either the internal or"
+                        " external turbine library."
                     )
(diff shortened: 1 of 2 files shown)


rafmudaf (approved):

rafmudaf:
Good catch @RHammond2 and thanks for adding the test!

rafmudaf:
Closed #609

Issue #624, opened 2023-04-06
Implement yaw impact of power and thrust via Heck et al


paulf81:
# Discussion on FLASC priorities for code sprint

The paper: https://www.cambridge.org/core/journals/journal-of-fluid-mechanics/article/modelling-the-induction-thrust-and-power-of-a-yawmisaligned-actuator-disk/3A34FC48A6BC52A78B6D221C13F4FC3A

by K.S. Heck, H.M. Johlas, and M.F. Howland proposes a replacement to the cos^(Pp, Pt) approach to modeling the effect of yaw misalignment accounting for changes in axial induction.  Propose to implement in FLORIS.

Note this implementation will take the simplifying assumption that thrust is not being actively controlled to optimize against yaw, which is currently typical of wake steering, but leaves open future opportunities to optimized thurst.

## Proposed solution
Add the new model into turbine.py, and configure it so that it works with the yaw_effective_velocity approach to ensure consistency of above rated conditions.

paulf81:
Pushing to v4.1, in our current thinking (@misi9170  double checking) v4.0 provides the framework / template and then this specific implementation is added in 4.1

paulf81:
Now planning to address in #797

misi9170:
Addressed in #924

PR #625, opened 2023-04-18
Feature/implement mit yaw

paulf81:
# First pass computing yaw loss the way discussed with MIke Howland

This pull request offers a first very hacky implementation of the method of computing yaw loss from the paper:

I add a quick calculation of the "Version 1a" calculation of ct-prime and than power loss, assuming yaw doesn't impact ct-prime.  I don't apply the cube so that, as with pP, the calculation can be applied to wind speed and accomodate above rated conditions.

I'm temporarily calling this the MIT method but that is just a place-holder name

## Related issue
Issue #624 

Most changes to turbine.py, but related changes to floris.py, floris_interface.py, solver.py

Include a small example file examples/18b_test_mit_method.py, which gives the following plot, comparing the yaw loss with assuming default Pp behavior and showing similar behavior

image


paulf81:
Moving to pull request #638 following updates in turbine.py following updates for floating/tilt (fyi @mhowlan3)

paulf81:
Closing this one

PR #638, opened 2023-04-25
Add MIT yaw correction method

paulf81:
# SECOND pass computing yaw loss the way discussed with MIke Howland

(Redoing this after merging in the changes for floating cp/ct)

This pull request offers a first very hacky implementation of the method of computing yaw loss from the paper:

I add a quick calculation of the "Version 1a" calculation of ct-prime and than power loss, assuming yaw doesn't impact ct-prime.  I don't apply the cube so that, as with pP, the calculation can be applied to wind speed and accomodate above rated conditions.

I'm temporarily calling this the MIT method but that is just a place-holder name

## Related issue
Issue #624 

Most changes to turbine.py, but related changes to floris.py, floris_interface.py, solver.py

Include a small example file examples/18b_test_mit_method.py, which gives the following plot, comparing the yaw loss with assuming default Pp behavior and showing similar behavior

image


paulf81:
@mhowlan3 this pull replaces the old one following updates to new turbine.py structure

sTamaroTum:
Hi all,

In Munich we have recently published a preprint describing a model we have developed to predict the effect of yaw and tilt on power and thrust. We have also modeled the asymmetrical effect of inflow shear (that seem to favor positive yaw angles for maximum power).

The preprint is available here: https://wes.copernicus.org/preprints/wes-2023-133/ . We have also attempted an implementation into floris, with some modifications, available here: https://doi.org/10.5281/zenodo.8388901 _(main/figs_14_15/floris)_.
Briefly, our model requires lookup tables C_P(pitch,tsr) and C_T(pitch,tsr), as well as the pitch and tsr as inputs. The outputs are the power and thrust based on yaw, tilt and inflow shear.

The code can also be tested in a Jupyter notebook here: https://tinyurl.com/btcl-main , at _(main/figs_14_15/fig14.ipynb)_.

I hope this helps in the discussion!

paulf81:
This is very helpful to see @sTamaroTum , thank you for sending!  We're currently road-mapping are next batch of efforts and will plan to incorporate this.  @rafmudaf  @christiannvaughn

paulf81:
note development here is now moving to #797

paulf81:
Closing as replace by #924

PR #646, opened 2023-05-05
Improve error handling and robustness in turbine model loads

rafmudaf:
# Redesign the turbine model loading routine to handle errors and be more robust
An issue was noted were the `!include turbine.yaml` method of including a turbine model was failing. This was due to how the Farm class was supporting the various methods of loading a turbine model. Turbine definitions could be loaded in three ways:
1. Pass a string that matches a turbine included in Floris's turbine library
2. Pass a dictionary representation of a turbine definition
  - Note that there's an option to use the yaml keyword "!include" which results in the yaml library preprocessing the inputs and loading the specified file directly into main input file. The result is that floris sees the turbine definition as a dict.
3. A string selecting an turbine that exists in an external turbine library specified in `turbine_library_path`

Currently, if the second option is used, the `Farm.turbine_type` is a list of dictionaries. However, dictionaries cannot be hashed (i.e. compared) so the subsequent `np.unique` command in [Farm's turbine loading](https://github.com/NREL/floris/blob/develop/floris/simulation/farm.py#L108) fails. While it may have been possible to find a workaround, I decided to split the work to load the turbine definitions into distinct steps. This pull request does the following to Farm:

- Check turbine_types with an attrs validator to raise an error if it is not a valid length
- Enforce type checking with an attrs field-validator for turbine_type and turbine_definitions
- Creates a turbine definition cache to avoid doing file I/o repeatedly if a definition has already been loaded
- Treats turbine_types as immutable so that exporting Farm preserves the input data as it was given
- Adds a unit test for the turbine definition handling

Bonuses:
- Remove all mention of JSON except from the legacy input reader
- Expand unit tests for FromDictMixin
- Let `load_yaml` function fail if the data given is invalid
- Consolidate floating and land-based test turbine definitions

## Related issue
I don't see an open issue for this.

Primarily `farm.py`. Other changes are cosmetic, related to removing JSON, or testing

All tests pass + new tests are added


floris/simulation/farm.py
```diff
@@ -83,7 +84,81 @@ class Farm(BaseClass):
...
+        # for mapping turbines later.
+        # We use a private variable here instead of self.turbine_type because self.turbine_type
+        # should always retain the input data. When this class is exported as_dict, the input
+        # types must be used. If we modify that directly and change its shape, recreating this
+        # class with a different layout but not a new self.turbine_type could cause the data
+        # to be out of sync.
+        _turbine_types = [
+            copy.deepcopy(t["turbine_type"]) if isinstance(t, dict) else t
+            for t in self.turbine_type
+        ]
```
misi9170, line 153: How do you feel about inserting the following two lines here (after line 153), so that `turbine_types` always ends up being a list of strings after this method is done?
```
        # Replace turbine_type list with all strings version
        self.turbine_type = _turbine_types.copy()
```

misi9170 (approved):
@rafmudaf I've put in one suggestion to think about, but it isn't critical to functionality. Everything runs as expected now.

floris/simulation/farm.py
misi9170: One quirk that this introduces is that `fi.reinitialize()` takes an optional argument `turbine_type`, which I think can be a list of strings or dicts (confirming now); and if the `Farm` attribute `turbine_type` ends up being a list of strings only, even when `fi.reinitialize()` is supplied with dicts, this _could_ be confusing to the user.

rafmudaf: We shouldn't modify `self.turbine_type`. Since it is used in the `as_dict` method to export the Farm object, it needs to remain as input. The reason is because if we modify it, then export Farm but don't overwrite this attribute and do change other things that depend on it like `layout_x`, the data is out of sync and there's no way to resolve it. That's why I instead created the new local / private variable.

Where do we need to use `Farm.turbine_type` other than in this context? In other words, what's the need to make the elements of that list consistent? I'm wondering if there's another way to achieve the same thing.

misi9170: Ah fair, got it. The reason (for me) is that the exported Farm object ends up having this sort of odd structure of having attribute `turbine_type` which is a list of dicts, with each dict containing a key `"turbine_type"`; and the `turbine_type` attribute is then identical to the `turbine_definitions` attribute. But that doesn't create an actual runtime issue anywhere, just take a little bit to understand if exporting the configuration to a dictionary, changing things, and reinstantiating with the altered dictionary. As an example, I had a farm with mutliple turbine types (and many of each turbine type), and I wanted to change the pP value for just one of the turbine types. It was doable, but a bit confusing with pP exists in both the `"turbine_type"` and `"turbine_definition"` keys of the farm dict (in the end, changing the pP value in the `"turbine_type"` dict worked and overwrote those in `"turbine_definition"`, but I basically just got lucky)

misi9170: UPDATE: @rafmudaf and I spoke about this, and I had this wrong: "turbine_definitions" is _not_ an available key in the "farm" subdictionary of an exported FLORIS dictionary---only "turbine_type" is---so there is no extra confusion created (past "turbine_type" possibly being a list of dicts, each containing a key "turbine_type"). The code should be left as is without adding the lines I suggested. @rafmudaf, feel free to mark this conversation as resolved.

floris/simulation/floris.py
```diff
@@ -225,48 +224,24 @@ def finalize(self):
...
-            filetype (str): The type to export: [YAML | JSON]
         """
         with open(output_file_path, "w+") as f:
-            if filetype.lower() == "yaml":
-                yaml.dump(self.as_dict(), f, default_flow_style=False)
-            elif filetype.lower() == "json":
-                json.dump(self.as_dict(), f, indent=2, sort_keys=False)
-            else:
-                raise ValueError("Supported export filetypes are JSON and YAML")
+            yaml.dump(self.as_dict(), f, default_flow_style=False)
```
RHammond2, line 247: I think we just need to update the keyword arguments to include `sort_keys=False`, to keep the same behavior. It's not in the original version, but I believe it should be, especially since the JSON had it, which was previously the more used form.

rafmudaf: Yep good catch

RHammond2 (approved):
This looks good to me, especially the additional commentary on what's going on in each step. I was originally unsure of the change in control flow, but agree that with the expanded commentary it's worthwhile.

I would just add that one keyword argument, and then this is good to go from my perspective.

floris/simulation/floris.py
rafmudaf: Fixed. Thanks for looking through this

rafmudaf:
Thanks for the reviews @RHammond2 @misi9170 @paulf81

PR #647, opened 2023-05-05
Print error when turbine models are missing reference values

paulf81:
# Add informative error about new input requirements

This pull request adds an informative error to explain why new parameters are now required to be in the turbine input file.  

Think it's mostly ready except on my computer print out of message is oddly spaced

farm.py
Can test by deleting either ref_density_cp_ct or ref_tilt_cp_ct from floris/turbine_library/nrel_5MW.yaml and running example examples/01_opening_floris_computing_power.py


rafmudaf:
Just heads up a very similar error message was recently removed in #603 ([see this line](https://github.com/NREL/floris/pull/603/files#diff-750f702579a3aa6f23bd0bab073c17d9b437697d00559434d8c624c87ac6d5fcL158)).

In my opinion, this sort of messaging belongs either in the documentation or in a very high level, user facing area of the code. Doing input validation in various places throughout the low level code adds noise to the context of the low level functionality and makes tracking these things down more difficult. You also know my opinion about printing these sorts of things to the console - it should be used very sparingly because even a little too much information and it just gets lost in excess output.

A good smell test is to consider what the function called `construct_turbine_map` is doing, as in what's its scope. Before this change, it was a one-liner that created a list of Turbine instances for every turbine modeled in the farm. After this change, it does that plus it check for valid input and communicates to the user.

I suggest either removing this in favor of an API change page in the online documentation or including these checks in a high level portion of the code where other similar input validation is happening. One place we've been aggregating these is in the [FlorisInterface.__init__](https://github.com/NREL/floris/blob/main/floris/tools/floris_interface.py#L70) routine.

paulf81:
These are good comments @rafmudaf I'll make some changes and resubmit

paulf81:
ok!  have a look now @rafmudaf ?  I couldn't move it all the way up to floris interface, but I could go up to floris.py, and I tried to not muddle any exisiting functions by making this step clearly it's own thing.  I know I'm still using a console output, but my thinking is a console error will happen if I don't do this to, so maybe this solution is splitting the difference?

rafmudaf:
@paulf81 I've made a few changes here. Primarily, instead of triggering an error and catching it, I'm testing whether the key exists in the turbine dictionary. Then, the error messages are aggregates and sent in one one error message. I also added the turbine name. This allows for one message to include all things missing from the input file. One thing we could add is to aggregate these for all turbines, so it would mean raising the error outside of the for-loop. That might be useful, and it's also easy to do. I can go either way on that.

I also added a page in the documentation to describe the turbine input file and pointed to that in the error message that is raised.

rafmudaf (approved):

paulf81:
> 

I think these changes are a big improvement, thank you!!

rafmudaf:
The two errors are reported like this:
```
ValueError: nrel_5MW turbine model
From FLORIS v3.2, the turbine definition must include 'ref_density_cp_ct'. This value represents the air density at which the provided Cp and Ct curves are defined. Previously, this was assumed to be 1.225 kg/m^3, and other air density values applied were assumed to be a deviation from the defined level. FLORIS now requires the user to explicitly define the reference density. Add 'ref_density_cp_ct' to your turbine definition and try again. For a description of the turbine inputs, see https://nrel.github.io/floris/input_reference_turbine.html.

From FLORIS v3.4, the turbine definition must include 'ref_tilt_cp_ct'. This value represents the tilt angle at which the provided Cp and Ct curves are defined. Add 'ref_tilt_cp_ct' to your turbine definition and try again. For a description of the turbine inputs, see https://nrel.github.io/floris/input_reference_turbine.html.
```

Issue #499, opened 2022-09-15
Bug report: default of turbine_grid_points=3 not close to convergence


Bartdoekemeijer:
**Bug description**
Setting `turbine_grid_points = 3` gives fairly coarse results in densely spaced farms and is not necessarily close to convergence. The error can be in the order of 5%. The variable `turbine_grid_points` must be set to `10` or higher to get close to convergence. Is this the expected behavior?

**To Reproduce**
Here is a minimal example to reproduce the problem:
```
import matplotlib.pyplot as plt
import numpy as np
import yaml

from floris.tools import FlorisInterface


# Instantiate FLORIS with the GCH model
with open('inputs/gch.yaml', 'r') as file:
    input_dict = yaml.safe_load(file)
    solver_dict = input_dict["solver"]

# Set up a 4-turbine wind farm with some partial overlap
D = 126.
spacing = 3.5
layout_x = np.array([0, D*1*spacing, D*2*spacing, D*3*spacing])
layout_y = [0, 0, 0.5 * D, -0.5 * D]

grid_points_array = range(2, 40)
farm_power_array = np.zeros(len(grid_points_array))
for ii, no_rotor_points in enumerate(grid_points_array):
    print("Evaluating FLORIS with {} grid points.".format(no_rotor_points))
    
    # Update number of rotor points
    solver_dict["turbine_grid_points"] = no_rotor_points

    fi = FlorisInterface(input_dict)
    fi.reinitialize(layout = [layout_x, layout_y], wind_speeds=[8.0])

    # Calculate wind farm power
    fi.calculate_wake()
    farm_power_array[ii] = fi.get_farm_power()[0][0]

# Plot
fig, ax = plt.subplots()
ax.plot(grid_points_array, farm_power_array, '-o')
ax.plot([0, grid_points_array[-1]], 2 * [farm_power_array[-1]], '--')
ax.plot([0, grid_points_array[1]], 2 * [farm_power_array[1]], '--')
ax.grid(True)
ax.set_xlabel('Number of rotor points (-)')
ax.set_ylabel('Farm power (W)')
plt.show()
```

With a spacing of 3.5 D:
![image](https://user-images.githubusercontent.com/22119448/190341714-a12c42af-3c8d-40f1-921c-57a3be9ac23a.png)
Where the error between the farm power with `turbine_grid_points=3` is about 6% off the converged value.

And convergence with a spacing of 5.0 D:
[...]

PR #649, opened 2023-05-07
Add cubature integration method for rotor velocities

rafmudaf:
# Add a cubature integration grid to average rotor points

This pull request adds a new grid type for placing and weighting grid points at the rotor so that a more accurate average can be calculated with fewer grid points. As @Bartdoekemeijer noted in #499, the square grid of points on the rotor swept area is inadequate since the calculated performance is highly dependent on grid resolution. Furthermore, the outer region of the blades are not captured at all and this is the area most sensitive to power production or power loss due to wakes.

The plot below shows a comparison of the convergence between the new and old method for averaging the velocity on the rotor for a turbine under no wake and with a shear profile exponent of 0.12. The x-axis for both series are the grid points input, and this translates to the square root of the number of points actually created on the grid. Here, you can see the square grid has a slow convergence rate, while the cubature grid is nearly converged by N=3.
 
convergence velcoity

This slow convergence in the current method is due to coarse sampling of both the shear profile and the wind turbine wake profiles. As shown in the image below, the typical 3 x 3 rectangular grid on the rotor swept area only covers a portion of the discs, and the new cubature grid covers a larger portion of the swept area.

grids

The following plot again shows the convergence of the cubature method for three turbines in line so two downstream turbines are waked.

all turbines

## Usage

The cubature integration method works seamlessly with the existing solvers. To enable it, simply use these solver settings:

```python
    solver_dict = {
        "type": "turbine_cubature_grid",
        "turbine_grid_points": 4
    }
```

## API Changes
Initially, this pull request changed the API's for the turbine-module functions for calculating performance (`Turbine.rotor_effective_velocity`, `Turbine.Ct`, `Turbine.axial_induction`). However, after reviewing the current implementation and results, we've decided to approach this change as an intermediate step for including this type of grid and integration method. The current API's are retained so that no breaking changes are introduced when using the existing square grid (`TurbineGrid`). To use the new grid (`TurbineCubatureGrid`), additional input arguments are passed to the following functions:

```python
def rotor_effective_velocity(
    ...
    average_method: str = "cubic-mean",
    cubature_weights: np.array | None = None
) -> NDArrayFloat:

def Ct(
    ...
    average_method: str = "cubic-mean",
    cubature_weights: np.array | None = None
) -> NDArrayFloat:

def axial_induction(
    ...
    average_method: str = "cubic-mean",
    cubature_weights: np.array | None = None
) -> NDArrayFloat:

def average_velocity(
    ...
    method: str = "cubic-mean",
    cubature_weights: np.array | None = None
) -> NDArrayFloat:
```

After further validation, it is intended that this grid-type will be the default, and the API's will be changed accordingly.

## Future Work

The changes here have raised some important questions about the current velocity averaging in FLORIS. Specifically, we currently average velocities with a cube root of the mean of the velocities cubed. This accurately captures the average velocity for the power calculation since that is a function of the velocity cubed, but other turbine properties have a different relationship to the velocity yet we use the same method.

Additionally, it may be worthwhile to consider how the velocities are weighted across the blade span since the root does less work than the tip, but the velocities at both stations are weighted equally.

Finally, more work is required to show that the cubature method arrives at the *correct answer* in addition to arriving at the answer faster. The existing model defaults may need to be retuned based on the performance calculated from a different velocity integration, as well. 

## Related issue
#499

A new grid-type is added to the grid module, the turbine module performance functions have new optional arguments, and the solvers are updated to reflect these API changes.


floris/simulation/grid.py
```diff
@@ -79,6 +79,7 @@ class Grid(ABC):
     x_sorted: NDArrayFloat = field(init=False)
     y_sorted: NDArrayFloat = field(init=False)
     z_sorted: NDArrayFloat = field(init=False)
+    cubature_weights: NDArrayFloat = field(init=False, default=None)
```
misi9170, line 82: I can see why this is needed here; I wonder whether in future, we should have an abstract TurbineGrid class (inherits from abstract Grid, or possibly general PointsGrid as we have discussed elsewhere) from which TurbineCubatureGrid and TurbineSquareGrid would inherit. Could get a little complicated with multiple levels on inheritance though.

floris/simulation/turbine.py
```diff
@@ -429,9 +443,28 @@ def axial_induction(
...
+    weights = cubature_weights.flatten()
+    weights = weights * len(weights) / np.sum(weights)
+    return np.cbrt(np.mean((array**3.0 * weights[None, None, None, :, None]), axis=axis))
+
 def average_velocity(
     velocities: NDArrayFloat,
-    ix_filter: NDArrayFilter | Iterable[int] | None = None
+    ix_filter: NDArrayFilter | Iterable[int] | None = None,
+    method: str = "cubic-mean",
+    cubature_weights: np.array | None = None
```
misi9170, line 467: `NDArrayFloat` instead of `np.array` for consistency?

misi9170 (commented):
This looks good to me @rafmudaf. I'll look now into why the empirical_gauss examples aren't running and push changes up.

paulf81 (approved):
Reviewed the code, and I know a few more changes are coming but wanted to log approval that this is looking really great

floris/simulation/turbine.py
misi9170: @rafmudaf I've gone ahead and made this change, please revert if this was incorrect.

floris/simulation/grid.py
rafmudaf: I agree. The grid module overall needs some work. It would be helpful to identify where there's overlap between these grid-types. Possibly one path forward is to get rid of one grid super class in favor of categories of grids. At the least, the grid abstract base class could only define the required API's for all grids like `x`, `x_sorted` and `sorted_indeces` etc.

floris/simulation/turbine.py
rafmudaf: I used `np.array` since this is specifically a 1D array and `NDArrayFloat` is n-dimensional array. We don't have a correct type for this instance, but I like the form that you or Paul used in [FlorisInterface.same_flow_at_points](https://github.com/NREL/floris/blob/develop/floris/tools/floris_interface.py#L942) for `1DArrayFloat`.

Lets leave it as you've changed it here, and consider `1DArrayFloat` and `5DArrayFloat` an option for future refinement.

misi9170 (approved):

rafmudaf:
Thanks for catching that @Bartdoekemeijer. I've made a note to clarify this in a couple of weeks, so I'll reach out then.

PR #641, opened 2023-04-27
Add arbitrary flow field point probing support

paulf81:
Not yet ready to merge

# Add ability to get velocity at arbitrary points

This branch, based on the branch on @rafmudaf adds the ability to collect the velocity at arbitrary points in space, based on the the new PointsGrid class.  The feature is meant to be similar in some ways to visualization methods, but avoids duplicate calls to underlying turbine grid solvers such that adding a few test points should not substantially slow down computation as adding a whole plane of points would.  

Further, the method seems to enable simplification of the code in that other visualization methods could be converted to be effectively wrappers to this call passed in differently oriented planes of points, or even volumes.

Currently also includes 3-4 additional examples but this number may change as we get nearer a final version.  

Still to go:

- [ ] Resolve issue that turbines apparently move when wind direction changes
- [ ] Add examples to directly test that wake is where we'd think by using farm layouts more complicated than lines
- [ ] Confirm works with at least most models (GCH/empirical/Jensen)
- [ ] Convert planar visualizations to be wrappers?
- [ ] Clean up things

PR #650, opened 2023-05-08
Add support for probing velocities at arbitrary flow field points

misi9170:
Ready for review

This pull request, based on the branch on @rafmudaf, adds the ability to collect the velocity at arbitrary points in space with the the new PointsGrid class. The feature is meant to be similar in some ways to visualization methods. This is similar to the draft pull request #641, but does not include the changes to `full_flow_sequential_solver` and instead wraps around the existing `full_flow_sequential_solver`. The downside of this is that `sequential_solver` is called multiple times for no real benefit other than compatibility with existing code. Redesigning the code to handle both visualization and a point probe without repeating solves is left as a future task.

To do:
- [X] Check passes all tests
- [x] Decide on placement/usage of examples


misi9170:
Ready for review.

floris/simulation/floris.py
```diff
@@ -253,6 +254,42 @@ def solve_for_viz(self):
...
+    def solve_for_points(self, points_x, points_y, points_z):
+        # Do the calculation with the TurbineGrid for a single wind speed
+        # and wind direction and 1 point on the grid. Then, use the result
+        # to construct the full flow field grid.
+        # This function call should be for a single wind direction and wind speed
+        # since the memory consumption is very large.
+
+        vel_model = self.wake.model_strings["velocity_model"]
+
+        if vel_model=="cc":
```
rafmudaf, line 266: Is it correct to have these if-statements? I'm not sure its relevant for this function.

floris/tools/floris_interface.py
```diff
@@ -943,6 +943,37 @@ def get_farm_AEP_wind_rose_class(
...
+        points_x: NDArrayFloat,
+        points_y: NDArrayFloat,
+        points_z: NDArrayFloat,
+    ):
+        """
+        To be written
+        Args:
+            probe_points (list[tuple], optional): List of probe points to add.
+        Returns:
+            Nothing?
```
rafmudaf, line 957: Please update docstring

examples/02a_visualizations_small.py
rafmudaf, line 1: Is this one meant to be here?

examples/_extract_wind_speed_at_points.py
```diff
@@ -0,0 +1,89 @@
...
+import floris.tools.visualization as wakeviz
+from floris.tools import FlorisInterface
+from floris.tools.visualization import (
+    calculate_horizontal_plane_with_turbines,
+    visualize_cut_plane,
+)
+
+
+"""
+Test the use of add points
```
rafmudaf, line 28: It would be helpful to describe what this example is intending to demonstrate. For example, it's noted below that this simulates a met mast between turbines. That would be good to explain at the top maybe with a little diagram of the layout of the turbines vs met mast or at least mention it. Initially, I didn't know what I was looking at in the plot and then went in to read the code.

examples/_extract_wind_speed_at_points.py
```diff
@@ -0,0 +1,89 @@
...
+fig, ax = plt.subplots(1,2)
+fig.set_size_inches(10,4)
+ax[0].scatter(fi.layout_x, fi.layout_y, color="black", label="turbine")
+
+# Set the wind direction to run 360 degrees
+wd_array = np.arange(0, 360, 1)
+fi.reinitialize(wind_directions=wd_array)
+
+# Grab the power before we change anything
+fi.calculate_wake()
```
rafmudaf, line 50: `fi.sample_flow_at_points` runs the wake calculation. Is this line needed?

examples/_temp_test_added_points.py
```diff
@@ -0,0 +1,106 @@
...
+
+# Get the same points using the turbine-based method
+cut_plane_from_turbine_sweep = calculate_horizontal_plane_with_turbines(fi,
+                                         x_resolution=100,
+                                         y_resolution=1,
+                                         x_bounds=[-D, 30*D],
+                                         y_bounds=[0, 0],)
+
+# Get the same points using horizontal plane
+horizontal_plane = fi.calculate_horizontal_plane(
```
rafmudaf, line 76: This is a really clever use of `calculate_horizontal_plane`. I guess is not a plane but a line. Anyway, good call.

examples/02a_visualizations_small.py
misi9170: Oops, no, good catch!

examples/_extract_wind_speed_at_points.py
misi9170: `fi.sample_flow_at_points()` was set up to raise an error if `calculate_wake()` had not been run; however, that was in the previous iteration that assumed that the flow would need to be solved before `full_flow_sequential_solver()` was called. I've now commented out the block that raises the error and can now remove this `calculate_wake()`.

examples/_temp_test_added_points.py
misi9170: That comes from @paulf81!

floris/simulation/floris.py
misi9170: You're right; fixing this now.

examples/_extract_wind_speed_at_points.py
rafmudaf: Excellent, thanks for addressing this.

floris/simulation/floris.py
```diff
@@ -253,6 +254,42 @@ def solve_for_viz(self):
         else:
             full_flow_sequential_solver(self.farm, self.flow_field, self.grid, self.wake)
 
+    def solve_for_points(self, points_x, points_y, points_z):
+        # Do the calculation with the TurbineGrid for a single wind speed
+        # and wind direction and 1 point on the grid. Then, use the result
```
rafmudaf, line 259: It's actually currently using a 3x3 turbine grid

floris/simulation/grid.py
```diff
@@ -473,3 +486,38 @@ def set_grid(self) -> None:
...
+    Args:
+        grid_resolution (`Vec3`): The number of grid points to be created in each direction.
+        turbine_coordinates (`list[Vec3]`): The collection of turbine coordinate (`Vec3`) objects.
+        reference_turbine_diameter (:py:obj:`float`): The reference turbine's rotor diameter.
+        grid_resolution (:py:obj:`int`): The number of points on each turbine
+    """
+    points_x: NDArrayFloat = field()
+    points_y: NDArrayFloat = field()
+    points_z: NDArrayFloat = field()
+    x_center_of_rotation: float | None = field(default=None)
```
rafmudaf, line 501: @misi9170 what would it mean if these center of rotation inputs are None? Is that valid?

misi9170: @rafmudaf it is valid, yes---they're only used in `rotate_coordiantes_rel_west`, which handles the case where they are None. However, there is a slight naming inconsistency here: in the `Grid` parent class, these fields are called `xc_rot` and `yc_rot`, so the names here should probably be updated to match.

PR #651, opened 2023-05-09
Disable user-provided tilt angle settings in wake calculation

paulf81:
Comment out tilt angles from calc wake function

floris_interface's calculate_wake function currently accepts tilt_angles as an input, but does not apparently apply them.  Propose to comment these out (and not delete in the assumption we will want to offer this function some day)

floris_interface.py

### DIFF
--- a/floris/simulation/farm.py
+++ b/floris/simulation/farm.py
@@ -192,6 +192,11 @@ class Farm(BaseClass):
             sorted_indices[:, :, :, 0, 0],
             axis=2,
         )
+        self.tilt_angles_sorted = np.take_along_axis(
+            self.tilt_angles,
+            sorted_indices[:, :, :, 0, 0],
+            axis=2,
+        )
         self.state = State.INITIALIZED
 
     def construct_hub_heights(self):
--- a/floris/tools/floris_interface.py
+++ b/floris/tools/floris_interface.py
@@ -116,7 +116,7 @@ class FlorisInterface(LoggerBase):
     def calculate_wake(
         self,
         yaw_angles: NDArrayFloat | list[float] | None = None,
-        tilt_angles: NDArrayFloat | list[float] | None = None,
+        # tilt_angles: NDArrayFloat | list[float] | None = None,
         # points: NDArrayFloat | list[float] | None = None,
         # track_n_upstream_wakes: bool = False,
     ) -> None:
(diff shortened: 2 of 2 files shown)


bayc:
> @paulf81 I see that they're used throughout the farm module: https://github.com/paulf81/floris/blob/feature/remove_tilt_from_calc_wake/floris/simulation/farm.py. Are they set in another way if it's removed from this function?

Hey @rafmudaf, I believe @paulf81 is right that the input right now doesn't assign the tilt angles in such a way that they get used. I believe I confirmed this with the test code below. I was able to track down that the tilt angles were not being initialized in the same was as the yaw angles when they are supplied to `calculate_wake` which I updated in the latest commit. The new output is shown below under Output 2.

Code:
```
import numpy as np

print(fi.floris.farm.tilt_angles)
print(fi.floris.farm.tilt_angles_sorted)

fi.calculate_wake()

print(fi.floris.farm.tilt_angles)
print(fi.floris.farm.tilt_angles_sorted)

fi.calculate_wake(tilt_angles=np.array([0.0, 0.0, 0.0]))

print(fi.floris.farm.tilt_angles)
print(fi.floris.farm.tilt_angles_sorted)
```
Output:
```
[[[5. 5. 5.]]]
[[[5. 5. 5.]]]
[[[5. 5. 5.]]]
[[[5. 5. 5.]]]
[[[5. 5. 5.]]]
[[[5. 5. 5.]]]
```

Output 2:
```
[[[5. 5. 5.]]]
[[[5. 5. 5.]]]
[[[5. 5. 5.]]]
[[[5. 5. 5.]]]
[[[0. 0. 0.]]]
[[[0. 0. 0.]]]
```

misi9170:
@bayc This does seem to work, but I don't fully understand the behavior when the turbine is defined with `floating_correct_cp_ct_for_tilt: True` and when the definition yaml has a `floating_tilt_table` field.

With your changes, running
```
import numpy as np

#fi = FlorisInterface("inputs_floating/gch_fixed.yaml")
#fi = FlorisInterface("inputs_floating/gch_floating_defined_floating.yaml")
fi = FlorisInterface("inputs_floating/gch_floating.yaml")

fi.reinitialize(layout_x = [0, 5*126, 10*126], layout_y = [0, 0, 0])

# print(fi.floris.farm.tilt_angles)
# print(fi.floris.farm.tilt_angles_sorted)

fi.calculate_wake(tilt_angles=np.array([[[5.0, 5.0, 5.0]]]))

# print(fi.floris.farm.tilt_angles)
# print(fi.floris.farm.tilt_angles_sorted)

print(fi.get_turbine_powers()/1000)

fi.calculate_wake(tilt_angles=np.array([[[0.0, 0.0, 0.0]]]))

# print(fi.floris.farm.tilt_angles)
# print(fi.floris.farm.tilt_angles_sorted)

print(fi.get_turbine_powers()/1000)
```

I get identical power at the first turbine (and very similar power at the second turbine). Is that expected?

Moreover, if we are going to let `calculate_wake()` accept a `tilt_angles` argument, I think that other methods on `FlorisInterface` should accept them too (e.g. `calculate_no_wake()`, `calculate_horizontal_plane()`, `calculate_cross_plane()`, etc), which is probably a bigger task to deal with. 

It seems to me the most straightforward path for v3.4 is to remove `tilt_angles` as an input from `calculate_wake()` as @paulf81 proposed, and create an issue to build out the code for allowing users to manually specify the tilt angle.

misi9170 (approved):

bayc (approved):

PR #663, opened 2023-05-26
Bug fix: erroneous rotation in turbine locations in horizontal flowfield plots

Bartdoekemeijer:
# Bug fix: Turbine locations in horizontal flow field plot

In the release version of FLORIS `v3.4`, the turbine locations are always rotated with respect to the wind direction in functions `visualization.plot_turbines` and `visualization.plot_turbines_with_fi`. However, with the recent change in horizontal flow field plots #578 where the turbine locations remain fixed and instead the flow field rotates, this puts the turbines in the wrong locations in the plot.

## Related issue

N/A

Visualization.

Here is a minimal script to reproduce the issue:
```

from floris.tools.visualization import plot_turbines_with_fi

# Load FLORIS
fi.reinitialize(wind_directions=[245.0])
horizontal_plane = fi.calculate_horizontal_plane(height=90.0)

# Create the plots
fig, ax = plt.subplots()
wakeviz.visualize_cut_plane(horizontal_plane, ax=ax, title="Horizontal")
plot_turbines_with_fi(fi=fi, ax=ax)
plt.show()
```

With the current `develop` branch and FLORIS `v3.4`, we find:
image

With this Pull Request, we find:
image

### DIFF
--- a/floris/tools/visualization.py
+++ b/floris/tools/visualization.py
@@ -76,14 +76,16 @@ def plot_turbines(
 
 def plot_turbines_with_fi(
     fi: FlorisInterface,
-    ax=None,
-    color=None,
-    wd=None,
-    yaw_angles=None,
+    ax: plt.Axes = None,
+    color: str = None,
+    wd: np.ndarray = None,
+    yaw_angles: np.ndarray = None,
 ):
     """
-    Wrapper function to plot turbines which extracts the data
-    from a FLORIS interface object
+    Plot the wind plant layout from turbine locations gotten from a FlorisInterface object.
+    Note that this function automatically uses the first wind direction and first wind speed.
+    Generally, it is most explicit to create a new FlorisInterface with only the single
+    wind condition that should be plotted.
 
     Args:
         fi (:py:class:`floris.tools.floris_interface.FlorisInterface`): FlorisInterface object.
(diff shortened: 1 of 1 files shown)


rafmudaf (commented):
Currently, we allow to pass yaw angles to `plot_turbines_with_fi` in order to rotate the lines plotted representing the turbines. However, `plot_turbines` does not take yaw angles so these two now have different functionality despite their intent to be equivalent. Also, note that both yaw angles and the wind direction are needed to rotate the yaw angles correctly. Somehow this difference between the two should be resolved, though I'm not sure what is best aside from finally dealing with how FLORIS handles yaw angles.

We could consider dropping `plot_turbines` in favor of the function requiring a `FlorisInterface` object for simplicity - it's the one I always use anyway.

Bartdoekemeijer:
That sounds good to me! We can maybe add a disclaimer when people run `plot_turbines` for now warning that this function will be deprecated in v3.5? Just so that we don't do anything API breaking immediately.

paulf81:
I like this approach, @rafmudaf?

rafmudaf:
I agree. I'll make the change here.

rafmudaf:
I've made the change suggested above. The warning message looks like this:
```
(floris) >>mbp@~/Development/floris/examples (pr/Bartdoekemeijer/663 *$)$ python 01_opening_floris_computing_power.py 
/Users/rmudafor/Development/floris/examples/01_opening_floris_computing_power.py:15: DeprecationWarning: The `plot_turbines` method is deprecated and will be removed in v3.5, use `plot_turbines_with_fi` instead.
  plot_turbines()
```

Note I used the stdlib warnings module instead of the floris logging module because this is a specific Python idiom. See [here](https://docs.python.org/3/library/warnings.html#warnings.warn) for more info.

rafmudaf:
I've made note of this API change in https://github.com/NREL/floris/discussions/488.

And with that, I'll go ahead and merge. Thanks for the fix @Bartdoekemeijer.

Issue #664, opened 2023-05-31
Bug: Axial induction calculation called from the floris interface crashes


MarcusBecker-GitHub:
# Axial induction calculation called from the floris interface crashes

The function `get_turbine_ais()` in the floris interface crashes, since it is called with `ref_tilt_cp_cts` instead of `ref_tilt_cp_cts_sorted`. `ref_tilt_cp_cts` is a n-turbine array, `ref_tilt_cp_cts_sorted` is a 1 x 1 x n-turbine array.
As a result, line 432 in /simulation/turbine.py throws an error:
```python
    if ix_filter is not None:
        yaw_angle = yaw_angle[:, :, ix_filter]
        tilt_angle = tilt_angle[:, :, ix_filter]
        ref_tilt_cp_ct = ref_tilt_cp_ct[:, :, ix_filter] # <---- ERROR
```

As I see it, the bug could be fixed by either changing the function `get_turbine_ais()` in the floris interface:
```python
    def get_turbine_ais(self) -> NDArrayFloat:
        turbine_ais = axial_induction(
            velocities=self.floris.flow_field.u,
            yaw_angle=self.floris.farm.yaw_angles,
            tilt_angle=self.floris.farm.tilt_angles,
            ref_tilt_cp_ct=self.floris.farm.ref_tilt_cp_cts_sorted, # <---- CHANGE
            fCt=self.floris.farm.turbine_fCts,
            tilt_interp=self.floris.farm.turbine_fTilts,
            correct_cp_ct_for_tilt=self.floris.farm.correct_cp_ct_for_tilt,
            turbine_type_map=self.floris.farm.turbine_type_map,
            average_method=self.floris.grid.average_method,
            cubature_weights=self.floris.grid.cubature_weights,
        )
        return turbine_ais
```
or by modifying the way `ix_filter` is used in the `axial_induction()` function in turbine.py, but I don't know what would be the favoured approach. Additionally, it might be good to add a comment to the `axial_induction()` function in turbine.py to define the input structure, as it is done with the other inputs.

## How to reproduce
```python
from floris.tools import FlorisInterface
fi = FlorisInterface('*Path*/examples/inputs/gch.yaml')
fi.calculate_wake()
fi.get_turbine_ais()
```

## Floris version
FLORIS v3.4

rafmudaf:
[...]

PR #665, opened 2023-06-01
Bug fix: Finalize Farm attributes

rafmudaf:
# Finalize Farm attributes and other Farm clean up
This pull request finalized three attributes on the Farm class:

```python
Farm.ref_density_cp_cts
Farm.ref_tilt_cp_cts
Farm.correct_cp_ct_for_tilt
```

Currently, these are created when the Farm attributes are expanded using the `Farm.construct_*` methods. However, these particular attributes are initially 1D with the length being the number of turbines. After the wake calculation, they must be expanded again via `Farm.finalize()` so that they have the three dimensions expected by the post-processing routines in `Turbine`, that is `(n wind directions, n wind speeds, n turbines)`.

Thanks to @marcusbecker-github for reporting this bug in #664.

Additionally, this pull request declares some Farm-class attributes that were otherwise added to the class after init. The attributes are also grouped into whether they have a corresponding `_sorted` version.

## Related issue
#664

Only changes are in farm.py

Since this is a bug, it could be released as v3.4.1 along with #663.

### DIFF
--- a/floris/simulation/farm.py
+++ b/floris/simulation/farm.py
@@ -75,10 +75,13 @@ class Farm(BaseClass):
     hub_heights: NDArrayFloat = field(init=False)
     hub_heights_sorted: NDArrayFloat = field(init=False, default=[])
 
+    turbine_type_map: NDArrayObject = field(init=False, default=[])
     turbine_type_map_sorted: NDArrayObject = field(init=False, default=[])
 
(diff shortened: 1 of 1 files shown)


rafmudaf:
ping @bayc - I think this PR is good, but I'd like to get your eyes on it to double check

Issue #621, opened 2023-04-04
Feature request: Implement yaw optimization via geometric relationship


paulf81:
# Implement yaw optimization via geometric relationship
Incorporate the rapid yaw optimziation approach descibed in PJ Stanley's paper here:

https://wes.copernicus.org/preprints/wes-2023-1/

Believe @bayc  has approach ready to implement

@pjstanle FYI

misi9170:
I'm working on this and will have a pull request up in the next few days.

misi9170:
Closed by #681

PR #681, opened 2023-07-12
Add geometric yaw code for rough yaw optimization (in layout optimization)

misi9170:
This pull request incorporates code from @pjstanle (https://github.com/pjstanle/GeometricYaw) to provide a rough estimate of optimal wake steering angles based purely on wind farm layout and wind direction (i.e., avoiding any wake calculations). See also https://wes.copernicus.org/preprints/wes-2023-1/

This is then incorporated as an option in the layout optimization routines so that efficient coupled layout and yaw optimizations can be conducted. 

All source code changes are in the tools/ subdirectory. These changes touch the yaw optimization and layout optimization classes. 

A new example (16c) is added to demonstrate the use of the geometric yaw optimization routine within the layout optimization routine. 

Addresses #621 
Tests pass.

Ready for review.


examples/15_optimize_layout.py
rafmudaf, line 1: As a suggestion, there may be some value in enabling the geometric yaw feature here all the time and doing two optimizations to compare the results. That demonstrates to the reader that there are two methods and you can explain the difference. Additionally, that allows for running the geometric yaw code at each commit through GitHub Actions. While it doesn't check for correctness, it would at least verify that the code doesn't crash. In fact, I had to download this pull request and change the flag to check whether this code runs.

misi9170: @rafmudaf Thanks, I did consider that and you're right, I might as well make that change. The results may not change much between the two cases (with and without geometric yaw), but it's still worth it for the automatic run.

rafmudaf: Yea and the fact that users will be able to see and understand that the results aren't intended to change much will add some clarity, as well.

examples/16b_heterogeneity_multiple_ws_wd.py
rafmudaf, line 1: Please change the examples listing here: https://nrel.github.io/floris/examples.html#b-heterogenaity-multiple-ws-wd-py

examples/15_optimize_layout.py
misi9170: @rafmudaf I decided to revert this example (back to the current one in develop) and will add a new example instead.

misi9170:
@rafmudaf @paulf81 after some delay, this is ready for review. @pjstanle , it would also be great to get your feedback if you have any!

rafmudaf (approved):

paulf81 (approved):
I pulled this down and ran and reviewed the examples, all looks great!!  Thanks @misi9170

Issue #670, opened 2023-06-09
Feature request: Example for floating wind farm


bayc:
# Example of floating wind farm functionality

There has been a request for an example to demonstrate the floating effects that were recently included in FLORIS at the farm level.

## Proposed solution

A sizable floating array (20 turbines or so?) that is simulated and power performance and wake flows are shown for a fixed bottom version and for a floating version.

misi9170:
@bayc I can take a look in the next couple of weeks. In the meantime, example 25_tilt_driven_vertical_wake_deflection.py would be a place to start for those interested in how (albeit, somewhat exaggerated) tilt affects wakes and power production.

rafmudaf:
@misi9170 @bayc Just raising this issue again. Is the v3.5 milestone still accurate?

misi9170:
Now addressed in #695

misi9170:
Closed by #695

PR #695, opened 2023-08-03
Add example of larger floating farm with comparison to fixed-bottom.

misi9170:
Ready for review.

Address #670 (request for more realistic example of floating wind farm, including turbine powers and visualization).

Note: uses "default" EmGauss parameters, which are subject to change and should be tuned to match observations from a particular farm.

TODO:
- [X] Create example, including necessary input files
- [X] Check tests, ruff, isort pass
- [x] Update examples index in documentation

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -129,7 +129,7 @@ the two-turbine farm.
 
 ### 29_floating_vs_fixedbottom_farm.py
 
-Compares a fixed-bottom wind farm (with a gridded layout) to a floating 
+Compares a fixed-bottom wind farm (with a gridded layout) to a floating
 wind farm with the same layout. Includes:
 - Turbine-by-turbine power comparison for a single wind speed and direction
 - Flow visualizations for a single wind speed and direction
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -100,7 +100,7 @@ Demonstrates the definition of a floating turbine and how to enable the effects 
 on Cp and Ct. 
 
 For further examples on floating wind turbines, see also examples
-25 (vertical wake deflection by a forced tilt angle) and 29 (comparison between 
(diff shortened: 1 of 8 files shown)


rafmudaf (approved):

rafmudaf:
Maybe one thing to add is how examples 24 and 29 are different. As a reader, I can guess that 29 has a larger problem size, but otherwise why would I need to go through example 24 if I can study 29?

misi9170:
> Maybe one thing to add is how examples 24 and 29 are different. As a reader, I can guess that 29 has a larger problem size, but otherwise why would I need to go through example 24 if I can study 29?

Right, we now have three examples for floating, each looking at something slightly different:
- 24 looks at how tilt affects turbine power and thrust curves, but does not include any wake effects
- 25 looks at how tilt affects wake deflections, but does so by essentially hard-coding the tilt angle (so is useful for understanding, but is not very realistic)
- 29 is a more realistic situation where the tilt is a function of wind speed. Both Cp/Ct curve effects and wake effects are included. So, it's the best example of how we would handle a real floating farm.

rafmudaf:
> @rafmudaf I've now provided a small update to the Examples Index to reflect this.

Excellent, thanks!

docs/examples.md
```diff
@@ -97,7 +97,11 @@ of a turbine layout within FLORIS.
 
 ### 24_floating_turbine_models.py
 Demonstrates the definition of a floating turbine and how to enable the effects of tilt
-on Cp and Ct.
+on Cp and Ct.
```
rafmudaf, line 100: You've got a trailing space here

bayc (approved):
Looks great @misi9170 ! I know you are looking into the divide by zero warning on the EmG model, but otherwise good to go from my side!

PR #697, opened 2023-08-04
Feature/random opt

paulf81:
# Add randomized layout optimization

@misi9170  and I have been working on an optimization method which works a little like a particle filter.  We used some ideas for parrelelization from @Bartdoekemeijer 's UncertainityInterface so in much the same way it should work parallelized on a computer or hpc with mpi as @Bartdoekemeijer had set up the UncertainityInterface (nice code @Bartdoekemeijer!)

The main idea of the algorithm is to maintain N number of open possible optimal layouts and randomly perturb turbines within the layout.  Periodically the candidate layouts are re-sampled such that the best performing layouts are duplicated and the worst are removed.  

Some nice features of the algorithm are since it's working are exposed we can order the operations of checks from cheapest (check in bounds, check min-distance) to more expensive (compute AEP) and only compute AEP if viable.  

We include the possibility for random jumps in turbine locations to avoid getting stuck in local minima.  We believe the algorithm will scale to more cores, and to disconnected layout regions pretty seamlessly but haven't tested yet.  It should intrinsically support heterogeneous inflows but haven't tested yet.  The algorithm also uses a time limit (rather than convergence test) to decide when to quit, this could be improved later on.

Sharing now, though still in progress to allow easier collaboration and feedback.  We include an example 29_test_random_opt.py to show usage


Bartdoekemeijer:
Sounds really interesting! The way you describe it, it seems to be some sort of genetic algorithm, where only the strongest contenders survive every iteration and turbine locations are randomly mutated. Looking forward to seeing how this all pans out compared to the existing methods!

misi9170:
@rafmudaf @paulf81 Given that this is not a bug fix or particular cleanup, I'm going to put this on the 4.1 milestone. There may be a bit of refactoring and merge conflict handling needed to align with 4.0, but I can deal with that when it happens.

misi9170:
Examples are currently failing due to something with Matplotlib that I don't understand... perhaps it will fix itself, otherwise we may need to put a version specifier in

paulf81:
> Examples are currently failing due to something with Matplotlib that I don't understand... perhaps it will fix itself, otherwise we may need to put a version specifier in

I'll have a look

paulf81:
> > Examples are currently failing due to something with Matplotlib that I don't understand... perhaps it will fix itself, otherwise we may need to put a version specifier in
> 
> I'll have a look

I think it fixed it, I think in earlier versions of matplotlib you could say `cm.get_cmap` or `plt.cm.get_cmap` but now `plt.get_cmap` is preferred.  I checked that the change works with matplotliblib 3.7.1 (what I had) and 3.9.  Seems to follow from: https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.9.0.html#removals

misi9170:
Thanks @paulf81 !

misi9170:
@Bartdoekemeijer , @bayc , any chance you'd be willing to review this PR? I'm not really an appropriate reviewer myself, given my contribution here

floris/optimization/layout_optimization/layout_optimization_random_search.py
```diff
@@ -0,0 +1,707 @@
...
+        self.y = self.y_initial # Required by _get_geoyaw_angles
+        self.objective_initial = _get_objective(
+            self.x_initial,
+            self.y_initial,
+            self.fmodel,
+            self._get_geoyaw_angles(),
+            self.use_value,
+        )
+        #del self.x # Deleting to avoid confusion, since not updated
+        #del self.y # Deleting to avoid confusion, since not updated
```
misi9170, line 294: Remove these?

misi9170 (changes requested):

PR #706, opened 2023-09-01
[Bugfix]: Update geometric yaw optimizer layout when called during layout optimization

misi9170:
This pull request fixes a bug that I missed when I created #621. When the geometric yaw code is called from within a layout optimization, the layout should be updated with each call, but this was not happening---instead, the yaw angles for the initial layout were being given. 

This fixes the bug. One artifact of the fix is that any child optimization class of `LayoutOptimizationBase` that calls `_get_geoyaw_angles()` with `self.enable_geometric_yaw = True` is _required_ to have attributes `self.x` and `self.y` (representing the $x$ and $y$ locations of the turbines). This is already the case for `LayoutOptimizationScipy` and `LayoutOptimizationPyOptSparse`.

The difference (although slight) can be observed by running example 16c. 

Ready for review and merge.


paulf81 (approved):
Tanks @misi9170, this looks good to me.  I pulled it down and ran 16c successfully.

rafmudaf (approved):

Issue #713, opened 2023-09-19
Feature request: Ambient TI as a function of wind direction and wind speed


Bartdoekemeijer:
# Feature request: turbulence_intensity as a function of wind direction and wind speed
The industry typically evaluates the annual energy production of a site by using a wind rose of the site, specifying the frequency of each combination of wind direction and wind speed. Additionally, we are often provided one ambient turbulence intensity value for each of those wind direction/speed bins. Alternatively, we may be given a wind-speed vs. turbulence intensity curve, showing that turbulence is high at low wind speeds, and much lower at high wind speeds. FLORIS only allows us to insert one turbulence value for the entire wind rose. With a relatively small change, I think we could update this and provide this very useful functionality.

## Proposed solution
Allow users to insert two-dimensional table for `turbulence_intensity`, which matches the dimensions `(n_wind_directions, n_wind_speeds)`.

## Alternatives considered
One can also simply apply a for-loop over each wind direction and wind speed and apply the relevant TI for each bin, but that completely beats the purpose of the refactorizations of FLORIS v3.

## Additional context
I have written up a preliminary solution that I'll post over in the PR section.

maapasan:
Hello @Bartdoekemeijer 

I agree with you, turbulence intensity is a significant characteristic and it is related with the wake simulation (basis of FLORIS).
I sent an issue that I think that is totally related with your suggestion. https://github.com/NREL/floris/issues/666

Kind regards from CENER :)

Bartdoekemeijer:
Ah of course, @maapasan! Thanks for highlighting that! I'm becoming a bit lazy and not searching the existing Issues like I should have! Will leave my Issue here but will link yours too in my PR.

paulf81:
Plan to address this is in #775

paulf81:
now connecting to Issue #778

paulf81:
@Bartdoekemeijer , if you have a chance, have a look at #782 which I think addresses this issue

paulf81:
@Bartdoekemeijer #782 is now in review if you want to have a look to make sure this does what you wanted

paulf81:
This is included in v4 via #782

Issue #666, opened 2023-06-02
Support for turbulence intensity field


maapasan:
# Manage turbulence intensity arrays in FlorisInterface.reinitialize

While wind speed or wind direction are defined using arrays (which allow the simultaneous simulation of several wind inputs), the turbulence intensity is defined as float.

In FLORIS, turbulence intensity is established for all combinations of direction and speed.

However, the turbulence intensity values are dependent on wind speed (turbulence intensity decrease with wind speed in real operating conditions). 

## Proposed solution
Update implementation in order to be able to run the simulations with array input turbulence intensity 
These changes will include the definition of turbulence_intensity in FlowField, as it is defined as float and it is needed NDArrayFloat, similar to wind_speeds and wind_directions.

## Additional context
Current use of the function, where wd and ws are np.array with same shape
`fi.reinitialize(wind_directions=wd, wind_speeds=ws, time_series=True)`
Suggested use (wd, ws and ti are np.array with same shape
`fi.reinitialize(wind_directions=wd, wind_speeds=ws, turbulence_intensity= ti, time_series=True)`

Thank you very much!!!

paulf81:
A good suggestion @maapasan , I'm adding some tracking info and we'll put it on the agenda for v3.5

paulf81:
@bayc , is this something you've looked at before?

paulf81:
I hope to address this request in #775

paulf81:
Now connecting to Issue #778

PR #714, opened 2023-09-19
Feature: Include WD and WS dependency in ambient turbulence intensity

Bartdoekemeijer:
This PR is not yet ready to be merged.

# Feature: turn turbulence_intensity into a table of which its values vary with wind direction and wind speed.
In this PR, I change `turbulence_intensity` into a 2D array with shape `(n_wind_directions, n_wind_speeds)`, while guaranteeing backward compatibility. It gives much greater freedom and better accuracy when modeling sites, at practically no downsides, I think.

## Related issue
See Issue #666 and #713.

`FlowField`, `FlorisInterface`

This nicely fits in how information is provided in the industry in terms of turbulence intensity dependency. We either receive turbulence intensity as a single value, or as a function of wind speed, or as a function of wind direction and wind speed (most commonly).

Using the `develop` branch and running the following code, I find:
```
import numpy as np

# Instantiate FLORIS using either the GCH or CC model
ws_array = np.array([6.0, 7.0, 8.0, 9.0])
ti_array = np.array([0.10, 0.09, 0.08, 0.07])
farm_powers = []
for wsii, ws in enumerate(ws_array):
    fi.reinitialize(
        wind_directions=[270.0],
        wind_speeds=[ws],
        turbulence_intensity=ti_array[wsii]
    )
    fi.calculate_wake()
    farm_powers.append(float(fi.get_farm_power()))
print(f"Farm powers (old method): {np.round(farm_powers)}")
```
The results is: `Farm powers (old method): [1214040. 1948831. 2863529. 3944825.]`

Then implementing it directly with this new PR, with the following code:
```
# Option two: as a table
fi.reinitialize(
    wind_directions=[270.0],
    wind_speeds=ws_array,
    turbulence_intensity=ti_array[None, :],
)
fi.calculate_wake()
farm_powers = fi.get_farm_power().flatten()
print(f"Farm powers (new method): {np.round(farm_powers)}")
```
We find the same values: `Farm powers (new method): [1214040. 1948831. 2863529. 3944825.]`. The big benefit is that we completely omit the need for users to implement for-loops or split/merge solutions in pre- and post-FLORIS runs.

It feels like I am getting pretty deep into the core code of FLORIS with these changes. @rafmudaf, @bayc, could you review and let me know if these changes align with your views?

### DIFF
--- a/examples/test_turbulence_intensity_table.py
+++ b/examples/test_turbulence_intensity_table.py
@@ -34,22 +34,26 @@ The power of both turbines for each wind direction is then plotted
 # Instantiate FLORIS using either the GCH or CC model
 fi = FlorisInterface("inputs/gch.yaml") # GCH model matched to the default "legacy_gauss" of V2
 
-# # Option one: 
-# fi.reinitialize(
-#     wind_directions=[260.0, 270.0, 280.0],
-#     wind_speeds=[7.0, 9.0],
-#     turbulence_intensity=0.06
-# )
-# fi.calculate_wake()
-# farm_powers = fi.get_farm_power()
-# print(f"Farm powers: {farm_powers}")
+# Option one:
+ws_array = np.array([6.0, 7.0, 8.0, 9.0])
+ti_array = np.array([0.10, 0.09, 0.08, 0.07])
+farm_powers = []
+for wsii, ws in enumerate(ws_array):
+    fi.reinitialize(
+        wind_directions=[270.0],
+        wind_speeds=[ws],
+        turbulence_intensity=ti_array[wsii]
+    )
+    fi.calculate_wake()
+    farm_powers.append(float(fi.get_farm_power()))
+print(f"Farm powers (old method): {np.round(farm_powers)}")
 
 # Option two: as a table
 fi.reinitialize(
-    wind_directions=[260.0, 270.0, 280.0],
-    wind_speeds=[7.0, 9.0],
-    turbulence_intensity=[[0.08, 0.06], [0.081, 0.064], [0.079, 0.058]]
+    wind_directions=[270.0],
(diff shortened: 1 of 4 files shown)


misi9170:
Hi @Bartdoekemeijer , thank you for this pull request! I think we have now incorporated similar content into the upcoming v4, which defaults to "time series" mode and allows users to specify the TI condition for each index (along with the wind speed and dimension). See #782. There is also now an option to generate a `TimeSeries` wind data object to pass to `fi.reinitialize()` that can define TI as a function of wind direction and wind speed, either with a user-specified function (https://github.com/NREL/floris/blob/v4/floris/tools/wind_data.py#L336) or with the IEC specification (https://github.com/NREL/floris/blob/v4/floris/tools/wind_data.py#L348). Examples 35 and 36 on the v4 branch demonstrate this functionality (these examples are dynamic and will likely change before v4 is released).

Bartdoekemeijer:
> Hi @Bartdoekemeijer , thank you for this pull request! I think we have now incorporated similar content into the upcoming v4, which defaults to "time series" mode and allows users to specify the TI condition for each index (along with the wind speed and dimension). See #782. There is also now an option to generate a `TimeSeries` wind data object to pass to `fi.reinitialize()` that can define TI as a function of wind direction and wind speed, either with a user-specified function (https://github.com/NREL/floris/blob/v4/floris/tools/wind_data.py#L336) or with the IEC specification (https://github.com/NREL/floris/blob/v4/floris/tools/wind_data.py#L348). Examples 35 and 36 on the v4 branch demonstrate this functionality (these examples are dynamic and will likely change before v4 is released).

Sounds good -- even better! Good to close this PR in that case.

PR #719, opened 2023-10-20
Allow switching on and off contour plot numerical labels at top level

misi9170:
## Description

The flow visualization tools were hardcoded to plot numerical labels on the flow contour plots. This PR makes that an optional choice at the top level in the `visualize_cut_plane()` and `visualize_heterogeneous_cut_plane()` functions of tools/visualization.py, which are passed through to the mid-level function `line_contour_cut_plane()`.

I have set the default value of the new `label_contours` Boolean to False, as I believe that most users will not want the contour labels shown. However, I've now updated examples 02_visualizations.py and 16_heterogeneous_inflow.py so that `label_contours` is set to True for those examples. I have left the default value (False) in other examples involving flow plots.

## Other notes
- No changes are made to source code
- Output from most examples involving flow contour plots has changed, as the contour labels are now largely gone. The exceptions are 02 and 16, where I have reincluded the contour labels.
- The function `visualize_heterogeneous_cut_plane()` of tools/visualization.py does not appear to be called anywhere in the floris code or examples, and may be redundant. However, I have not confirmed this and have left the function in place.
- Tests pass.

## Examples
Running 02_visualizations.py with `label_contours = True` produces 
image

while setting `label_contours = False` produces
image

### DIFF
--- a/examples/02_visualizations.py
+++ b/examples/02_visualizations.py
@@ -18,10 +18,6 @@ import numpy as np
 
 import floris.tools.visualization as wakeviz
 from floris.tools import FlorisInterface
-from floris.tools.visualization import (
-    calculate_horizontal_plane_with_turbines,
-    visualize_cut_plane,
-)
 
 
 """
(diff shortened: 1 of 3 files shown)


rafmudaf (approved):
Nicely done!

PR #723, opened 2023-10-24
Update default parameters for the Empirical Gaussian model

misi9170:
## Description

When the Empirical Gaussian model was implemented, not a lot of attention was paid to the default parameter settings because it was intended to be tuned to match the user's SCADA records. While that intention is still the case, we have now had a chance to provide a rough tuning to various SCADA records and can provide an updated default set of parameters. 

Note that we may update the defaults again in future to provide different parameter sets for, for example, onshore and offshore farms, daytime and nighttime boundary layers, etc.

## Affected code
- examples
  - Main Empirical Gaussian example input file inputs/emgauss.yaml 
  - Secondary Empirical Gaussian examples inputs for floating under inputs_floating/
  - Examples 26, 27, and 29 (minor updates to improve understanding with new defaults)
- floris/simulation
  - attrs default values in wake_deflection/empirical_gauss.py and wake_velocity/empirical_gauss.py
 
## Functional changes
- No changes to main source code
- Examples 25, 26, 27, and 29 now produce slightly different results
- Tests pass.


misi9170:
Converting to draft until @paulf81 has had a chance to look at this

rafmudaf:
I expected that the regression tests would fail after changing the default model parameters. Do you have a thought why they didn't @misi9170?

misi9170:
> I expected that the regression tests would fail after changing the default model parameters. Do you have a thought why they didn't @misi9170?

@rafmudaf Thank you for pointing this out. As we discussed, this was caused by my overlooking the parameter values in conftest.py. I have now updated those (which initially causes the regression test to fail), and then updated the comparison outputs in the regression tests so that they're now passing once again.

rafmudaf (approved):

PR #748, opened 2023-11-30
Use compatible-release specifier for dependency versions

### LINKED ISSUE #745: Change dependent versioning to semantic versioning using the compatible release specifier
Improve dependencies to use compatible release specifier such that the dependencies won't go up a major version unless intentionally changed in an update to this repo.  This is to avoid breaks due to API changes.  Info on compatible release specifier:

https://packaging.python.org/en/latest/specifications/version-specifiers/#id4

paulf81:
# Use compatible release specifiers for dependency versions

This pull request adds version specifications for the dependencies with the [compatible release specifier](https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release). The change uses the latest release of each dependent version of the requirement without incrementing a major version which in [semantic versioning](https://semver.org) could include breaking changes.

From the docs:
> For a given release identifier V.N, the compatible release clause is approximately equivalent to the pair of comparison clauses:
    >= V.N, == V.*

Some examples from the docs:

```
~= 2.2                     # <- this line is approximately equivalent
>= 2.2, == 2.*         # <- to this line

~= 1.4.5                   # <- this line is approximately equivalent
>= 1.4.5, == 1.4.*.    # <- to this line
```

More concretely, if we say `numpy~=1.20`, it means that FLORIS requires Numpy to be at least v1.20 and any other larger v1 is valid but not v2.

## Related issue
Closes #745 

setup.py and the dependencies chosen during installation

@kflemin

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,20 +29,20 @@ REQUIRES_PYTHON = ">=3.8.0"
 # What packages are required for this module to be executed?
 REQUIRED = [
     # simulation
-    "attrs",
-    "pyyaml",
-    "numexpr",
-    "numpy>=1.20",
-    "scipy>=1.1",
+    "attrs~=23.1",
+    "pyyaml~=6.0",
+    "numexpr~=2.8",
+    "numpy~=1.20",
+    "scipy~=1.1",
 
     # tools
-    "matplotlib>=3",
-    "pandas",
-    "shapely",
+    "matplotlib~=3.8",
+    "pandas~=2.1",
+    "shapely~=2.0",
 
     # utilities
-    "coloredlogs>=10.0",
-    "flatten_dict",
+    "coloredlogs~=10.0",
+    "flatten_dict~0.4",
 ]
 
 # What packages are optional?
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ REQUIRED = [
 
     # utilities
     "coloredlogs~=10.0",
-    "flatten_dict~0.4",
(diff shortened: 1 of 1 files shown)


paulf81:
ok @rafmudaf all tests passing now and I got rid of the overly picky versions, good to go?

rafmudaf:
Can you show that you tested the "docs" extra dependencies? These are not covered by the tests unless the documentation is built, and in this case it isn't

setup.py
```diff
@@ -30,19 +30,19 @@
 REQUIRED = [
     # simulation
     "attrs",
-    "pyyaml",
-    "numexpr",
-    "numpy>=1.20",
-    "scipy>=1.1",
+    "pyyaml~=6.0",
+    "numexpr~=2.0",
+    "numpy~=1.20",
```
rafmudaf, line 35: This line pins the Numpy version to 1.20, so I think the Numpy version should stay as `>=1.20`. FLORIS requires this minimum version, but it supports greater than this:
```
(floris) >>$ pip freeze | grep numpy
numpy==1.24.2
```

To keep the bounds, I think this will work: `numpy>=1.20,<2`. Since there is no Numpy v2, I wonder if this is valid, but worth trying.

rafmudaf: Ah shoot I keep misreading the `~=` - disregard this comment

paulf81:
And all tests pass

rafmudaf:
Thanks for checking that. Well done fixing this!

PR #755, opened 2023-12-06
Fix the turbine.power function unit test

paulf81:
# Fix the power unit test

The unit test of the power() function in turbine.py, test_power() in turbine_unit_test.py contained several errors which were masked by the fact that the tolerance value passed assert_allclose() function was very large.  This pull request fixes these issues.  Specifically:

  - Previously, the cp_interp function was passed to power, yielding a cp value multiplied by air_density.  However, this is not expected.  get_turbine_powers() in floris_interface.py passes self.floris.farm.turbine_power_interps, which are in W.  Correcting this yields a p, and a power_truth both in W
  - The "truth" calculation of p was missing the air density term
  - The new test compares the successfully with a very small, default tolerance.
  - Added to the unit test are checks that in high wind speeds the power is 5MW (since the test case is the reference 5MW turbine) and at near 0 wind speeds, power should be close to 0
  - Both the test_function, and the doc-string for power implied that rotor_effective_velocities should be 5D (including the last 2 dimensions for grid).  However, since these are rotor-averaged quantities they should be only 3D, and this pull request fixes that in the test, and in the doc-string of power() (it was not necessary to change the code)
 - Finally, a block of commented out code in turbine_unit_test.py was deleted

turbine.py
turbine_unit_test.py

### DIFF
--- a/tests/turbine_unit_test.py
+++ b/tests/turbine_unit_test.py
@@ -304,31 +304,26 @@ def test_ct():
 
 
 def test_power():
-
-    # Test that power is computed as expected for a single turbine
-    N_TURBINES = 1
     AIR_DENSITY = 1.225
 
+    # Test that power is computed as expected for a single turbine
+    n_turbines = 1
+    wind_speed = 10.0
     turbine_data = SampleInputs().turbine
     turbine = Turbine.from_dict(turbine_data)
-    turbine_type_map = np.array(N_TURBINES * [turbine.turbine_type])
+    turbine_type_map = np.array(n_turbines * [turbine.turbine_type])
     turbine_type_map = turbine_type_map[None, None, :]
-
-    # Use wind speed = 10.
-    wind_speed = 10.0
-
-    # Compute power using the power function
-    p = power(
+    test_power = power(
         ref_density_cp_ct=AIR_DENSITY,
         rotor_effective_velocities=wind_speed * np.ones((1, 1, 1)),
         power_interp={turbine.turbine_type: turbine.power_interp},
         turbine_type_map=turbine_type_map[:,:,0]
     )
 
-    # Recompute using the original provided cp table
+    # Recompute using the provided Cp table
     truth_index = turbine_data["power_thrust_table"]["wind_speed"].index(wind_speed)
     cp_truth = turbine_data["power_thrust_table"]["power"][truth_index]
-    power_truth = (
+    baseline_power = (
         0.5
         * AIR_DENSITY
         * turbine.rotor_area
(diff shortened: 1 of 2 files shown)


tests/turbine_unit_test.py
```diff
@@ -304,62 +304,66 @@ def test_ct():
...
+    wind_speed = 0.1
+
+    p = power(
+        ref_density_cp_ct=AIR_DENSITY,
+        rotor_effective_velocities=wind_speed * np.ones((1, 1, 1)),
+        power_interp={turbine.turbine_type: turbine.power_interp},
+        turbine_type_map=turbine_type_map[:,:,0]
+    )
+
+    np.testing.assert_allclose(p,0.0,10.0 )
```
rafmudaf, line 364: Since this is defined as 0.0 in the power curve for this turbine, the tolerance on this test should be lower.

tests/turbine_unit_test.py
```diff
@@ -304,62 +304,66 @@ def test_ct():
 
 
 def test_power():
-    N_TURBINES = 4
```
rafmudaf, line 307: Using multiple turbines tests an aspect of the vectorization. Could you maintain a test with multiple turbines?

rafmudaf (commented):
Well done @paulf81! Here are a few comments to expand the tests and add a little more clarity. Overall, this is great, thanks for fixing this.

tests/turbine_unit_test.py
```diff
@@ -304,62 +304,66 @@ def test_ct():
...
-    #         * cp_truth
-    #         * turbine.generator_efficiency
-    #         * effective_velocity_trurth ** 3
-    #     )
-    #     print(i,WIND_SPEEDS, effective_velocity_trurth, cp_truth, p[0, 0, i], power_truth)
-    #     np.testing.assert_allclose(p[0, 0, i], power_truth)
+    np.testing.assert_allclose(p,power_truth )
+
+
+    # Test that above rated wind speeds, the power should be very close to 5MW
```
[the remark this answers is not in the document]
paulf81: Done!

paulf81:
Improvements made in commit: "Add github revisions":

1. Include multiple turbine test
2. Include test with grid dimension
3. Include comment that test turbine is NREL 5MW
4. Lower tolerance for 0 power test to 1 W

rafmudaf (approved):

rafmudaf:
I changed the array comparisons here to `assert np.allclose(..., ...)` instead of `np.testing.assert_allclose(..., ...)` because the defaults for relative and absolute tolerance are different:
- [np.allclose](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html): rtol=1e-05, atol=1e-08
- [np.testing.assert_allclose](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html): rtol=1e-07, atol=0

This means `np.allclose` tests for 5 significant digits until numbers very close to zero and then it checks for the absolute tolerance. On the other hand, `np.testing.assert_allclose` always tests for 7 significant digits. In this particular case, it's the rtol that fails in the power calculation, so switching comparison methods also relaxes the relative tolerance.

PR #759, opened 2023-12-07
Fix issue with mutable default attributes

### LINKED ISSUE #758: Mutable datatypes initialized with unsafe defaults
# Mutable datatypes initialized with unsafe defaults

Some of the attrs-derived classes in `floris.simulation` may be initialized with unsafe defaults for mutable data types ([mutable vs immutable types in Python](https://realpython.com/pointers-in-python/#immutable-vs-mutable-objects)). In Python, `list`, `set`, and `dict` types are all mutable, and variables referring to mutable types act like pointers to the same memory. For mutable-type attributes in attrs-derived classes with defaults of empty instances of the mutable type, any instance of the class will refer to the same memory for these particular attributes.

See https://github.com/NREL/floris/pull/750#discussion_r1412596120 for the initial report.

## Example
```python
from attrs import define, field

@define
class A:
    x: list = field(default=[])
    y: list = field(factory=list)

a1 = A()
a2 = A()

print(a1, a2)
a1.x.append(4)
a1.y.append(5)
print(a1, a2)
```

The code above yields the following output:
```
A(x=[], y=[]) A(x=[], y=[])
A(x=[4], y=[5]) A(x=[4], y=[])
```

The first print-statement shows that the instances of class `A` are initialized with empty lists for both `x` and `y` attributes. In the second print-statement, appending a value to both attributes for the `a1` instance is shown to also affect the `x` attribute of the `a2` instance. The `y` attribute is not affected because it is declared with `factory=list` so a new instance of `list` is created for new instances of `A`.

This is also true of NumPy arrays, as shown below. NumPy arrays don't have an `append` method, but they are also mutable and their values can be changed directly leading to this same side effect.

```python
from attrs import define, field
import numpy as np

@define
class A:
    x: np.array = field(default=np.array([1,2,3]))

a1 = A()
a2 = A()

print(a1, a2)
a1.x[0] = 9
print(a1, a2)
```
```
A(x=array([1, 2, 3])) A(x=array([1, 2, 3]))
A(x=array([9, 2, 3])) A(x=array([9, 2, 3]))
```

## Impacts on FLORIS
The attrs library is used only in `floris.simulation`, so this issue is isolated to that package. While many class definitions have this issue, the use of `append` is generally discouraged for performance considerations. However, there are many instances where the values in NumPy arrays are modified directly including all of the solvers.

The ultimate impact is unclear, but I'll update this section as this issue is understood more.

### Affected classes
Here's a first-pass list at affected classes.
- Nearly all attributes in [Farm](https://github.com/NREL/floris/blob/main/floris/simulation/farm.py#L44)
- Nearly all attributes in [FlowField](https://github.com/NREL/floris/blob/main/floris/simulation/flow_field.py#L36)
- Empirical Gauss [velocity model defaults](https://github.com/NREL/floris/blob/main/floris/simulation/wake_velocity/empirical_gauss.py#L68)

paulf81:
# Fix issue with mutable default attributes

This pull request address the fact that if you use a list, dict, array or other mutable object as a default parameter in attrs (or more generally in function defaults), all instances of the declared object will point to the same list/dict/array, possibly generating confusing bugs.  See Issue #758 for more detail on the problem and proposed solution.

The approach used in this pull request was determined in Issue #758 and is also identified as the correct solution in the attrs documentation (https://www.attrs.org/en/stable/api-attr.html#attr.ib) (can see also a quick explanation here: https://refex.readthedocs.io/en/latest/guide/fixers/attrib_default.html), specifically to use a factory to generate a new list/dict/array, rather declaring a single instance.  The resulting changes all take the form of converting:

```
power: NDArrayFloat = field(default=[], converter=floris_array_converter)
```

to:

```
power: NDArrayFloat = field(factory=list, converter=floris_array_converter)
```
For empty lists, while using a lambda for empty numpy arrays:

```
w_initial_sorted: NDArrayFloat = field(init=False, factory=lambda: np.array([]))
```

## Related issue
Closes #758

farm.py
turbine.py
flow_field.py
wake.py
empirical_gauss.py

### DIFF
--- a/floris/simulation/farm.py
+++ b/floris/simulation/farm.py
@@ -81,10 +81,10 @@ class Farm(BaseClass):
 
     turbine_definitions: list = field(init=False, validator=iter_validator(list, dict))
 
-    turbine_fCts: Dict[str, interp1d] | List[interp1d] = field(init=False, default=[])
-    turbine_fCts_sorted: NDArrayFloat = field(init=False, default=[])
+    turbine_fCts: Dict[str, interp1d] | List[interp1d] = field(init=False, factory=list)
+    turbine_fCts_sorted: NDArrayFloat = field(init=False, factory=list)
 
-    turbine_fTilts: list = field(init=False, default=[])
+    turbine_fTilts: list = field(init=False, factory=list)
 
     yaw_angles: NDArrayFloat = field(init=False)
     yaw_angles_sorted: NDArrayFloat = field(init=False)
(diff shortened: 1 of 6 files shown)


rafmudaf (approved):

floris/simulation/farm.py
```diff
@@ -81,10 +81,10 @@ class Farm(BaseClass):
 
     turbine_definitions: list = field(init=False, validator=iter_validator(list, dict))
 
-    turbine_fCts: Dict[str, interp1d] | List[interp1d] = field(init=False, default=[])
-    turbine_fCts_sorted: NDArrayFloat = field(init=False, default=[])
+    turbine_fCts: Dict[str, interp1d] | List[interp1d] = field(init=False, factory=list)
+    turbine_fCts_sorted: NDArrayFloat = field(init=False, factory=list)
```
RHammond2, line 85: For instances like this below, I would recommend modifying slightly to ensure the typing is valid (even if unchecked currently) to use the following paragdigm:

```python
x: NDArrayFloat = field(init=False, factory=list, converter=floris_array_converter)
```

Essentially what will happen if typing is implemented, without converting to the array type, it will fail. Additionally it creates some confusion (though some already exists tbc) to create a list for an object indicated to be an array.

The only place this will cause issues is with non-float arrays, in which case I'd recommend having two variations of `floris_array_converter` such as `float_array_converter` and `object_array_converter`. I don't know that it needs to have `floris_` at the front at this point, and might be a lingering (but dated) idea from early in the design process.

Because this is a suggested course correction somewhat late in the game, i could see this being a separate issue/PR since that usage is pretty common.

floris/simulation/flow_field.py
```diff
@@ -48,21 +48,25 @@ class FlowField(BaseClass):
...
-    u_initial_sorted: NDArrayFloat = field(init=False, default=np.array([]))
-    v_initial_sorted: NDArrayFloat = field(init=False, default=np.array([]))
-    w_initial_sorted: NDArrayFloat = field(init=False, default=np.array([]))
-    u_sorted: NDArrayFloat = field(init=False, default=np.array([]))
-    v_sorted: NDArrayFloat = field(init=False, default=np.array([]))
-    w_sorted: NDArrayFloat = field(init=False, default=np.array([]))
-    u: NDArrayFloat = field(init=False, default=np.array([]))
-    v: NDArrayFloat = field(init=False, default=np.array([]))
-    w: NDArrayFloat = field(init=False, default=np.array([]))
+    u_initial_sorted: NDArrayFloat = field(init=False, factory=lambda: np.array([]))
```
RHammond2, line 51: My comment from `farm.py` stands for this as well since that would have the same effect without a lambda for an empty object, but is also not a required change.

floris/simulation/turbine.py
```diff
@@ -472,9 +472,9 @@ class PowerThrustTable(FromDictMixin):
         ValueError: Raised if the power, thrust, and wind_speed are not all 1-d array-like shapes.
         ValueError: Raised if power, thrust, and wind_speed don't have the same number of values.
     """
-    power: NDArrayFloat = field(default=[], converter=floris_array_converter)
-    thrust: NDArrayFloat = field(default=[], converter=floris_array_converter)
-    wind_speed: NDArrayFloat = field(default=[], converter=floris_array_converter)
+    power: NDArrayFloat = field(factory=list, converter=floris_array_converter)
```
RHammond2, line 475: I guess I'd say we should have a unified approach now that I see it's a separate approach for the same results in each of the files so far (love Python for this!), at least to make it easier for future developers.

RHammond2 (changes requested):
I thinks is great! My only requested change is to pick one approach for the empty object creation process, like I mentioned in my comment on `turbine.py`. I'm not really picky (even if I've demonstrated a preference in the commecnts) about which it ends up being, so long as it's consistent (which is really focused on future maintenance).

paulf81:
Hi @RHammond2 , thank you for your review!  I opened a new issue (#761) to capture your request, would you mind to both check if that issue captures the problem/solution well?  And then also would you mind to re-review this pull request and let us know if we can merge?  Thank you!

paulf81:
Thank you @RHammond2 !

PR #763, opened 2023-12-13
Enhancement: Add a Dimension Validator for 5-D and 3-D Array Structures

### LINKED ISSUE #761: Use a consistent method for empty object creation process
# Use one process for empty object creation process

In the course of resolving the issue with mutable types as default parameters in attrs classes within FLORIS in pull request #759, @RHammond2 points out that we need to pick a consistent method for empty object creation.  Currently, we are at risk that if in the future we implement type checking, we could fail tests.

@RHammond2 says:

For instances like this below, I would recommend modifying slightly to ensure the typing is valid (even if unchecked currently) to use the following paragdigm:

```python
x: NDArrayFloat = field(init=False, factory=list, converter=floris_array_converter)
```

Essentially what will happen if typing is implemented, without converting to the array type, it will fail. Additionally it creates some confusion (though some already exists tbc) to create a list for an object indicated to be an array.

The only place this will cause issues is with non-float arrays, in which case I'd recommend having two variations of `floris_array_converter` such as `float_array_converter` and `object_array_converter`. I don't know that it needs to have `floris_` at the front at this point, and might be a lingering (but dated) idea from early in the design process.

RHammond2:
# Validate the Dimensions of 5-D and 3-D Arrays

This PR adds new validation functionality for arrays that should adhere to dimensions requirements of N wind directions x N wind speeds x N turbines or N wind directions x N wind speeds x N turbines x N grid x N grid. An additional mixin class that wraps `attrs.validate(class_instance)` has been added to easily enable attrs class validators to be run at key points in a simulation.

## Related issue

At least partially resolves #761 by using the `factory=lambda: np.array([])` paradigm to not build extensive carveouts for data that should only be valid at initialization.

### `floris.type_dec.py`
- `validate_5DArray_shape` is an attrs validator checking the adherence to the shape being N wind directions x N wind speeds x N turbines x N grid x N grid, with the exception of empty arrays and those that can be broadcast along the grid axes.
- `validate_3DArray_shape` is an attrs validator checking the adherence to the shape being N wind directions x N wind speeds x N turbines, with the exception of empty arrays and those that can be broadcast along the wind speed axis.
- `validate_mixed_dim` runs the 5D, then 3D validation to check for the `turbulence_intensity` attributes that start as 5D, then are reduced to 3D.
- `array_3D_field` is a  custom field for defining a field with the 3D validator and lambda factory built in to reduce the number of multi-line attribute initializations.
- `array_5D_field` is a  custom field for defining a field with the 5D validator and lambda factory built into reduce the number of multi-line attribute initializations.
- `array_mixed_dim_field` is a  custom field for defining a field with the 3D/5D validator and lambda factory built into reduce the number of multi-line attribute initializations.
- `ValidateMixin` provides `self.validate()`, which is a wrapper for `attrs.validate(class_instance)` to easily run all validators in a class.

### `floris/simulation/base.py`
- `BaseClass`: now inherits `ValidateMixin`

### `floris/simulation/farm.py` and `floris/simulation/flow_field.py`
- Adopts the `array_3D_field`, `array_5D_field`, and `array_mixed_dim_field` as appropriate

### `floris/simulation/grid.py`
- Fixes some typing errors.
- Adopts the `array_3D_field` and `array_5D_field` where appropriate

### `floris/tests/floris_unit_test.py`
- Fixes a bug in creating a simulation object with incorrect wind speed and wind direction dimensions

### `floris/tests/type_dec_unit_test.py`
- Adds in a test for `ValidateMixin.validate()` and the new dimensions validators.

This largely stems from a conversation with @rafmudaf about questions of how to check if attributes are staying the same shape throughout a simulation, and adding methods to ensure the correctness of dimension sizes.

Tests are passing with a slight modification to an existing test that was using incorrect dimensions.

### DIFF
--- a/floris/simulation/base.py
+++ b/floris/simulation/base.py
@@ -33,7 +33,7 @@ from attrs import (
 )
 
 from floris.logging_manager import LoggingManager
-from floris.type_dec import FromDictMixin
+from floris.type_dec import FromDictMixin, ValidateMixin
 
 
 class State(Enum):
@@ -43,7 +43,7 @@ class State(Enum):
 
 
 @define
-class BaseClass(FromDictMixin):
+class BaseClass(FromDictMixin, ValidateMixin):
     """
     BaseClass object class. This class does the logging and MixIn class inheritance.
     """
--- a/floris/simulation/farm.py
+++ b/floris/simulation/farm.py
@@ -33,6 +33,7 @@ from floris.simulation import (
 )
 from floris.simulation.turbine import compute_tilt_angles_for_floating_turbines
 from floris.type_dec import (
+    array_3D_field,
     convert_to_path,
     floris_array_converter,
     iter_validator,
@@ -79,6 +80,9 @@ class Farm(BaseClass):
         default=default_turbine_library_path, converter=convert_to_path
     )
 
+    n_wind_directions: int = field(init=False)
+    n_wind_speeds: int = field(init=False)
+
     turbine_definitions: list = field(init=False, validator=iter_validator(list, dict))
 
     turbine_fCts: Dict[str, interp1d] | List[interp1d] = field(init=False, factory=list)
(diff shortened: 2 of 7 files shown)


paulf81 (approved):
hi @RHammond2, I reviewed this code and all looks good to me!  I went through the changes and re-ran the tests locally.  Do I understand right the plan is to confirm this all works in 5D, then merge into 4D work and update there?

RHammond2:
@paulf81, that's the plan for now, or if the 4D work happens first, this branch will get updated prior to merging. The idea was to keep the work separated and avoiding dependent PRs on either end.

RHammond2:
@rafmudaf the examples all work on my machine now, so this should be ready for review.

misi9170:
@RHammond2 @paulf81 @rafmudaf Although this PR is stale, I still think something like your validator functions could be useful for a later version of FLORIS so I'm going to leave it open with an on-hold flag, as we already changed the dimensions of data structures for v4 and [may change them again](https://github.com/NREL/floris/discussions/1111#discussion-8314180) in a future release.

PR #767, opened 2023-12-18
Update reference turbine model data

misi9170:
In converting from Cp to power curves as the standard for v4, we also noted that the data for some of the models in turbine library (most notably, the NREL 5MW) are out of date.

This pull request would update that data. 

- turbine_library/ (turbine models updated; v3 models archived in legacy_v3 subdirectory; x_20MW turbine removed from library for v4)
- examples/inputs_floating (turbine models updated)
- tests/ (regression tests updated; minor changes to some turbine unit tests to reflect changes in curves and elimination of x_20MW turbine) **SEE NOTE BELOW**

## For discussion
A couple of important issues arise in this pull request, which may mean that we do not want to merge this as is. I will mark them complete once if/when we resolve these issues.
- [ ] Regression tests for small_grid_rotations are failing, because the second-through-last turbines in the top column no longer produce the same power as the second turbine. It is not clear that this _should_ be the case anyway. @rafmudaf and I discussed this to some extent---this is not a deal-breaker, but should be updated if we'd like to proceed with merging.
- [ ] Updates to the reference turbine data, and in particular, the NREL 5MW, will change outputs from examples. I will attempt to document the extent of these changes. This _shouldn't_ affect the default wake model tunings, but as outputs will change, the two effects (turbine model parameters and wake model parameters) could be conflated. I will update this shortly with a graphic to show the changes to the NREL 5MW data curves.


misi9170:
Closing as superseded by #771

PR #770, opened 2023-12-28
Update to more flexible turbine operation model definitions

misi9170:
## Description 
This pull request makes significant changes to the `Turbine()` class and surrounding code (most heavily located in turbine.py, but spread throughout the FLORIS source code) to enable specification of more advanced functionality for computing the thrust coefficient and power of a turbine off of "nominal" operation. In doing so, I've added two turbine operation models, `CosineLossTurbine` (which is the default, and matches the previous behavior of FLORIS v3) and `SimpleTurbine`, which is an actuator disk without handling for rotor misalignment with the flow.

We envisage that other operation models will be added in FLORIS v4.1.

In doing so, I've also incorporated the multidimensional Cp/Ct turbine code. There are likely still some minor updates needed here to get all naming conventions aligned (e.g. dropping "cp" in favor of "power", see #765 ); however, the necessary functional changes have been made and multidimensional turbine tests pass.

______
## Previous discussion
_Modified from the [FLORIS v4 project board](https://github.com/orgs/NREL/projects/96)_
Currently, the turbine model is essentially implemented using interpolation functions, where each `turbine_type` has its own interpolation function and a mapping for which turbine is which type is created before the solver is run to mask the outputs of the interpolators to the correct turbine positions.

Proposed changes for FLORIS v4 would not change this underlying structure, but the interpolators would be replaced by more general callable function definitions to implement different ways of computing power and thrust coefficient given turbine inputs. _Now referred to as different "operation models"_

A ~turbine model~ _operation model_ is then a set of ~two~ _three_ functions: 
- `power()`, which returns the turbine power at the given atmospheric conditions and control settings
- `thrust_coefficient()` (or possibly `Ct()`), which returns the turbine thrust coefficient at the given atmospheric conditions and control settings.
- `axial_induction()`, returns the axial induction factor

For flexibility, there may also be a `prepare` function (similar to that of the [wake models](https://github.com/NREL/floris/blob/main/floris/simulation/wake_velocity/gauss.py#L41)) that establishes a dictionary of keyword arguments needed for each turbine type; this will be unpacked in the call to the turbine functions.

We imagine that `power()`, `thrust_coefficient()`, may call from a library of lower level functions that implement, for example, sliding along the power curve for yaw angle or air density changes.

We may represent a turbine model as a "static class", which has now memory but has various `@staticmethod`s for inheritance reasons and to enforce implementation of certain methods. 

The reason for not making them class objects is to save memory, but this could possibly be revisited if necessary as only one instantiated object would be needed for each `turbine_type` (not for each turbine).

It is not yet clear exactly how the selection of different ~subturbine~ _operation_ models will be configured in the inputs files. However, the ~subturbine~ _operation_ models could be specified within the turbine yamls, as an option.

How are we going to deal with multidimensional turbines? Should be able to be handled in this.

### Steps
- [x] Move to form where turbine.py is essentially a masker; `SimpleTurbine` implements main functionality, ` CosinePpTurbine` implements results of yaw misalignment. Check tests pass.
- [x] Add a base class for turbine operation models to inherit from, add tests to check inheritance. 
- [x] Update examples as needed.

### Completion criteria
- [x] Existing tests pass
- [x] All examples run
- [x] Add tests for each operation model

______

Still to do:
- [x] Multidimensional model
- ~axial induction models~
- [x] Base class to inherit from (?)
- [x] New tests
- [x] clean up (names, filenames, check for any remaining TODOs)
- ~Handling for rotor averaging method~
- [x] Docstrings update for simulation/turbine/operation_models.py
- [x] Get all examples running
- [x] Make sure all checks pass

## Architecture 
Architecture of new turbine module within floris.simulation (and main connections to other parts of code). Arrows indicate where a function, method, or class is used by something upstream of it (i.e., if `b` points to `a`, then `a` calls/uses `b`)
turbine_module_layout

## Further discussion needed (not included in this PR)
To discuss further:
- Best location for axial_induction model (on turbine.py (applies to all turbine models) vs on specified turbine submodels) _This has now been placed on the operation models_
- Current use of pT (seems very limited and somewhat inconsistent) _This was my misunderstanding---pT refers to the power loss exponent for tilt---I will create a separate PR clarifying this_
- Should the `"multidim_cp_ct"` velocity model be removed to allow the user to use any velocity deficit model? _Likely yes, but will be it's own PR_
- Should the multidimensional Power and Ct data structure be the default with the one-entry form being a special case? It would require a change to most users' input file, but it would simplify the code a bit.
- Should the averaging method be user-definable? Could live in the turbine yaml, or the main floris input yaml.
- Has this slowed down execution? Visualization runs seem slower, but I'm not sure about "normal" solves? _Approximately 10% slower on normal solves_
- I've been inconsistent about actually removing code vs just commenting things out, but once we are happy I can remove all commented out code. _These have now been removed_
- I've removed pP, pT, and ref_air_density from the farm object (mostly commenting out lines). Were they needed for anything, or am I breaking a convention here? _No_


rafmudaf:
> Modified from the FLORIS v4 project board (which is private)

Heads up that [it is not private](https://github.com/orgs/NREL/projects/96), so feel free to link to it and share as needed.

rafmudaf:
@misi9170 There's currently a bit of namespace overlap between the `Turbine` classes and the functions in the turbine submodules (power, Ct). I think I have a mental map, but it would be helpful to have a diagram of how all of these components relate to each other similar to the [floris.simulation diagram](https://nrel.github.io/floris/architecture.html#floris-simulation). Can you map the relationship between the turbine class and turbine submodels in a similar manner?

misi9170:
> @misi9170 There's currently a bit of namespace overlap between the `Turbine` classes and the functions in the turbine submodules (power, Ct). I think I have a mental map, but it would be helpful to have a diagram of how all of these components relate to each other similar to the [floris.simulation diagram](https://nrel.github.io/floris/architecture.html#floris-simulation). Can you map the relationship between the turbine class and turbine submodels in a similar manner?

misi9170:
> @misi9170 There's currently a bit of namespace overlap between the `Turbine` classes and the functions in the turbine submodules (power, Ct). I think I have a mental map, but it would be helpful to have a diagram of how all of these components relate to each other similar to the [floris.simulation diagram](https://nrel.github.io/floris/architecture.html#floris-simulation). Can you map the relationship between the turbine class and turbine submodels in a similar manner?

Thanks for the suggestion---I've now added a rough diagram to the PR description, which I could clean up for final documentation

floris/turbine_library/turbine_previewer.py
```diff
@@ -130,25 +130,25 @@ def  power_curve(
...
         if self.turbine.multi_dimensional_cp_ct:
-            power_interps = {
+            power_functions = {
                 k: multidim_power_down_select(
-                    np.full(shape, self.turbine.power_interp),
+                    np.full(shape, self.turbine.power_function),
                     dict(zip(self.turbine.condition_keys, k)),
                 )
-                for k in self.turbine.power_interp
+                for k in self.turbine.power_function
```
rafmudaf, line 138: Just a quick note before the weekend - I think this is an incorrect change. This iterates on `power_function` but it isn't utterable.

floris/simulation/farm.py
```diff
@@ -277,39 +258,27 @@ def construct_turbine_correct_cp_ct_for_tilt(self):
...
-                try:
-                    self.turbine_map.append(TurbineMultiDimensional.from_dict(_turb))
-                except FileNotFoundError:
-                    _turb["turbine_library_path"] = self.turbine_library_path
-                    self.turbine_map.append(TurbineMultiDimensional.from_dict(_turb))
-        else:
-            self.turbine_map = [Turbine.from_dict(turb) for turb in self.turbine_definitions]
+        self.turbine_map = [Turbine.from_dict(turb) for turb in self.turbine_definitions]
 
     def construct_turbine_fCts(self):
```
rafmudaf, line 263: Change function name to reflect that it isn't `fCt` anymore

floris/turbine_library/turbine_previewer.py
misi9170: Thanks for calling this out @rafmudaf. I didn't look at turbine_previewer.py---I was thinking I'd open a seperate bugfix for that at some point. This change probably got caught up in a usual variable name swap and I didn't notice.

floris/simulation/farm.py
```diff
@@ -81,8 +81,11 @@ class Farm(BaseClass):
 
     turbine_definitions: list = field(init=False, validator=iter_validator(list, dict))
 
-    turbine_fCts: Dict[str, interp1d] | List[interp1d] = field(init=False, factory=list)
-    turbine_fCts_sorted: NDArrayFloat = field(init=False, factory=list)
+    turbine_thrust_coefficient_functions: Dict[str, Callable] | List[Callable] = \
```
rafmudaf, line 84: These were previously either a dict or a list depending whether it was a multidimensional turbine. Now I believe they're only dict. Is that true?

floris/simulation/farm.py
```diff
@@ -81,8 +81,11 @@ class Farm(BaseClass):
...
+        field(init=False, factory=list)
+
+    turbine_axial_induction_functions: Dict[str, Callable] | List[Callable] = \
```
rafmudaf, line 87: These were previously either a dict or a list depending whether it was a multidimensional turbine. Now I believe they're only dict. Is that true?

floris/simulation/farm.py
```diff
@@ -95,29 +98,21 @@ class Farm(BaseClass):
...
 
-    turbine_map: List[Turbine | TurbineMultiDimensional] = field(init=False, factory=list)
+    turbine_map: List[Turbine] = field(init=False, factory=list)
 
     turbine_type_map: NDArrayObject = field(init=False, factory=list)
     turbine_type_map_sorted: NDArrayObject = field(init=False, factory=list)
 
-    turbine_power_interps: Dict[str, interp1d] | List[interp1d] = field(init=False, factory=list)
-    turbine_power_interps_sorted: NDArrayFloat = field(init=False, factory=list)
+    turbine_power_functions: Dict[str, Callable] | List[Callable] = field(init=False, factory=list)
```
rafmudaf, line 106: turbine_power_functions

floris/simulation/farm.py
rafmudaf: I believe I've fixed these in the last commit @misi9170 but I'll let you hit the resolve button

rafmudaf (approved):

misi9170:
We decided to use the name "operation model" to describe these models. Now, "turbine model" refers to the turbine physical properties, such as rotor diameter and hub height, which are normally specified in the turbine yamls. We chose the name "operation model" to highlight that these models describe how the turbine  performs in different _operating modes_.  For example, intentional yaw misalignment is an operating mode, and the manifestation of how yaw misaligment affects the turbine's thrust and power is specified in an operation model. Similarly, the Helix wake mixing method is a "mode" that the turbine may operate in, and we plan to implement an operation model to represent the effects on thrust, axial induction, and power.

PR #771, opened 2024-01-03
Update reference wind turbines in the default turbine library

misi9170:
## Description

In converting from Cp to power curves as the standard for v4, we also noted that the data for some of the models in turbine library (most notably, the NREL 5MW) are out of date. This pull request would update that data so that the turbine models now follow those specified [here](https://nrel.github.io/turbine-models/Offshore.html) and [here](https://github.com/NREL/turbine-models/tree/master/Offshore)*.

Legacy FLORIS v3 turbines are first archived in the subdirectory legacy_v3 for clarify, and then removed. Similarly, the convert_turbine_v3_to_v4 utility is used to generate turbines exactly matching the v3 turbines, which are archived in converted_from_v3 and then deleted. Finally, the (deleted) subdirectory updated_for_v4 contained turbines whose data matches the updated versions in turbine_library/.

### Affected areas
- floris/turbine_library
- tests/conftest.py and all regression tests
- examples/inputs_floating/turbine_files

Supersedes #767, which will be closed. Should be merged _after_ #770.

## Changes to the turbine power and thrust curves
Updating the data has slightly changed the power and thrust curves of the NREL 5MW, IEA 10MW, and IEA 15MW turbines (the x_20MW turbine has been archived and has not been brought forward at this stage for lack of a public data source that we could easily find).

The new and previous power curves are as follows:
compare_nrel_5MW_conversion

compare_iea_10MW_conversion

compare_iea_15MW_conversion

These can be generated by running the following script from within the turbine_library directory:
```
import numpy as np

from floris.utilities import load_yaml


l_color = "black" 
c_color = "gray"
u_color = "red"

l_ls = "solid"
c_ls = "dashed"
u_ls = "solid"

turb = "iea_15MW"

# Original v3 model
legacy = load_yaml("./legacy_v3/"+turb+"_v3legacy.yaml")
l_ws = legacy["power_thrust_table"]["wind_speed"]
l_Ct = legacy["power_thrust_table"]["thrust"]
l_P = (
    np.array(legacy["power_thrust_table"]["power"])
    * 0.5 * legacy["ref_density_cp_ct"] * legacy["rotor_diameter"]**2/4 * np.pi
    * np.array(l_ws)**3
) / 1e3

# Converted model
converted = load_yaml("./converted_from_v3/"+turb+"_v4converted.yaml")
c_ws = converted["power_thrust_table"]["wind_speed"]
c_P = converted["power_thrust_table"]["power"]
c_Ct = converted["power_thrust_table"]["thrust_coefficient"]

# Updated model
updated = load_yaml("./"+turb+".yaml")
u_ws = updated["power_thrust_table"]["wind_speed"]
u_P = updated["power_thrust_table"]["power"]
u_Ct = updated["power_thrust_table"]["thrust_coefficient"]

fig, ax = plt.subplots(2, 1, sharex=True)

ax[0].plot(l_ws, l_P, color=l_color, ls=l_ls, label="original")
ax[1].plot(l_ws, l_Ct, color=l_color, ls=l_ls, label="original")

ax[0].plot(c_ws, c_P, color=c_color, ls=c_ls, label="converted")
ax[1].plot(c_ws, c_Ct, color=c_color, ls=c_ls, label="converted")

ax[0].plot(u_ws, u_P, color=u_color, ls=u_ls, label="updated")
ax[1].plot(u_ws, u_Ct, color=u_color, ls=u_ls, label="updated")

ax[0].set_ylabel("Power [kW]")
ax[1].set_ylabel("Thrust coefficient [-]")
ax[1].set_xlabel("Wind speed [m/s]")
ax[0].grid()
ax[1].grid()
ax[1].set_xlim([c_ws[0], c_ws[-1]])

ax[0].legend(loc="upper right")

fig.savefig("compare_"+turb+"_conversion.jpg", dpi=300, format="jpg")

plt.show()
```

## Other notes
- The NREL 5MW thrust curve is now slightly higher than before, which is the main source of the change to regression tests
- The IEA 10MW specified [here](https://github.com/NREL/turbine-models/blob/master/Offshore/IEA_10MW_198_RWT.csv) reports generating 10.6MW above rated, but this is the mechanical power and a generator efficiency of 0.94% is used to bring this back to 10.0MW of electrical power.
- The IEA 15MW now generates very close to 15MW above rated, down from approximately 15.2MW in the legacy v3 model.
- The updated 5MW turbine generates very close to 5MW above rated, but the above-rated power is not perfectly flat. This could be updated to exactly 5000.0 kW, if desired. It is not expected to change results or even reg tests significantly.

\* The current definition of the NREL 5MW reference turbine specified a Ct value that is not the aerodynamic thrust coefficient. We're working with @pduff-code to update that.


misi9170:
@paulf81 , could you take a quick look at the power and thrust curves in the description above and see if there is anything there of concern?

floris/turbine_library/nrel_5MW.yaml
```diff
@@ -1,25 +1,24 @@
-
+# NREL 5MW reference wind turbine.
+# Data based on:
+# TODO: update URL
```
rafmudaf, line 3: Do we have the correct URL?

tests/data/nrel_5MW_v3legacy.yaml
rafmudaf, line 1: Is this still used?

tests/data/nrel_5MW_v4converted.yaml
rafmudaf, line 1: This can probably be called "nrel_5MW" or also symlink to the turbine library

tests/data/nrel_5MW_v3legacy.yaml
misi9170: This is used, along with nrel_5MW_v4converted, to test the utility that builds turbine models based on non-standard data (for example, user supplies Cp curve rather than absolute power curve) in turbine_utilities_unit_test.py. These had been simlinks to their equivalents in the turbine_library, but as those have now been deleted to leave only the updated turbines, I've established the files here so that the test continues to run.

tests/data/nrel_5MW_v4converted.yaml
misi9170: Unfortunately, that doesn't work---nrel_5MW_v4converted.yaml has data matching the v3legacy.yaml model, which is slightly different to the (new) nrel_5MW.yaml in turbine_library because we have updated the data slightly.

An option is to construct a v3-like turbine (but that does not exactly match what is in v3) to compare the nrel_5MW.yaml that is in turbine_library to.

floris/turbine_library/nrel_5MW.yaml
misi9170: The plan is for this to point here: https://github.com/NREL/turbine-models/blob/master/Offshore/NREL_5MW_126_RWT.csv but at the moment, the data hosted at that location is not intuitive (in particular, the thrust coefficient listed does not represent aerodynamic thrust). The plan is to update the data hosted at that location, and I think the URL will stay the same, but I'm hesitant to put it in when it points to unmatched data. 

How do you feel about leaving this as a TODO item and possibly creating an issue to remind us to add the URL once the data on https://github.com/NREL/turbine-models/ is updated?

rafmudaf: That sounds like a good plan to me

tests/data/nrel_5MW_v3legacy.yaml
rafmudaf: Ok got it. It's worth keeping in mind that there are quite a few NREL 5MW turbine definition files in the repository now, so keeping them in sync and the differences clear will become increasingly difficult. One strategy could be to develop tests that don't use a specific turbine model. In this particular case, the Cp to Power conversion and v3 to v4 conversion don't *need* the NREL 5MW turbine, they simply need a set of numbers that map to another set of numbers. I'm not suggesting a change here, but we should all keep it in mind.
```
(base) >>mbp@floris (v4-ms/turbine-library $)$ find . -name "*nrel*.y*ml"
./floris/turbine_library/nrel_5MW.yaml
./tests/data/nrel_5MW.yaml  (link to ./floris/turbine_library/nrel_5MW.yaml)
./tests/data/nrel_5MW_custom.yaml
./tests/data/nrel_5MW_v4converted.yaml
./tests/data/nrel_5MW_v3legacy.yaml
./docs/nrel_5MW.yaml  (link to ./floris/turbine_library/nrel_5MW.yaml)
./examples/inputs_floating/turbine_files/nrel_5MW_floating.yaml
./examples/inputs_floating/turbine_files/nrel_5MW_floating_fixedtilt15.yaml
./examples/inputs_floating/turbine_files/nrel_5MW_floating_defined_floating.yaml
./examples/inputs_floating/turbine_files/nrel_5MW_floating_fixedtilt5.yaml
./examples/inputs_floating/turbine_files/nrel_5MW_fixed.yaml
```

tests/data/nrel_5MW_v4converted.yaml
rafmudaf: I think this is a good idea. Otherwise, we'll have multiple NREL 5MW definitions within the same version of floris.

misi9170: I'll try to make that change today.

misi9170: @rafmudaf I've now updated that by removing nrel_5MW_v4converted.yaml and nrel_5MW_v3legacy.yaml from tests/data and adding data for a v3-type turbine that matches the updated nrel_5MW data to `SampleInputs` on conftest.py. Let me know if you think that's an improvement.

There is still a symlink to the NREL_5MW.yaml in tests/data, alongside a turbine model NREL_5MW_custom.yaml. I believe these are used in farm_unit_test.py, but are not touched by this PR aside from updating their data for consistency.

floris/turbine_library/x_20MW.yaml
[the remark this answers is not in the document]
misi9170: Are you referring to turbine_interaction.ipynb? Do you want to update that, or should I take a first pass?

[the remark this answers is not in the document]
rafmudaf: Yep that one, I've udpated it just now

floris/turbine_library/nrel_5MW.yaml
misi9170: The updated curves have now been provided in the turbine_models repo---thanks @pduff-code!! I've updated the URL.

rafmudaf (approved):

misi9170:
@paulf81 I've now created an issue (#774) to explain the redundancy of the `generator_efficiency` entry on the turbine yamls and to suggest removal. If you're otherwise happy with this PR, do you mind approving?

paulf81 (approved):
Thanks @misi9170 , I went through the code and ran the examples, all good on my end!

PR #775, opened 2024-01-12
Adds classes to structure wind energy data for FLORIS

paulf81:
@rafmudaf , @misi9170 , @ejsimley, editing this now that it is ready for review

# Wind Rose Refactor

Refactor wind rose class to work with new v4 framework.  Main points of inclusion in this pull request:

- Split wind_rose class into two classes, WindRose and TimeSeries, within one module to avoid circular imports
- Defines a super class WindData which both TimeSeries and WindRose inherit from
- Provide _unpack() functions that provide the information within either class to floris' reintialize function
- Provides testing for the new classes and functions in wind_rose_time_series.test.py
- Ruff formant floris_interface and add capability to pass WindData objects to reinitialize and get_aep functions
- Adds an example 34 showing new syntax of WindRose and TimeSeries

## Context

Originally this pull request was described as addressing more capabilities for the new WindRose TimeSeries but we decided to limit this to just the the above points and upcoming issues/pull requests will address turbulence_intensity based on wind rose


paulf81:
Thanks for the chat today @misi9170  and @rafmudaf , just logging this for when we're back next week.  I tried to implement the points we discussed and I added some capability to pass the objects directly to reinitialize (I think it works?  There's an example now testing it). One thing though, can reinitialize accept an array of turbulence intensity?  Otherwise have a good weekend!

misi9170:
> One thing though, can reinitialize accept an array of turbulence intensity? Otherwise have a good weekend!

At this stage, the `turbulence_intensity` keyword argument to `reinitialize()` is intended to be scalar, I believe. However, [this line](https://github.com/NREL/floris/blob/v4/floris/simulation/solver.py#L79) broadcasts that scalar value to have size `n_findex` x `n_turbines` x 1 x 1; so, it is seems that it could be straightforward to provide a different turbulence intensity for each findex, we'd just have to make sure to pass everything through correctly.

paulf81:
> > One thing though, can reinitialize accept an array of turbulence intensity? Otherwise have a good weekend!
> 
> At this stage, the `turbulence_intensity` keyword argument to `reinitialize()` is intended to be scalar, I believe. However, [this line](https://github.com/NREL/floris/blob/v4/floris/simulation/solver.py#L79) broadcasts that scalar value to have size `n_findex` x `n_turbines` x 1 x 1; so, it is seems that it could be straightforward to provide a different turbulence intensity for each findex, we'd just have to make sure to pass everything through correctly.

That sounds good, I think we should move towards a number of scalar inputs being allowed to be specified for each findex, but we can discuss more

paulf81:
@rafmudaf  @ejsimley  and @misi9170 , this pull request has been pared down as discussed, I've cleaned up and merged back in V4 and I think it's now ready for review

floris/tools/wind_rose_time_series.py
```diff
@@ -0,0 +1,529 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+from pandas.api.types import CategoricalDtype
+
+
+# Define the super lass that WindRose and TimeSeries inherit
+# Define functions here that are either the same for both WindRose and
+# TimeSeries or will be overloaded
+class WindData:
```
misi9170, line 25: Consider renaming to `WindDataBase` to clarify that we do not expect `WindData` to be instantiated directly.

misi9170 (commented):
I'll also plan to pull this down and run the example shortly.

floris/tools/wind_rose_time_series.py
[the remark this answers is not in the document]
paulf81: Good suggestion!  Taking it...

tests/wind_rose_time_series_test.py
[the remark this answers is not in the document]
paulf81: Done!

floris/tools/wind_rose_time_series.py
paulf81: Done

misi9170 (approved):
Thanks for the changes Paul. I think this looks good. I can imagine some enhancements, but as well be developing new features that use this functionality, I'm happy to get this in and start working with it and we can improve things as we go along.

examples/34_wind_data_temporary.py
rafmudaf, line 1: Somehow resolve this file's "temporary" status

floris/tools/floris_interface.py
```diff
@@ -32,6 +32,9 @@
 from floris.type_dec import NDArrayFloat
 
 
+# from floris.tools import WindData
```
rafmudaf, line 35: Is this needed?

floris/tools/floris_interface.py
```diff
@@ -200,6 +191,7 @@ def reinitialize(
         turbine_library_path: str | Path | None = None,
         solver_settings: dict | None = None,
         heterogenous_inflow_config=None,
+        wind_data=None,
```
rafmudaf, line 194: Is there an expected data type for "wind_data"?

floris/tools/floris_interface.py
```diff
@@ -401,10 +400,7 @@ def calculate_horizontal_plane(
 
         # Compute the cutplane
         horizontal_plane = CutPlane(
-            df,
-            self.floris.grid.grid_resolution[0],
-            self.floris.grid.grid_resolution[1],
-            "z"
+            df, self.floris.grid.grid_resolution[0], self.floris.grid.grid_resolution[1], "z"
```
rafmudaf, line 403: In the case of many arguments to a function, the preferred style is to list each on a new line. This enables easily eyeballing the inputs and swapping out one for another. If you add a trailing comma to the list of arguments, ruff will leave it as is.

floris/tools/floris_interface.py
```diff
@@ -796,8 +784,7 @@ def get_farm_AEP(
         # Check if frequency vector sums to 1.0. If not, raise a warning
         if np.abs(np.sum(freq) - 1.0) > 0.001:
             self.logger.warning(
-                "WARNING: The frequency array provided to get_farm_AEP() "
-                "does not sum to 1.0."
+                "WARNING: The frequency array provided to get_farm_AEP() " "does not sum to 1.0."
```
rafmudaf, line 787: If this string is one line, you don't need the double-double quotes.

floris/tools/floris_interface.py
```diff
@@ -838,6 +824,81 @@ def get_farm_AEP(
...
+        yaw_angles=None,
+        turbine_weights=None,
+        no_wake=False,
+    ) -> float:
+        """
+        Estimate annual energy production (AEP) for distributions of wind speed, wind
+        direction, frequency of occurrence, and yaw offset.
+
+        Args:
+            freq (NDArrayFloat): NumPy array with shape (n_findex)
```
rafmudaf, line 841: Double check that this docstring is correct and not a remnant of copy/paste. There's no `freq` in the arguments to this function.

floris/tools/floris_interface.py
```diff
@@ -838,6 +824,81 @@ def get_farm_AEP(
...
+                passed to reinitialize
+
+        Returns:
+            float:
+                The Annual Energy Production (AEP) for the wind farm in
+                watt-hours.
+        """
+
+        # Verify the wind_data object matches FLORIS' initialization
+        if wind_data is not None:
```
rafmudaf, line 886: What's the scenario where `wind_data` could be `None`?

rafmudaf: If it's possible to be None, that should be noted in the function signature.

floris/tools/wind_data.py
```diff
@@ -0,0 +1,529 @@
...
+# See https://floris.readthedocs.io for documentation
+
+import matplotlib.cm as cm
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+from pandas.api.types import CategoricalDtype
+
+
+# Define the super lass that WindRose and TimeSeries inherit
```
rafmudaf, line 22: Typo "lass"

rafmudaf: This should be the docstring for the class, not a comment above it. This way, Python includes it as information on the class definition itself.

floris/tools/wind_data.py
```diff
@@ -0,0 +1,529 @@
...
+# Define functions here that are either the same for both WindRose and
+# TimeSeries or will be overloaded
+class WindDataBase:
+    def __init__():
```
rafmudaf, line 26: Needed?

floris/tools/wind_data.py
```diff
@@ -0,0 +1,529 @@
...
+    def __init__(
+        self,
+        wind_directions,
+        wind_speeds,
+        freq_table=None,
+        ti_table=None,
+        price_table=None,
+        compute_zero_freq_occurence=False,
+    ):
+        if not isinstance(wind_directions, np.ndarray):
```
rafmudaf, line 86: Since you're enforcing the type, add type hints so users can easily know the function signature.

floris/tools/wind_data.py
```diff
@@ -0,0 +1,529 @@
...
+            raise TypeError("wind_directions must be a NumPy array")
+
+        if not isinstance(wind_speeds, np.ndarray):
+            raise TypeError("wind_directions must be a NumPy array")
```
rafmudaf, line 90: Copy / paste error

floris/tools/wind_data.py
```diff
@@ -0,0 +1,529 @@
...
+        # otherwise leave it None
+        if ti_table is not None:
+            if not ti_table.shape[0] == len(wind_directions):
+                raise ValueError("ti_table first dimension must equal len(wind_directions)")
+            if not ti_table.shape[1] == len(wind_speeds):
+                raise ValueError("ti_table second dimension must equal len(wind_speeds)")
+            self.ti_table = ti_table
+
+        else:
+            self.ti_table = None
```
rafmudaf, line 120: This is redundant

floris/tools/wind_data.py
```diff
@@ -0,0 +1,529 @@
...
+        # If price_table is not None, confirm it has correct dimension,
+        # otherwise initialze to all ones
+        if price_table is not None:
+            if not price_table.shape[0] == len(wind_directions):
+                raise ValueError("price_table first dimension must equal len(wind_directions)")
+            if not price_table.shape[1] == len(wind_speeds):
+                raise ValueError("price_table second dimension must equal len(wind_speeds)")
+            self.price_table = price_table
+        else:
+            self.price_table = None
```
rafmudaf, line 131: Also redundant

floris/tools/wind_data.py
```diff
@@ -0,0 +1,529 @@
...
+        """
+        Return only the variables need for reinitialize
+        """
+        (
+            wind_directions_unpack,
+            wind_speeds_unpack,
+            _,
+            ti_table_unpack,
+            _,
+        ) = self.unpack()
```
rafmudaf, line 39: This is actually doing the opposite of the usual inheritance. The super-class is calling a function that's on the sub-class. At a minimum, there should be an `unpack` function on `WindDataBase` that raises a `NotImplementedError` if it's called.

floris/tools/wind_data.py
```diff
@@ -0,0 +1,529 @@
...
+        Notes:
+            - Returns a resampled version of the wind rose using new `ws_step` and `wd_step`.
+            - Uses the bin weights feature in TimeSeries to resample the wind rose.
+            - If `ws_step` or `wd_step` is not specified, it uses the current values.
+        """
+        if ws_step is None:
+            if len(self.wind_speeds) >= 2:
+                ws_step = self.wind_speeds[1] - self.wind_speeds[0]
+            else:
+                # It doesn't matter, just set to 1
```
rafmudaf, line 235: Some more detail here would improve the readability and help in understanding the logic. What doesn't matter and why doesn't it matter?

floris/tools/wind_data.py
```diff
@@ -0,0 +1,529 @@
...
+
+        # to match wind_rose, make a uniform frequency
+        uniform_frequency = np.ones_like(self.wind_directions)
+        uniform_frequency = uniform_frequency / uniform_frequency.sum()
+
+        return (
+            self.wind_directions.copy(),
+            self.wind_speeds.copy(),
+            uniform_frequency,
+            self.turbulence_intensity,  # can be none so can't copy
```
rafmudaf, line 371: What is the impact of copying the other arrays but not these? in other words, is it safe to _not_ copy these?

rafmudaf (changes requested):

examples/34_wind_data_temporary.py
misi9170: Thanks @rafmudaf . The idea here is that this example will likely grow as more advanced functionality is added to the WindRose and TimeSeries objects; however, the "temporary" in the name did not reflect that well. I've removed the "temporary" in favor of simply "34_wind_data.py". There's also a comment in the file indicating that we will likely grow the example in future.

floris/tools/floris_interface.py
misi9170: This had been reincluded for the purpose of type hinting (see next comment)

floris/tools/floris_interface.py
misi9170: This was possibly a ruff format fix? I've now reorganized as you've suggested

floris/tools/floris_interface.py
misi9170: Same thing, I think this came from ruff format. Now fixed.

floris/tools/floris_interface.py
misi9170: Updated, thanks

floris/tools/floris_interface.py
misi9170: Good catch---no, `wind_data` cannot be None here. I've removed that if statement.

floris/tools/wind_data.py
misi9170: No, never called, good catch

floris/tools/wind_data.py
misi9170: Thanks, this and the one below are updated

floris/tools/wind_data.py
misi9170: .

floris/tools/wind_data.py
misi9170: Thanks for the point here. I've now add an `unpack()` method to the base class that raises a `NotImplementedError` as you suggest, and also added a test to check that works as expected.

floris/tools/wind_data.py
misi9170: @paulf81 and I spoke a bit about this, and decided it was ok _not_ to copy `wind_directions` and `wind_speeds`, as `calculate_wake()` is not expected to modify these in any way (so it's ok to point to the same memory)

paulf81:
Thank you @misi9170  and @rafmudaf !

paulf81:
@rafmudaf  ok to merge?

rafmudaf:
> @rafmudaf ok to merge?
No not yet. I did a mostly syntax review but haven't reviewed the substance. I'll get to that asap.

paulf81:
> > @rafmudaf ok to merge?
> > No not yet. I did a mostly syntax review but haven't reviewed the substance. I'll get to that asap.

ok no problem thanks!

pyproject.toml
```diff
@@ -122,6 +122,8 @@ dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
 "floris/simulation/wake_velocity/jensen.py" = ["F841"]
 "floris/simulation/wake_velocity/gauss.py" = ["F841"]
 "floris/simulation/wake_velocity/empirical_gauss.py" = ["F841"]
+# Ignore `F401` (import violations) in all `__init__.py` files, and in `path/to/file.py`.
+"__init__.py" = ["F401"]
```
rafmudaf, line 126: What's the background for this line?

paulf81: ruff will automatically delete imports in init if it doesn't detect their use, this stops that behavior:
https://github.com/astral-sh/ruff/issues/484

rafmudaf: I wonder how it related to this line: https://github.com/NREL/floris/blob/efd3a9589d30abd516856a36a4b2ca874214917f/pyproject.toml#L83. Is it ruff or isort that makes the change?

floris/tools/floris_interface.py
```diff
@@ -208,6 +198,13 @@ def reinitialize(
 
         # Make the given changes
 
+        # First check if wind data is not None,
+        # if not, get wind speeds, wind direction and
+        # turbulence intensity using the unpack_for_reinitialize
+        # method
+        if wind_data is not None:
+            wind_directions, wind_speeds, turbulence_intensity = wind_data.unpack_for_reinitialize()
```
rafmudaf, line 206: What is the expected behavior if `wind_data` and `wind_direction` are both not None?

floris/tools/floris_interface.py
```diff
@@ -838,6 +825,76 @@ def get_farm_AEP(
...
+
+        Returns:
+            float:
+                The Annual Energy Production (AEP) for the wind farm in
+                watt-hours.
+        """
+
+        # Verify the wind_data object matches FLORIS' initialization
+        if wind_data.n_findex != self.floris.flow_field.n_findex:
+            raise ValueError("WindData object and floris do not have same findex")
```
rafmudaf, line 884: I think think could also be `TimeSeries`, so the message might need to be more generic. Alternatively, you could let the check in `get_farm_AEP` do the error handling.

floris/tools/floris_interface.py
paulf81: Maybe it's best we throw an error in this case, as it is wind_data is used,

floris/tools/floris_interface.py
paulf81: TimeSeries is a WindData object though right?

floris/tools/wind_data.py
paulf81: isn't NDArrayFloat the type hint?

floris/tools/wind_data.py
paulf81: Agreed!

pyproject.toml
paulf81: ruff, I had hoped that that line in pyproject would handle it, but some how when you run ruff through pre-commit this can be ignored

floris/tools/wind_data.py
paulf81: Gotcha, so this comment is good then, my bad

floris/tools/wind_data.py
```diff
@@ -0,0 +1,536 @@
...
+class WindRose(WindDataBase):
+    """
+    In FLORIS v4, the WindRose class is used to drive FLORIS and optimization
+    operations in which the inflow is characterized by the frequency of
+    binned wind speed, wind direction and turbulence intensity values
+
+    Args:
+        wind_directions: NumPy array of wind directions (NDArrayFloat).
+        wind_speeds: NumPy array of wind speeds (NDArrayFloat).
+        freq_table: Frequency table for binned wind direction, wind speed
```
rafmudaf, line 80: The unit tests nicely capture that the frequency has to have the shape `(n wind directions, n wind speeds)`, so that should also be reflected here

rafmudaf: At least, thats how I understood the unit tests, so let me know if I'm incorrect

tests/wind_data_test.py
```diff
@@ -0,0 +1,264 @@
...
+    assert np.allclose(wind_rose.wind_directions, [260, 262, 264])
+
+    # Freq table should have dimension of 3 wd x 1 ws because the wind speeds
+    # are all binned to the same value given the `ws_step` size
+    freq_table = wind_rose.freq_table
+    assert freq_table.shape[0] == 3
+    assert freq_table.shape[1] == 1
+
+    # The frequencies should [2/3, 0, 1/3]
+    # TODO this one I dont follow, @paul or @misha?
```
rafmudaf, line 208: @paulf81 @misi9170 could one of you add some detail here? Why should the frequencies be 2/3, 0, 1/3

examples/34_wind_data_temporary.py
rafmudaf: Ok sounds good, I've added a card on the project for this

floris/tools/floris_interface.py
rafmudaf: Ah got it. It's now not comments and used for type hinting. Thanks

floris/tools/wind_data.py
rafmudaf: Yes that's better, thanks

rafmudaf:
Thanks @misi9170 and @paulf81, my comments are mostly addressed. Let me know when the remaining comments are addresses, and I'll resolve the review and approve.

floris/tools/wind_data.py
paulf81: Done!

paulf81:
> Thanks @misi9170 and @paulf81, my comments are mostly addressed. Let me know when the remaining comments are addresses, and I'll resolve the review and approve.

Thank @rafmudaf !

tests/wind_data_test.py
paulf81: how is this?

floris/tools/floris_interface.py
paulf81: how is this @rafmudaf?

paulf81:
@rafmudaf I think all comments now addressed, thank you!

tests/wind_data_test.py
rafmudaf: That's great, thank you

ejsimley (changes requested):
See the comments I left, but this looks good overall!

floris/tools/floris_interface.py
paulf81: This is addressed in the next pull request, #782

floris/tools/floris_interface.py
```diff
@@ -838,6 +851,76 @@ def get_farm_AEP(
 
         return aep
 
+    def get_farm_AEP_with_wind_data(
```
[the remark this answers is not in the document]
paulf81: This function is a light wrapper around the other (it calls get_farm_AEP), @misi9170 and I thought this was pretty clean, the other way works too but then you have ambiguous situations depending which values are passed in

floris/tools/wind_data.py
```diff
@@ -0,0 +1,539 @@
...
+            values (NDArrayFloat, optional).   Must have dimension
+            (n_wind_directions, n_wind_speeds).  Defaults to None.
```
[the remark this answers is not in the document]
paulf81: That is a good suggestion, implementing now

floris/tools/wind_data.py
```diff
@@ -0,0 +1,539 @@
...
+        ti_table: Turbulence intensity table for binned wind direction, wind
+            speed values (NDArrayFloat, optional).   Must have dimension
+            (n_wind_directions, n_wind_speeds). Defaults to None.
+        price_table: Price table for binned binned wind direction, wind
```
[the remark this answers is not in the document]
paulf81: Good call!  Implementing now,

floris/tools/wind_data.py
```diff
@@ -0,0 +1,539 @@
...
+    def __init__(
+        self,
+        wind_directions: NDArrayFloat,
+        wind_speeds: NDArrayFloat,
+        turbulence_intensity: NDArrayFloat | None = None,
+        prices: NDArrayFloat | None = None,
+    ):
+        # Wind speeds and wind directions must be the same length
+        if len(wind_directions) != len(wind_speeds):
+            raise ValueError("wind_directions and wind_speeds must be the same length")
```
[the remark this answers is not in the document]
paulf81: Done!

floris/tools/wind_data.py
```diff
@@ -0,0 +1,539 @@
...
+        else:
+            wd_edges = np.arange(0.0 - wd_step / 2.0, 360.0, wd_step)
+
+            # use wd_step to produce a wrapped version of wind_directions
+            wind_directions_wrapped = self._wrap_wind_directions_near_360(
+                self.wind_directions, wd_step
+            )
+
+            # Only keep the range with values in it
+            wd_edges = wd_edges[wd_edges + wd_step >= wind_directions_wrapped.min()]
```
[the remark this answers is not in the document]
paulf81: You are right, that is a good catch, thank you!

floris/tools/wind_data.py
```diff
@@ -0,0 +1,539 @@
...
+import matplotlib.cm as cm
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+from pandas.api.types import CategoricalDtype
+
+from floris.type_dec import NDArrayFloat
+
+
+class WindDataBase:
```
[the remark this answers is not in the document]
paulf81: I actually don't know!  @rafmudaf or @misi9170, should I it inherit from LoggingManager?

paulf81:
Thank you @ejsimley  and @rafmudaf  for the round of comments!  I think all (except for the one question for you @rafmudaf  and @misi9170 on whether WindData should inherit LoggingManager), this one should be good to go?  I'll hit request re-review now...

floris/tools/floris_interface.py
rafmudaf: I think the change here should be either or both of:
- "WindData" becomes the name of the variable "wind_data"
- Give the user some information relevant to the use case such as "The number of atmospheric conditions in `wind_data` does not match the number of conditions used to initialize FLORIS. TODO: Here's a top on getting past this error based on how we expect you to use this function."

floris/tools/wind_data.py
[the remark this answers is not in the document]
rafmudaf: It wouldn't hurt to have the logging functionality available. One immediate use case might be to give some additional information when the length of the WindRose or TimeSeries don't match the length of findex.

floris/tools/__init__.py
```diff
@@ -46,18 +46,22 @@
     visualize_cut_plane,
     visualize_quiver,
 )
-from .wind_rose import WindRose
+from .wind_data import (
+    TimeSeries,
+    WindDataBase,
```
rafmudaf, line 51: I don't think users will be using this base-class, they'll probably only use the derived classes. So we can remove `WindDataBase` from the user-facing imports.

paulf81: Done!

paulf81:
I fixed the issue that cropped up after not importing Base, let me know if ready to merge, thank you @rafmudaf !

floris/tools/floris_interface.py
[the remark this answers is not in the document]
ejsimley: OK, I think that makes sense. But I think this function would be a little easier to use if it didn't require wind_data to "match the wind_data object passed to reinitialize()" as mentioned in the docstring. What if you added the line `self.reinitialize(wind_data)` so users wouldn't have to call that separately beforehand if they wanted to update the wind conditions?

ejsimley (commented):
Thanks for updating Paul. I just have one comment on `get_farm_AEP_with_wind_data` to consider.

floris/tools/floris_interface.py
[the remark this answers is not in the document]
ejsimley: ok, no worries! Since that's the convention in get_farm_AEP, no need to change anything.

ejsimley (approved):

paulf81:
thanks @ejsimley !  @rafmudaf, merge city and I'm the mayor?

rafmudaf (approved):

PR #781, opened 2024-01-18
Add support for multiple turbine types in calculate_horizontal_plane_with_turbines()

scottryn:
# Add support for multiple turbine types in calculate_horizontal_plane_with_turbines()

Function now copies turbine type list from input FlorisInterface and appends a `nrel_5MW` to ensure correct number of turbines in `fi.reinitialize()` calls.

## Related issue

No related open issues.

floris.tools.visualization.py

Small modification to prevent `turbine_type must have the same number of entries as layout_x/layout_y or have a single turbine_type value` error when generating a CutPlane using `calculate_horizontal_plane_with_turbines()` with multiple turbine definitions. Previously the test turbine was left undefined leading to a potential mismatch between the layout arguments and number of turbines if `turbine_type` was specified in earlier reinitializations.

### DIFF
--- a/floris/tools/visualization.py
+++ b/floris/tools/visualization.py
@@ -697,8 +697,8 @@ def calculate_horizontal_plane_with_turbines(
                 layout_x_test[-1] = x
                 layout_y_test[-1] = y
                 fi.reinitialize(
-                    layout_x=layout_x_test, 
-                    layout_y=layout_y_test, 
+                    layout_x=layout_x_test,
+                    layout_y=layout_y_test,
                     turbine_type=turbine_types_test
                 )
                 fi.calculate_wake(yaw_angles=yaw_angles)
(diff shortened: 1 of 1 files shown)


rafmudaf:
I've assigned myself to handle this pull request, but @paulf81 since you wrote this function could you review it first?

paulf81 (approved):
Approved, reviewed the code changes and this is a good solution to the issue

paulf81:
Actually hold on, let me debug the example

paulf81:
@rafmudaf I pushed a small change that I think fixes the issue that was calling example 02 to fail.  In a nutshell, the turbine type array in that case is length 1, which farm.py knows to expand to n_turbines using:
_turbine_types *= self.n_turbines (line 201)

I added a similar catch/logic here

paulf81 (approved):
All tests/examples passing now so approving

Issue #778, opened 2024-01-17
Allow turbulence intensity to vary across all f_index 


paulf81:
# Allow turbulence intensity to vary across all f_index 
This issue is meant to capture within it Issue #666  and Issue #713 by building off the new wind rose structure in pull request #775 and accomplishing the following improvements:

1. Turbulence Intensity can be passed in as a float (uniform across all findex, current setup) or as an array with length n_findex that defines a TI individiually for each findex
2. This change is made and chained through floris_interface, floris, flowfield
3. WindRose and TimeSeries are updated to take advantage of this capability
4. Functions are added to WindRose and TimeSeries that TI can be generated either by an arbitrary function, or a standard IEC relationship

paulf81:
This is accomplished in #782

PR #782, opened 2024-01-19
Convert turbulence intensity from single value to n_findex length array

paulf81:
# Convert turbulence intensity from single value to n_findex length array

This pull request builds on Pull Request #775 , and is intended to address Issue #778 and by extension issues #666 and #713.  It converts turbulence_intensity (float) to turbulence_intensities (np.array).  Further it enables turbulence_intensities to be specified:

- As an array within the yaml
- As an array to fi.reinitialize
- As an array to the TimeSeries object
- As a wd x ws table to WindRose object

Or to be generated via helper functions within the WindRose and TimeSeries objects.  In more detail:

- [x] Rename flow_field's turbulence_intensity to turbulence_intensities
- [x] Refactor the simulation files to expect turbulence_intensities to be an array, rather than a float
- [x] Add additional tests of turbulence intensity to confirm correct behavior of new features
- [x] Complete WindRose and TimeSeries handling of turbulence intensities
- [x] Add helper functions to WindRose and TimeSeries which allow turbulence intensities to be generated, rather than provided, as a function of wind directions and wind speeds
- [x] Add additional examples of usage

All tests passing.


Bartdoekemeijer:
This seems really great! Both being able to run it in timeseries mode and as a wd x ws table will be super useful. I like how the WindRose object is being reformatted to facilitate a FLORIS v3-like workflow and that the default in v4 is defaulting towards a timeseries mode.

paulf81:
This pull request includes a number of un-related format changes, going to close it and re-open a pull request without them

paulf81:
@rafmudaf , I think I've not gotten rid of all the extraneous formatting changes so this should be again ready for your review.  I think I also fixed my vscode so this shouldn't happen again, thank you!  It will probably be a little easier to review after we merge #775 , since this is built on top of that

paulf81:
I'm glad and sorry again for the headache!

paulf81:
@rafmudaf , @misi9170 @misi9170 , finished cleaning out the format bric-a-brac, and with #775 merged into V4 and a few updates to docs this pull request is truly ready for review when you get a chance!

Bartdoekemeijer:
Just ran the examples and this looks really great. Thanks for all the work! Something I used a lot in the old `WindRose` class is the `interpolate(...)` function to upsample the provided frequency table (usually in steps of 30 deg) to a higher resolution for FLORIS evaluation (e.g., 2 or 3 deg resolution). I think this is covered by `resample_wind_rose` so should be good. If it's not, I can always implement it later. Good from my side!

tests/data/input_full_v3.yaml
rafmudaf, line 1: I think this will not work with the v3 API

floris/simulation/flow_field.py
```diff
@@ -66,6 +66,22 @@ class FlowField(BaseClass):
...
+    def turbulence_intensities_validator(
+        self, instance: attrs.Attribute, value: NDArrayFloat
+    ) -> None:
+        try:
+            # Check the turbulence intensity is either length 1 or n_findex
+            if len(value) != 1 and len(value) != self.n_findex:
+                raise ValueError("turbulence_intensities should either be length 1 or n_findex")
+        except TypeError as te:
+            # Handle the TypeError here
+            print(f"Caught a TypeError: {te}")
```
rafmudaf, line 79: This should either be printed though the logger or included in the message in the TypeError

floris/simulation/flow_field.py
```diff
@@ -66,6 +66,22 @@ class FlowField(BaseClass):
...
 
+    @turbulence_intensities.validator
+    def turbulence_intensities_validator(
+        self, instance: attrs.Attribute, value: NDArrayFloat
+    ) -> None:
+        try:
+            # Check the turbulence intensity is either length 1 or n_findex
+            if len(value) != 1 and len(value) != self.n_findex:
+                raise ValueError("turbulence_intensities should either be length 1 or n_findex")
+        except TypeError as te:
```
rafmudaf, line 77: What is the TypeError that would be caught here? Note that [attrs converters](https://www.attrs.org/en/stable/init.html#converters) are run before validators, so `value` here is guaranteed to be a `np.array`.

floris/simulation/flow_field.py
```diff
@@ -197,14 +217,13 @@ def initialize_velocity_field(self, grid: Grid) -> None:
...
 
-        self.turbulence_intensity_field = self.turbulence_intensity * np.ones(
-            (
-                self.n_findex,
-                grid.n_turbines,
-                1,
-                1,
-            )
+        self.turbulence_intensity_field = self.turbulence_intensities[
+            :, np.newaxis, np.newaxis, np.newaxis
```
rafmudaf, line 221: We've used `None` instead of `np.newaxis` throughout the code, so I suggest to keep that consistent.

floris/simulation/solver.py
```diff
@@ -76,11 +76,17 @@ def sequential_solver(
...
     w_wake = np.zeros_like(flow_field.w_initial_sorted)
 
-    turbine_turbulence_intensity = (
-        flow_field.turbulence_intensity
-        * np.ones((flow_field.n_findex, farm.n_turbines, 1, 1))
-    )
-    ambient_turbulence_intensity = flow_field.turbulence_intensity
+    # Set up turbulence arrays
+    turbine_turbulence_intensity = flow_field.turbulence_intensities[
+        :, np.newaxis, np.newaxis, np.newaxis
```
rafmudaf, line 81: Use `None` instead of `np.newaxis` for consistency.
Also, I see this same array expansion happens in `FlowField`. What's the relationship between this local variable and the attribute on `FlowField`?

rafmudaf (changes requested):

floris/simulation/flow_field.py
paulf81: I was having trouble catching the issue if you pass something through that doesn't have a len operation, even if it is of type np.array.  The example I found was that I would pass in a float, this would be automatically be upped to an np.array, but it would be a special case without a len operation, calling len(value) produced an error that was not screened out by confirming it is of type np.array

floris/simulation/flow_field.py
paulf81: Done!

floris/simulation/flow_field.py
paulf81: Done!

floris/simulation/solver.py
paulf81: Done on the conversion!  On the other point, I'm not positive actually, here I was just following the pre-exisiting patterns for which variables were local and which were on FlowField

tests/data/input_full_v3.yaml
paulf81: Good catch

rafmudaf: Ignore this comment, I incorrectly assumed this file was meant to maintain the v3 API.

floris/simulation/flow_field.py
rafmudaf: Does the change in https://github.com/NREL/floris/pull/782/commits/598aa81293b61ceb5c7c340b68a8a2d533ff4086 affect this?

paulf81: I think it does, if the floris_array_converter comes before this one, then there is no possibility that the len() call could generate a TypeError, would you like me to remove that try/except code on those grounds?

rafmudaf: Yep that would be great

rafmudaf: Or more specifically, check if your previous commit affects this code and update as needed

paulf81: no problem, this is actually pretty easy to do, there is now a specific check of this case in test_reinitailize_ti (line 258), where it passes a float in, you can see that the expected TypeError is raised before this code is executed

floris/tools/optimization/yaw_optimization/yaw_optimization_base.py
```diff
@@ -530,20 +530,26 @@ def _finalize(self, farm_power_opt_subset=None, yaw_angles_opt_subset=None):
         self.yaw_angles_opt = self._unreduce_variable(yaw_angles_opt_subset)
 
         # Produce output table
-        ti = np.min(self.fi.floris.flow_field.turbulence_intensity)
+        ti = np.min(self.fi.floris.flow_field.turbulence_intensities)
```
ejsimley, line 533: I think the optimization classes are getting updated anyway, so this will probably get fixed elsewhere, but it doesn't seem like using the minimum TI value from turbulence_intensities is correct here.

floris/tools/wind_data.py
```diff
@@ -333,6 +333,67 @@ def plot_wind_rose(
...
+
+        Args:
+            func (function): Function which accepts wind_directions as its
+                first argument and wind_speeds as second argument and returns
+                turbulence_intensities
+        """
+        self.ti_table = func(self.wd_grid, self.ws_grid)
+        self._build_gridded_and_flattened_version()
+
+    def assign_ti_using_IEC_method(self, Iref=0.08, offset=3.8):
```
ejsimley, line 348: Iref=0.08 gives a TI of 9.8% at 8 m/s. That seems a little high compared to what is usually used in FLORIS. Maybe we could change this to 0.07?

ejsimley (commented):
No major comments, except for the impact on the yaw optimization classes (but I think that is addressed elsewhere). I just added some more details to the doc strings about the IEC TI method.

floris/tools/optimization/yaw_optimization/yaw_optimization_base.py
paulf81: You're right, I think this will get reconciled either in the current #790 which is updated all optimization to 4d (@bayc)?  Or else a second follow-on pull request which will reconfigure around TimeSeries and WindRose objects

floris/tools/wind_data.py
paulf81: Done!  Thank you!

paulf81:
ok @ejsimley @rafmudaf and @misi9170 , I think all comments addressed, good to merge and clean shop before the weekend?  Thanks!!

misi9170:
@paulf81 I was just cleaning up the plotting a bit for example 35, and noticed that the rear turbine power doesn't vary very smoothly with TI:

image

Is that expected? If not, it's likely not an issue with this PR per se, but something we may need to look into

examples/inputs/gch.yaml
```diff
@@ -112,7 +112,8 @@ flow_field:
 
   ###
   # The level of turbulence intensity level in the wind.
-  turbulence_intensity: 0.06
+  turbulence_intensities:
+  - 0.06
```
misi9170, line 116: This is the only change so far to the floris input files from v3 to v4 (that `turbulence_intensity`, a scalar, changes to `turbulence_intensities`, a list). When running the new code with an v3 input file, the generic error gets raised:

`AttributeError: The initialization for FlowField was given extraneous inputs: ['turbulence_intensity']`

Is that OK, or do we need either a converter to convert v3 models to v4 (similar to what we now have for v3-style turbine models) or a more explicit explanation of the change in the error raised?

floris/tools/floris_interface_legacy_reader.py
```diff
@@ -123,7 +123,7 @@ def _convert_v24_dictionary_to_v3(dict_legacy):
     dict_floris["flow_field"] = {
         "air_density": fp["air_density"],
         "reference_wind_height": ref_height,
-        "turbulence_intensity": fp["turbulence_intensity"][0],
+        "turbulence_intensities": fp["turbulence_intensities"][0],
```
misi9170, line 126: I don't think this should be changed; this function is called `_convert_v24_dictionary_to_v3`, so it would be confusing if it didn't take as an input a valid v2 model and return a valid v3 model. However, perhaps the entire floris_interface_legacy_reader.py module is going away?

floris/tools/optimization/legacy/scipy/yaw_wind_rose.py
```diff
@@ -243,7 +243,7 @@ def _get_initial_farm_power(self):
                     self.fi.reinitialize_flow_field(
                         wind_direction=[self.wd[i]],
                         wind_speed=[self.ws[i]],
-                        turbulence_intensity=self.ti[i],
+                        turbulence_intensities=[self.ti[i]],
```
misi9170, line 246: I'm not sure what the optimization/legacy/scipy/ package refers to---do these currently run with v3, and are they expected to run with v4? If not the latter (not expected to run with v4), I'd be in favor of returning these to `turbulence_intensity` so that it does not appear to be updated to v4 compatibility.

In that case, I'd just revert all files in optimization/legacy/scipy.

floris/tools/optimization/yaw_optimization/yaw_optimization_base.py
misi9170: We should create an issue for this to make sure we track it?

misi9170 (changes requested):
I think we should revert some of the legacy files to `turbulence_intensity` to avoid confusion about forward compatibility; also, I'm not sure whether we need a v3 to v4 input file converter now or whether we should raise a less generic error if users still have `turbulence_intensity` on their input files

paulf81:
> @paulf81 I was just cleaning up the plotting a bit for example 35, and noticed that the rear turbine power doesn't vary very smoothly with TI:
> 
> image
> 
> Is that expected? If not, it's likely not an issue with this PR per se, but something we may need to look into

I think this looks like the near-wake boundary in GCH, which is has this sharp change, my guess is that length of the near wake depends on TI so we cross it and transition from linear to gaussian wake

examples/inputs/gch.yaml
paulf81: Both sound good to me, we could add a specific error for when turbulence_intensity is provided, and maybe (with a seperate issue) prepare a converter?

misi9170:
> > @paulf81 I was just cleaning up the plotting a bit for example 35, and noticed that the rear turbine power doesn't vary very smoothly with TI:
> > image
> > Is that expected? If not, it's likely not an issue with this PR per se, but something we may need to look into
> 
> I think this looks like the near-wake boundary in GCH, which is has this sharp change, my guess is that length of the near wake depends on TI so we cross it and transition from linear to gaussian wake

That would make sense. Is there also another more subtle kink at about 14% TI, or am I imagining that?

floris/tools/optimization/legacy/scipy/yaw_wind_rose.py
paulf81: that works for me, or delete?  The legacy could, if it's for v3, can just remain attached to v3 code?  Do we need a legacy folder I guess?

paulf81:
> > > @paulf81 I was just cleaning up the plotting a bit for example 35, and noticed that the rear turbine power doesn't vary very smoothly with TI:
> > > image
> > > Is that expected? If not, it's likely not an issue with this PR per se, but something we may need to look into
> > 
> > 
> > I think this looks like the near-wake boundary in GCH, which is has this sharp change, my guess is that length of the near wake depends on TI so we cross it and transition from linear to gaussian wake
> 
> That would make sense. Is there also another more subtle kink at about 14% TI, or am I imagining that?

There might be an inflection there, but I don't 'have a ready answer if so...

floris/tools/optimization/legacy/scipy/yaw_wind_rose.py
misi9170: It can probably be deleted, but let's save that for a separate PR, so I'd say just change them back for this one.

paulf81: agree changing now

examples/35_sweep_ti.py
rafmudaf, line 1: I just want to point out that we're adding more and more example. This one and 36 are both fairly short and simple, and I could see them easily being consolidated into one, worked into one of 4 through 6, or being included in the Getting Started notebook. In any case, these should be described in the examples listing in the documentation.

paulf81: Oh definitely, I don't think the examples should stay as they are now, but reconsolidated in a separate pull request, but I was saving that effort until nearer the end when all the tools are ready.  Examples 35 and 36 will be folded into the earlier examples but for now I just wanted to produce some temp example

misi9170 (approved):
This looks fine to me. I'm OK for this to be merged, with the understanding that there may be some to do items left for later PRs:
- possibly raising a specific error if `turbulence_intensity` is specified in the input yaml, and/or creating a converter to write new v4 yamls
- consolidating/removing the examples, which can happen when we propagate the TimeSeries and WindRose data structures more thoroughly through the examples
- creating an issue for that `min` operation in the YawOptimizerBase.

ejsimley (approved):
Looks good to me. I'd suggest creating issues or at least project board cards for the points @misi9170 listed before merging so we don't lose track of them.

paulf81:
Thanks guys!  I can do those new issues before merging, ok for you @rafmudaf ?

rafmudaf (approved):

paulf81:
@rafmudaf @ejsimley @misi9170 @bayc 

I've added an issue and several project cards to track the issues not fully resolved by this pull request.  

Issue #791 
https://github.com/orgs/NREL/projects/96/views/2?pane=issue&itemId=52095313
https://github.com/orgs/NREL/projects/96/views/2?pane=issue&itemId=52095412
https://github.com/orgs/NREL/projects/96/views/2?pane=issue&itemId=52095420

Have put these all toward the end of the process because I believe they are best done after other actions are completed.  But it's important we not forget so including these here.

PR #784, opened 2024-01-25
Remove small variations in above rated power

paulf81:
# Remove small variations in above rated power

Small variations in power (say 0.01 kW with power at 5000 kW) have very small impact on power and are probably simply coming from minor numerical variations in the code used to derive them.  However, they can create useless (and possibly significant) gradients for different optimization strategies.  With the new power-based input yaml implemented in #765  it's easy to ensure region 3 power is perfectly flat.  This pull request adjusts the 3 turbine yaml files included to have flat power above region 3.

## Related Issues

This should firmly address Issue #554 as well as #583 

## Other notes

Change is quite small and didn't require updating any unit or regression tests


paulf81:
> @paulf81 The changes here wouldn't be used in the regression tests since they aren't defined in the conftest file. For consistency, we should maintain the same NREL 5MW definition in [conftest.py](https://github.com/paulf81/floris/blob/feature/flatten_v3/tests/conftest.py#L213).

Done!

misi9170:
> > @paulf81 The changes here wouldn't be used in the regression tests since they aren't defined in the conftest file. For consistency, we should maintain the same NREL 5MW definition in [conftest.py](https://github.com/paulf81/floris/blob/feature/flatten_v3/tests/conftest.py#L213).
> 
> Done!

Nice to see no change to reg test results, although I guess those mostly (possibly all) use below-rated wind speeds. Anyway, no issue there.

misi9170 (approved):
Looks good to me.

rafmudaf (approved):

PR #790, opened 2024-02-01
Update yaw and layout optimization tools for 4D

bayc:
# Update yaw and layout optimization tools and examples to new 4D format.

This PR makes the changes necessary for the yaw and layout optimization capabilities in FLORIS to support the new 4D wind condition format. It updates both the necessary tools and the related examples. It mainly focuses on eliminating the now extra dimension and introducing a wind speed array and a turbulence intensity array where necessary. It also removes the layout symmetry functionality as it would not work well with the new timeseries focus of the inputs.

## Related issue
None

`floris/tools/opitmization`
`examples`

These changes were evaluated compared to the execution and results of the same examples in the v3 develop branch (a88a1543). Note that for the examples to produce the exact same results, the older turbine performance data needed to be used with the new turbine definition file format (specifically, this version of the NREL 5MW was used: https://github.com/NREL/floris/blob/8e6fb6b54c1a7824d801499cf90037dd4bb59fb4/floris/turbine_library/nrel_5MW_v4converted.yaml).

Note that with the new support for time series as input, certain features were excluded from this update, specifically `exploit_layout_symmetry`. The related code was removed.

Also, `verify_convergence` is currently broken and needs to be updated. I noticed results that would eliminate yaw angles under certain conditions (e.g. a full wind rose) and not under other (e.g. a couple wind directions with one wind speed). I believe this could be due to some normalization that is happening, but I have not confirmed that.

I believe it would be worth while to add tests to some basic optimization results, just so we can track if results shift through other code development or updates in dependencies.

See above to produce test results.


floris/tools/optimization/yaw_optimization/yaw_optimization_base.py
```diff
@@ -42,7 +42,7 @@ def __init__(
         normalize_control_variables=False,
         calc_baseline_power=True,
         exclude_downstream_turbines=True,
-        exploit_layout_symmetry=True,
+        exploit_layout_symmetry=False,
```
paulf81, line 45: @bayc, @misi9170 and I were discussing if it makes sense to keep this option in the n_findex refactor, it seems like there are edge cases where this could not function as we expect without explicit wd/ws indexing.  Plus, I don't think it was a commonly applied feature since it required exact symmetry of the farm, and all turbines operating identically.  Seems safer to delete everywhere?

floris/tools/optimization/yaw_optimization/yaw_optimization_base.py
```diff
@@ -264,7 +259,7 @@ def _reduce_control_problem(self):
 
         # Initialize subset variables as full set
         self.fi_subset = self.fi.copy()
-        nwinddirections_subset = copy.deepcopy(self.fi.floris.flow_field.n_wind_directions)
+        n_wind_directions_subset = copy.deepcopy(self.fi.floris.flow_field.n_findex)
```
paulf81, line 262: Ah, now I see, you have it working, maybe ignore comment above if all is fine?

floris/tools/optimization/yaw_optimization/yaw_optimizer_scipy.py
```diff
@@ -83,68 +83,72 @@ def optimize(self):
...
-                    fun=cost,
-                    x0=x0,
-                    bounds=bnds,
-                    method=self.opt_method,
-                    options=self.opt_options,
-                )
+        for i, wc in enumerate(zip(wd_array, ws_array)):
+
+            wd = wc[0]
+            ws = wc[1]
```
paulf81, line 92: I think this can be shortened to:

for i, (wd, ws) in enumerate(zip(wd_array, ws_array)):

paulf81:
This looks great @bayc !   I just left a few small comments for your considerations

floris/tools/optimization/yaw_optimization/yaw_optimization_base.py
bayc: I have the defaults for it set to `False` everywhere, but we could also remove it outright if you think that would be better.

floris/tools/optimization/yaw_optimization/yaw_optimization_base.py
bayc: I think it could use some more testing to make sure it is working 100% as it should.

examples/12_optimize_yaw.py
```diff
@@ -74,23 +74,21 @@ def calculate_aep(fi, df_windrose, column_name="farm_power"):
...
     # Derive the wind directions and speeds we need to evaluate in FLORIS
-    wd_array = np.array(df_windrose["wd"].unique(), dtype=float)
-    ws_array = np.array(df_windrose["ws"].unique(), dtype=float)
+    wd_array = np.array(df_windrose["wd"], dtype=float)
+    ws_array = np.array(df_windrose["ws"], dtype=float)
     yaw_angles = np.array(df_windrose[yaw_cols], dtype=float)
     fi.reinitialize(wind_directions=wd_array, wind_speeds=ws_array)
 
     # Map angles from dataframe onto floris wind direction/speed grid
     X, Y = np.meshgrid(wd_array, ws_array, indexing='ij')
```
ejsimley, line 83: I don't think this line is necessary, since it doesn't appear X and Y are used anywhere.

floris/tools/optimization/yaw_optimization/yaw_optimization_base.py
```diff
@@ -533,23 +549,24 @@ def _finalize(self, farm_power_opt_subset=None, yaw_angles_opt_subset=None):
...
-                        "farm_power_opt": None
-                        if self.farm_power_opt is None
-                        else self.farm_power_opt[:, ii],
-                        "farm_power_baseline": None
-                        if self.farm_power_baseline is None
-                        else self.farm_power_baseline[:, ii],
-                    }
-                )
+        wd_array = self.fi.floris.flow_field.wind_directions
+        ws_array = self.fi.floris.flow_field.wind_speeds * np.ones(num_wind_directions)
```
ejsimley, line 553: `self.fi.floris.flow_field.wind_speeds` should be the same length as `self.fi.floris.flow_field.wind_directions` now, so I don't think the multiplication is needed here.

floris/tools/optimization/yaw_optimization/yaw_optimization_base.py
```diff
@@ -533,23 +549,24 @@ def _finalize(self, farm_power_opt_subset=None, yaw_angles_opt_subset=None):
...
+        df_list.append(
+            pd.DataFrame(
+                {
+                    "wind_direction": wd_array,
+                    "wind_speed": ws_array,
+                    "turbulence_intensity": ti * np.ones(num_wind_directions),
```
ejsimley, line 559: similarly, since `self.fi.floris.flow_field.turbulence_intensities` is an array now, I think this should just be `self.fi.floris.flow_field.turbulence_intensities`.

ejsimley (commented):
Thanks for taking care of this @bayc. I left a few comments, but one other comment throughout is that in addition to introducing the wind speed array to a lot of the optimization function calls, does the TI array also need to be added? (for example in `fi.reinitialize` and ` self._calculate_farm_power`) I think so because `self.fi.floris.flow_field.turbulence_intensities` is now defined as an array with the same length as the wind direction and wind speed arrays.

floris/tools/optimization/yaw_optimization/yaw_optimization_base.py
bayc: Yea, there are a lot of places it would need to be updated. I have removed it for now, we can add it back in later if we feel the need to.

examples/12_optimize_yaw.py
bayc: Good catch!

bayc:
> Thanks for taking care of this @bayc. I left a few comments, but one other comment throughout is that in addition to introducing the wind speed array to a lot of the optimization function calls, does the TI array also need to be added? (for example in `fi.reinitialize` and ` self._calculate_farm_power`) I think so because `self.fi.floris.flow_field.turbulence_intensities` is now defined as an array with the same length as the wind direction and wind speed arrays.

Good call @ejsimley , I have updated throughout where I believe the `ti_array` should be added, if you want to look at the latest commits.

floris/tools/optimization/yaw_optimization/yaw_optimization_base.py
paulf81: Thanks!

paulf81 (approved):
Thank you for addressing my comments @bayc , approve!

misi9170 (approved):
I've looked through the changes and pulled down and run the examples---all looks good to me. Thanks @bayc!

floris/tools/parallel_computing_interface.py
```diff
@@ -87,10 +84,8 @@ def __init__(
             object or can be an UncertaintyInterface object.
         max_workers (int): Number of parallel workers, typically equal to the number of cores
             you have on your system or HPC.
-        n_wind_direction_splits (int): Number of sectors to split the wind direction array over.
+        n_wind_condition_splits (int): Number of sectors to split the wind condition array over.
```
rafmudaf, line 87: In `floris.simulation`, this is generally called `findex` (see https://github.com/NREL/floris/discussions/753). For consistency, it may be best to stick with that convention.

floris/tools/optimization/yaw_optimization/yaw_optimization_base.py
```diff
@@ -255,16 +236,14 @@ def _reduce_control_problem(self):
...
-        exploit_layout_symmetry == True.
+        exclude_downstream_turbines == True.
         """
         # Initialize which turbines to optimize for
         self.turbs_to_opt = (self.maximum_yaw_angle - self.minimum_yaw_angle >= 0.001)
 
         # Initialize subset variables as full set
         self.fi_subset = self.fi.copy()
-        nwinddirections_subset = copy.deepcopy(self.fi.floris.flow_field.n_wind_directions)
+        n_wind_directions_subset = copy.deepcopy(self.fi.floris.flow_field.n_findex)
```
rafmudaf, line 246: Should this be `n_findex_subset` for consistency?

floris/tools/parallel_computing_interface.py
```diff
@@ -483,10 +445,14 @@ def get_farm_AEP(
...
         if np.any(conditions_to_evaluate):
             wind_speeds_subset = wind_speeds[conditions_to_evaluate]
+            wind_direction_subset = wind_directions[conditions_to_evaluate]
             yaw_angles_subset = None
             if yaw_angles is not None:
                 yaw_angles_subset = yaw_angles[:, conditions_to_evaluate]
-            self.fi.reinitialize(wind_speeds=wind_speeds_subset)
+            self.fi.reinitialize(
+                wind_directions=wind_direction_subset,
+                wind_speeds=wind_speeds_subset
```
ejsimley, line 454: I think this is another place where the turbulence_intensities subset should get passed

floris/tools/parallel_computing_interface.py
```diff
@@ -495,7 +461,7 @@ def get_farm_AEP(
         aep = np.sum(np.multiply(freq, farm_power) * 365 * 24)
 
         # Reset the FLORIS object to the full wind speed array
-        self.fi.reinitialize(wind_speeds=wind_speeds)
+        self.fi.reinitialize(wind_directions=wind_directions, wind_speeds=wind_speeds)
```
ejsimley, line 464: and turbulence_Intensities should be reset here too

ejsimley (commented):
Hi @bayc, the changes to add the turbulence intensity array inputs look good. I just found a couple places in parallel_computing_interface where I think they're needed as well.

floris/tools/parallel_computing_interface.py
bayc: Updated!

floris/tools/optimization/yaw_optimization/yaw_optimization_base.py
bayc: Good call.

ejsimley (approved):

PR #783, opened 2024-01-19
Add de-rating op model

paulf81:
# Add de-rating op model

Adds an operation model to implement a simple de-rating strategy. Also adds a "mixed" operation model that essentially allows switching between various different operation models for different findices/turbine, which may be extended in future.

Shutting down turbines will be a natural use case for this, but this will appear in a separate PR (see also #687, #693)
Taking the steps from: https://github.com/orgs/NREL/projects/96?pane=issue&itemId=50191456

Steps:
- [x] Create placeholder operating model that supports derating @misi9170 
- [x] Implement operating model that supports derating @paulf81  
- [x] Create tests to check operating model @paulf81, @misi9170  
- [x] Pass necessary control setpoints through from FlorisInterface() for derating @paulf81 
- [ ] ~Pass necessary control setpoints through from FlorisInterface() for shutdown~ @misi9170 (see #693, may need reworking to align)
- [x] Create example of derating, shutting turbines off, etc. @paulf81 , @misi9170

PR #793, opened 2024-02-07
Validate and test wind direction and wind speed

paulf81:
# Validate and test wind direction and wind speed

In the development of #783, @misi9170 noticed that running this code:

```
fi.reinitialize(
    wind_directions=270 * np.ones_like(yaw_angles),
    wind_speeds=10.0 * np.ones_like(yaw_angles),
    turbine_type=[turbine_type]*2
)
```

Which has the wrong number of dimensions for wind speed and direction, produces an error related to turbulence intensity.  Experimenting with the code I found that, contrary to expectations, passing in wind directions and wind speeds of different length does not produce an error, instead f_index = len(wind_directions), which is confusing.  Then also, having extra dimensions in wd/ws does not immediately error, explaining the complicated errors later.

This pull request attempts to clarify the issue by:

1) Adding to the wind_direction validator a test that it is 1D
2) Add a wind_speed validator to test that it is 1D and the same length as wind_direction
(Note TI already had a validator/test comparing lengths)

These changes are made in flow_field.py so they will catch the same errors in reinitialize as above, but also different lengths of values in an input yaml

Finally:

3) Adds test to confirm that making these errors in reinitialize raises value errors.
4) Fixing regression test that didn't conform to this standard

flow_field.py
tests/

### DIFF
--- a/floris/simulation/flow_field.py
+++ b/floris/simulation/flow_field.py
@@ -79,8 +79,30 @@ class FlowField(BaseClass):
 
     @wind_directions.validator
     def wind_directions_validator(self, instance: attrs.Attribute, value: NDArrayFloat) -> None:
+        # Check that the array is 1-dimensional
+        if self.wind_directions.ndim != 1:
+            raise ValueError(
+                "wind_directions must have 1-dimension"
+            )
+
         """Using the validator method to keep the `n_findex` attribute up to date."""
         self.n_findex = value.size
+
+    @wind_speeds.validator
+    def wind_speeds_validator(self, instance: attrs.Attribute, value: NDArrayFloat) -> None:
+
+        # Check that the array is 1-dimensional
+        if self.wind_speeds.ndim != 1:
+            raise ValueError(
+                "wind_directions must have 1-dimension"
+            )
+
+        """Confirm wind speeds and wind directions have the same lenght"""
+        if len(self.wind_directions) != len(self.wind_speeds):
+            raise ValueError(
+                f"wind_directions (length = {len(self.wind_directions)}) and "
+                f"wind_speeds (length = {len(self.wind_speeds)}) must have the same length"
+            )
 
     @heterogenous_inflow_config.validator
     def heterogenous_config_validator(self, instance: attrs.Attribute, value: dict | None) -> None:
--- a/floris/simulation/flow_field.py
+++ b/floris/simulation/flow_field.py
@@ -94,7 +94,7 @@ class FlowField(BaseClass):
         # Check that the array is 1-dimensional
         if self.wind_speeds.ndim != 1:
             raise ValueError(
-                "wind_directions must have 1-dimension"
+                "wind_speeds must have 1-dimension"
             )
 
         """Confirm wind speeds and wind directions have the same lenght"""
--- a/floris/simulation/flow_field.py
+++ b/floris/simulation/flow_field.py
@@ -71,6 +71,12 @@ class FlowField(BaseClass):
         self, instance: attrs.Attribute, value: NDArrayFloat
     ) -> None:
 
+        # Check that the array is 1-dimensional
+        if value.ndim != 1:
+            raise ValueError(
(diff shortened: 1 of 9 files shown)


paulf81:
@rafmudaf  and @misi9170 , just noticed this is causing some reg_tests to fail, this is maybe ok though, the code that fails includes lines like this:

    sample_inputs_fixture.floris["flow_field"]["wind_directions"] = [270.0, 360.0]
    sample_inputs_fixture.floris["flow_field"]["wind_speeds"] = [8.0]

This makes me think the validator should check for length 1, since this will get projected upwards right?

misi9170:
> @rafmudaf and @misi9170 , just noticed this is causing some reg_tests to fail, this is maybe ok though, the code that fails includes lines like this:
> 
> ```
> sample_inputs_fixture.floris["flow_field"]["wind_directions"] = [270.0, 360.0]
> sample_inputs_fixture.floris["flow_field"]["wind_speeds"] = [8.0]
> ```
> 
> This makes me think the validator should check for length 1, since this will get projected upwards right?

tests/floris_interface_test.py
```diff
@@ -15,6 +15,22 @@ def test_read_yaml():
     assert isinstance(fi, FlorisInterface)
 
 
+def test_reinitialize():
```
rafmudaf, line 18: Let's test this on the FlowField class directly rather than indirectly through the FlorisInterface

rafmudaf:
For what it's worth, considering the format for array-inputs (see https://github.com/NREL/floris/discussions/794) is a scope creep. I think it would be reasonable to maintain the current design here, and develop the design for the array inputs further in the discussion thread.

The current design is that wind directions and wind speeds must be the same size, and other array-inputs can be an array of length 1 or `n_findex`.

paulf81:
> The current design is that wind directions and wind speeds must be the same size, and other array-inputs can be an array of length 1 or `n_findex`.

The problem I hit is that the reg tests don't currently follow this convention, so I would have to change them

floris/simulation/flow_field.py
```diff
@@ -79,9 +79,31 @@ def turbulence_intensities_validator(
...
         """Using the validator method to keep the `n_findex` attribute up to date."""
         self.n_findex = value.size
 
+    @wind_speeds.validator
+    def wind_speeds_validator(self, instance: attrs.Attribute, value: NDArrayFloat) -> None:
+
+        # Check that the array is 1-dimensional
+        if self.wind_speeds.ndim != 1:
+            raise ValueError(
+                "wind_directions must have 1-dimension"
```
rafmudaf, line 97: copy / paste error

rafmudaf:
Based on the design we had chosen for v4, the bug is in both the regression tests and the FlowField validators. The regression tests should also be fixed in all of the [rotation tests](https://github.com/paulf81/floris/blob/feature/validate_ws_wd/tests/reg_tests/jensen_jimenez_regression_test.py#L226).

paulf81:
> Based on the design we had chosen for v4, the bug is in both the regression tests and the FlowField validators. The regression tests should also be fixed in all of the [rotation tests](https://github.com/paulf81/floris/blob/feature/validate_ws_wd/tests/reg_tests/jensen_jimenez_regression_test.py#L226).

OK I can fix both right now

tests/floris_interface_test.py
paulf81: Will do!

paulf81:
> > the bug is in both the regression tests and the FlowField validators.
> 
> Maybe it's more accurate to say the bug is in the FlowField validators and you've fixed those. Then, an error was raised, as it should have been, because the regression tests were incorrectly calling the API.

agreed

paulf81:
hi @rafmudaf I reconfigured the pull request around this discussion, is this better?

rafmudaf (approved):

misi9170 (approved):

Issue #804, opened 2024-02-14
Use correct turbine names in plots


aclerc:
I was just discussing this with @paulf81 and he encouraged me to add an issue :)
Please can any plots showing turbine names use the correct turbine names. For example in the farm shown below the turbines are actually named T02, T04, T05, etc. but T00, T01, T02, etc. is displayed. If the correct turbine names are not available using the name "T00" will probably never be right, so if you have to fall back to generated names I would at start with T01
![image](https://github.com/NREL/floris/assets/9821851/0496ef07-5ec9-46c1-8e49-74f97f6a8a60)

aclerc:
sorry looks like my screenshot is too fuzzy to read but @paulf81 can add the original image if needed!

rafmudaf:
@aclerc can you share more information so that we can recreate this issue? A simplified analysis script including the functions you used for plotting would be helpful. Also, which version of FLORIS are you using?

paulf81:
Believe this issue is resolved by #805

PR #805, opened 2024-02-15
Collect all layout visualization tools in FLORIS

paulf81:
# Collect all layout visualization tools in FLORIS

Presently, tools for interrogating/plotting the layouts existing, sometimes in duplicate, across FLORIS and FLASC, and even within FLORIS, tools were duplicated between visualizations.py and layout_functions.py.  The purpose of this pull request are pulling all the tools into one module (layout_functions.py) in FLORIS, and then importing into FLASC.  This pull request also seeks to resolve Issue #804 by ensuring all functions accept turbine_names.

In this PR (and companion pull request in flasc (https://github.com/NREL/flasc/pull/174)), functions are all moved to layout_functions.py, and their form and usage is standardized.  

A new example (to be re-organized later (23_layout_visualizations is introduced).  This example demonstrates the preferred usage of importing the layout_functions module as lf, and calling each function via lf.xxx(). 

 In each case, either a pyplot axes is passed in or else a new one is created and returned.  Also in each case, a FlorisInterface object is passed in to get the locations of the turbines.

The functions in layout_functions.py are intended to each implement one plotting action and can be called sequentially to add layers to the plot.

- `plot_turbine_points`: Plot turbines as points on axis
- `plot_turbine_labels`: Label the turbines, in default case given FLORIS' 0-index value as a label, but can also accept provided turbine names
- `plot_turbines_rotors`: Indicate the turbine rotors (including yaw) on a plot.  This is usually most useful for indicating turbine position onto flow visualizaitons
- `plot_waking_directions` Draw labeled lines in-between each turbine and label the distance (in D) and wind directions in which the turbines wake each other.
- `shade_region`: Add a shaded region to an existing plot, can be useful to identify specific areas, or boundary boxes or phases/farms
- `plot_farm_terrain`: Use the hub-height information in FLORIS to make a contour plot of elevation

An example plotting script would be:

```
import floris.tools.layout_functions as lf

fig, ax = plt.subplots()
lf.plot_turbine_points(fi, ax=ax)
lf.plot_turbine_labels(fi, ax=ax)
lf.plot_waking_directions(fi, ax=ax)
plt.show()
```

A small test module is started (layout_functions_test.py).

Finally, as shown in example above, we removed calls to plt.show() from within functions inside layout_functions.py and adopt a convention that this call should be made by user.

## Related issue
#804


paulf81:
@rafmudaf  and @misi9170  this is now ready for your review

rafmudaf:
Heads up the tests are still failing

paulf81:
> Heads up the tests are still failing

Its the examples, that is my bad, I'll get it

paulf81:
@rafmudaf I fixed the examples, sorry for that!

floris/tools/__init__.py
```diff
@@ -25,9 +25,8 @@
 from .floris_interface import FlorisInterface
 from .parallel_computing_interface import ParallelComputingInterface
 from .uncertainty_interface import UncertaintyInterface
-from .visualization import (
+from .visualization import (  # plot_turbines_with_fi,
```
rafmudaf, line 28: Remove?

floris/tools/layout_functions.py
rafmudaf, line 1: "layout_function" seems vague for a module that only contains functions to plot the farm layouts. Consider renaming to reflect that it's specifically for plotting layouts.

floris/tools/layout_functions.py
```diff
@@ -1,376 +1,323 @@
...
-):
+def plot_turbine_points(
+    fi: FlorisInterface,
+    ax: plt.Axes = None,
+    turbine_indices: List[int] = None,
+    plotting_dict: Dict[str, Any] = {},
+) -> plt.Axes:
     """
-    Make a plot which shows the turbine locations, and important wakes.
+    Plots turbine layout from a FlorisInterface object.
```
rafmudaf, line 27: Might be good to note that this assumed wind direction is 270

floris/tools/layout_functions.py
```diff
@@ -1,376 +1,323 @@
...
+    ax: plt.Axes = None,
+    turbine_names: List[str] = None,
+    turbine_indices: List[int] = None,
+    label_offset: float = None,
+    show_bbox: bool = False,
+    bbox_dict: Dict[str, Any] = {},
+    plotting_dict: Dict[str, Any] = {},
+) -> plt.Axes:
+    """
+    Adds turbine labels to a turbine layout plot.
```
rafmudaf, line 96: Does `plot_turbine_points` need to have been run on the given ax prior to using this?

floris/tools/layout_functions.py
```diff
@@ -1,376 +1,323 @@
...
-        wakeAngle = wakeAngle + 360.0
-    if wakeAngle > 360:
-        wakeAngle = wakeAngle - 360.0
+    dx = x_j - x_i
+    dy = y_j - y_i
+
+    angle_rad = np.arctan2(dy, dx)
+    angle_deg = 270 - np.rad2deg(angle_rad)
+
+    # Adjust for "from" direction (add 180 degrees) and wrap within 0-360
```
rafmudaf, line 265: I'm not sure whether this comments matches the code below. Is this implicit in the %?

floris/tools/layout_functions.py
```diff
@@ -397,18 +344,232 @@ def put_label(i, ax):
...
+            is provided. If None, defaults to the first turbine's rotor diameter.
+        limit_dist_D (float, optional): Maximum distance between turbines (in rotor diameters)
+            to plot waking lines. Defaults to None (no limit).
+        limit_dist_m (float, optional): Maximum distance (in meters) between turbines to plot
+            waking lines. Overrides `limit_dist_D` if provided. Defaults to None (no limit).
+        limit_num (int, optional):  Limits the number of waking lines plotted from each turbine
+            to the `limit_num` closest neighbors. Defaults to None (no limit).
+        wake_label_size (int, optional): Font size for labels showing wake distance and direction.
+            Defaults to 7.
+
```
rafmudaf, line 395: The other docstrings nicely add the exceptions raised, so it might be good to add that here.

floris/tools/layout_functions.py
```diff
@@ -397,18 +344,232 @@ def put_label(i, ax):
...
+    Args:
+        fi (FlorisInterface): Instantiated FlorisInterface object containing layout data.
+        fig (plt.Figure):  The figure object where the plot will be created.
+        ax (plt.Axes): The axes object where the plot will be drawn.
+    """
+    hub_heights = fi.floris.farm.hub_heights.flatten()
+    cntr = ax.tricontourf(fi.layout_x, fi.layout_y, hub_heights, levels=14, cmap="RdBu_r")
+
+    fig.colorbar(
```
rafmudaf, line 505: This is the only function in this module that operates on the figure. Consider refactoring for consistency with the others, if possible.

floris/tools/visualization.py
```diff
@@ -25,126 +25,6 @@
 from floris.utilities import rotate_coordinates_rel_west, wind_delta
 
 
-def show_plots():
```
rafmudaf, line 28: This function allows users to avoid having to import pyplot only to run `plt.show()`. In basic use cases for the plotting methods in floris, the plotting functions create and work with the matplotlib data that they need, so from the outside (the user perspective) it's often not important which plotting library is ultimately being used. I suggest to keep this function to preserve this support.

```python
from floris import FlorisInterface
import floris.visualization as wakeviz

fi = FlorisInterface(...)
plane = fi.calculate_horizontal_plane(...)
wakeviz.visualize_calculate_plane(plane)

wakeviz.show_plots()
```

floris/tools/layout_functions.py
paulf81: Should've been on the line above, moving now

floris/tools/layout_functions.py
paulf81: It's possible, but colorbar is a finnicky one, I think that is why it's attached to the figure, an apparant alternative would be:

```
from matplotlib.colorbar import Colorbar

# Assuming cntr is a contour plot
cbar = Colorbar(ax=ax, mappable=cntr, orientation='vertical', ticks=np.linspace(
            np.min(hub_heights) - 10.0,
            np.max(hub_heights) + 10.0,
            15,
        ))
cbar.set_label("Terrain-corrected hub height (m)")
```

floris/tools/layout_functions.py
paulf81: Something has to have been plotted or I think maplotlib doesn't no exactly how to scale.  Plotting rotors, or plotting wake lines also fine

floris/tools/layout_functions.py
paulf81: Done!

floris/tools/layout_functions.py
paulf81: This doesn't plot any flow so I don't think that is assumed?

floris/tools/__init__.py
paulf81: Done!

floris/tools/layout_functions.py
paulf81: layout_plots.py?  layout_vis.py?  layout_plotting.py?

floris/tools/visualization.py
paulf81: That's true, @misi9170 and I were thinking it was against our normal expectations.  plt.show is a matplotlib operation, like saving a figure to png or adding to a pdf, it's just one of several options open to the user.  It seemed more simple to us to end FLORIS' responsibilities to the object after we return it.  I guess this point we might discuss a little more,

misi9170: Interesting point @rafmudaf, I hadn't thought of that. I guess I feel that most users will be importing matplotlib.pyplot anyway---in examples 02, 03, and 17 that Paul changed here, `matplotlib.pyplot as plt` was always imported; and I don't think we have any intention of supporting any other plotting library anyway?

To me, it might occasionally save one import line for users that just want to see a figure as generated from our tools with no change; but the cost is that it (to me) creates confusion, since I'm so used to needing to call `plt.show()` to see plots, that it's not immediately clear what this function would do (when I see it called, I assume it will do something more than just call `plt.show()`, because it's use case for saving the matplotlib import was not obvious to me).

Of course, there's no need for users to call this function if we do keep it---they can just go ahead and call `plt.show()`. If we want to keep it, I'd suggest:
- Changing the function name to `show()`, rather than `show_plots()`, so that it exactly mimics the underlying matplotlib call
- Add a comment to the version on visualization.py explaining why we are directly wrapping matplotlib's `show()` with our own
- Add a comment to at least one of the examples (possibly 02) where we call `wakeviz.show_plots()` explaining that users can optionally directly call `matplotlib.pyplot.show()` if they'd prefer

misi9170:
I've added a small, indirectly related bugfix to this PR, introduced in #823: I noticed in a place where `fi.reinitialize()` was accidentally still being called (with argument `layout_x`, among others) that the error raised was
`TypeError: FlorisInterface.reinitialize() got an unexpected keyword argument 'layout_x'`,
which is pretty confusing when in fact, `reinitialize()` has been deprecated; the solution I have was to allow the empty `reinitialize()` function (as well as `calculate_wake()`) to accept (and ignore) arbitrary keyword arguments. Now, the intended 
`NotImplementedError: The reinitialize method has been removed. Please use the set method. See https://nrel.github.io/floris/upgrade_guides/v3_to_v4.html for more information.`
is raised when `reinitilize()` is called with arguments.

floris/tools/layout_functions.py
misi9170: How about:
- visualization.py -> flow_visualization.py
- layout_functions.py -> layout_visualization.py

floris/tools/layout_functions.py
misi9170: I've updated this so that `fig` is just grabbed from the passed `ax` using `ax.get_figure()`. That removes the need for `fig` to be passed separately. I've also added to the 23 example to check that works and see no difference in output between the original implementation (which received `fig`) and the change (which no longer requires/accepts `fig` as an argument).

misi9170 (approved):
Approving this now, but I do think @rafmudaf is right that the name of the file should be something more clearly to do with plotting/visualization than layout_functions.py. I've added a suggestion on that above.

floris/tools/layout_functions.py
paulf81: Nice suggestion @misi9170 , I added this in

floris/tools/layout_functions.py
paulf81: Thank you @misi9170 !

floris/tools/visualization.py
paulf81: This works for me @misi9170 if it works for you @rafmudaf , I can implement if you green light

examples/03_making_adjustments.py
```diff
@@ -2,7 +2,8 @@
 import matplotlib.pyplot as plt
 import numpy as np
 
-import floris.tools.visualization as wakeviz
+import floris.tools.flow_visualization as wakeviz
+import floris.tools.layout_visualization as lf
```
rafmudaf, line 6: Are we ready to commit to "lf" for "layout_visualization"? This decision could be left until the examples are reworked, but it's also relevant to this pull request.

floris/tools/visualization.py
paulf81: Done!

examples/03_making_adjustments.py
paulf81: I'm switching to layoutviz to match wakeviz

floris/tools/layout_functions.py
rafmudaf: Note this line is now at https://github.com/NREL/floris/pull/805/files#diff-a12fc22ccce42f15bb7383a32268cb73c30c13ed05556260633ed06fed0c5631R20.
Good point @paulf81 that this doesn't specifically depend on the wind direction. It was initially confusing for me because we currently have a form of the flow visualizations that rotate the turbines and keep the wind coming from the left instead of keeping the turbines fixed and oriented with the cardinal directions (left: west, up: north, etc). So, considering how this function relates to that one, it might be worth mentioning that this one plots the farm layout with the cardinal directions oriented with left/right/up/down. It's redundant, but this nuance got me so it'll probably get someone else.

examples/03_making_adjustments.py
misi9170: while we're at it... do you want to change `wakeviz` to `flowviz`? Or, change flow_visualization.py to wake_visualization.py (I know I was the one to suggest flow_visualization.py, sorry about that). Just seems they should probably be consistent (`wakeviz` with wake_visualization.py or `flowviz` with flow_visualization.py)

floris/tools/layout_functions.py
paulf81: Is this right though?  This plot from example 23, the wind is 300 degrees, and flow and layout visualizations line up:

image

examples/03_making_adjustments.py
paulf81: ha!  I thought of that, decided maybe it's fine but I bet I'll hear about it and indeed!  I'll fix it now

floris/tools/layout_functions.py
rafmudaf: Ah you're right. I was thinking about #578, but I found https://github.com/NREL/floris/pull/578#issuecomment-1534131964 that notes the plotting was changed to always keep the layout fixed and instead rotate the wind direction. Disregard this review comment, and thanks for helping me get there!

paulf81: Pleasure!

floris/tools/layout_visualization.py
```diff
@@ -0,0 +1,579 @@
...
+                **plotting_dict,
+            )
+
+    # Plot labels and aesthetics
+    ax.axis("equal")
+
+    return ax
+
+
+def plot_turbines_rotors(
```
rafmudaf, line 190: The other functions are "plot_turbine_xx" - singular turbine - so that might be good to be consistent

tests/layout_visualization_test.py
```diff
@@ -0,0 +1,33 @@
...
+
+    # Turbine 0 wakes Turbine 1 at 0 degrees
+    assert np.isclose(layoutviz.get_wake_direction(0, 1, 0, 0), 0.0)
+
+    # Winds from the south
+    assert np.isclose(layoutviz.get_wake_direction(0, -1, 0, 0), 180.0)
+
+def test_plotting_functions():
+
+    fi = FlorisInterface(configuration=YAML_INPUT)
```
rafmudaf, line 27: Include test for `plot_turbines_rotors`

floris/tools/layout_visualization.py
rafmudaf: Done

tests/layout_visualization_test.py
rafmudaf: I've added this test, but it would also be good to test `plot_farm_terrain`.
Also, I think I know the answer to this, but is there a better test available? It's good to run these functions, but as is this is only testing that the inputs are of the correct general data type.

rafmudaf (approved):

tests/layout_visualization_test.py
paulf81: Yeah, this was a puzzle for me, since they end in a plot, my main thought was just to make sure the functions can run all the way through without erroring.  I wasn't sure how to confirm if the plot is "right"

paulf81: I can add the test for plot_farm_terrain, then merge you think if I do it tonight?

rafmudaf: Yep sounds good to me.

paulf81: Ok great, done and merging!

Issue #806, opened 2024-02-16
[Bug] reference_wind_height does not update when turbine_type is changed


misi9170:
I'll start out by saying this isn't a true bug, but it's something that hung me up for quite a while today, so worth possibly discussing...

If the input floris .yaml file specifies `reference_wind_height: -1`, then `fi.floris.flow_field.reference_wind_height` is set to the hub height of whatever turbine is specified in the `turbine_type` field of the input yaml. If the turbine_type is then _changed_ with a `reinitialize()` call (without also explicitly setting the `reference_wind_height` argument of `reinitialize()`), the `reference_wind_height` stays at it's old value. 

For example, 
```
from floris.tools import FlorisInterface

fi = FlorisInterface("inputs/gch.yaml")
print(fi.floris.flow_field.reference_wind_height)

fi.reinitialize(turbine_type=["iea_15mw"]*len(fi.layout_x))
print(fi.floris.flow_field.reference_wind_height)
```
produces 
```
90.0
90.0
```
(which is the hub height of the NREL_5MW model that is specified by gch.yaml).

Again, this is not strictly wrong---however, since `reference_wind_height: -1` is specified in the original yaml, it's rather confusing (to me, at least!) that the reference wind height does not get updated when the `turbine_type` changes. 

Actually changing the `fi.floris.flow_field.reference_wind_height` to a new value when `turbine_type` is changed may not be straightforward; however, I propose that a warning be raised if `reinitialize()` is called with`turbine_type` specified and `reference_wind_height` _not_ specified. This warning should also state what the `reference_wind_height` is being left as.

### Background
This came up for me because I was wanting to quickly compare the outputs of various wake models for a farm of iea_15MW turbines in an analysis script. To do so, I copied and pasted turbopark.yaml, cc.yaml, gch.yaml, and emgauss.yaml to my project folder, and then created a script that:
1. Instantiated each of the models in turn as their own `FlorisInterface` object
2. Using `reintialize()`, changed the layout of the farm to my desired layout and the turbine types to `"iea_15mw"`, as well as setting my wind conditions of interest
3. Ran `calculate_wake()` and looked at the power output.

[...]

PR #809, opened 2024-02-19
[BUGFIX] Raise warning if turbine_type passed without updating reference_wind_height

misi9170:
Addresses #806 by raising a warning if `FlorisInterface.reinitialize()` is called with `turbine_type` specified and `reference_wind_height` unspecified. 

The following code
```

# Initialization uses NREL 5MW model with 90m hub height
print(fi.floris.flow_field.reference_wind_height)

# Changing the turbine type without specifying reference_wind_height
fi.reinitialize(turbine_type=["iea_15mw"]*len(fi.layout_x))
print(fi.floris.flow_field.reference_wind_height)

# Changing the turbine type and specifying reference_wind_height
fi.reinitialize(turbine_type=["iea_15mw"]*len(fi.layout_x), reference_wind_height=150.2)
print(fi.floris.flow_field.reference_wind_height)
```

now raises the following warning for the line `fi.reinitialize(turbine_type=["iea_15mw"]*len(fi.layout_x))`:

```
floris.tools.floris_interface.FlorisInterface WARNING turbine_type has been changed without specifying a new reference_wind_height. reference_wind_height remains 90.00 m.
```

This "bug" is not specified to v4, and a similar change could be made for v3. However, since it is simply raising a warning, that may not be necessary.

### DIFF
--- a/floris/tools/floris_interface.py
+++ b/floris/tools/floris_interface.py
@@ -298,6 +298,13 @@ class FlorisInterface(LoggingManager):
         if layout_y is not None:
             farm_dict["layout_y"] = layout_y
         if turbine_type is not None:
+            if reference_wind_height is None:
+                self.logger.warning(
+                    "turbine_type has been changed without specifying a new "
+                    +"reference_wind_height. reference_wind_height remains {0:.2f} m.".format(
+                        flow_field_dict["reference_wind_height"]
+                    )
+                )
             farm_dict["turbine_type"] = turbine_type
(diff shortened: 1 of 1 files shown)


misi9170:
I just saw this method that I had not noticed before (it's called in example 18): https://github.com/NREL/floris/blob/v4/floris/tools/floris_interface.py#L94. I could update the warning to tell the user that they may want to call this method.

misi9170:
Closing as superseded by #1000

PR #814, opened 2024-02-20
Enable heterogeneity to work with new 4D framework and wind_data objects

paulf81:
# Enable heterogeneity to work with new 4D framework and wind_data objects

Previously heterogeneity was specified by a dictionary which defined N points, and then a speed up table `speed_multipliers` of size MxN where M was aligned to the number of wind directions.  Following the move to 4D M was now aligned to n_findex.  This is in principle fine, however it could be awkward to define a table that is defined at every findex value.

This draft pull request introduces a new optional variable to pass to fi.reinitialize `heterogenous_inflow_config_by_wd`, that includes a similar dictionary, however a new value is included `het_wd` that is now matched to the number of rows in `speed_multipliers`.  If `heterogenous_inflow_config_by_wd` is passed into reinitialize a n_findex sized `heterogenous_inflow_config` is constructed using the member function `get_heterogenous_inflow_config`.  Essentially for every findex value, the rows of `speed_multipliers` are determined by finding the nearest defined wind direction in `heterogenous_inflow_config_by_wd`.  I believe this functionality will be more user-friendly then the current findex-based version (although that is maintained inside the 'core' code) and the previous version, since the defined wind directions are now up to the user.

Several tests are added to floris_interface_test.py to make sure the nearest wind direction is working properly, including cases involving wrapping around 0 degrees.  Finally example 16b is expanded to make use of this capability.

Still to go:

 - [ ] Does this approach sound reasonable?
 - [ ] Should we allow wind speed het in a similar way?
 - [ ] Using with TimeSeries should be simple whether using n_findex version of this by_wd version
 - [ ] With WindRose, it could be good to insist on the by_wd appraoch as aligning the findex can be ambiguous

## Related issue
No issue but there is a project card in sprint

floris_interface.py


paulf81:
@rafmudaf  and @misi9170 , merged in v4 so this is ready for an initial review, note still some questions above

misi9170:
@paulf81, to address your questions:
- Overall, I agree with the approach, but I don't like that `FlorisInterface` now takes both `heterogeneous_inflow_config` and `heterogeneous_inflow_config_by_wd`. An option would be to allow only the `heterogeneous_inflow_config` version if the user is passing to the `FlorisInterface` directly, but have the other option if they pass through a `wind_data` object. See below.
- I'm on the fence with wind speeds. It would be nice to allow users that flexibility; but if we're enabling heterogeneity by wind speed, we should probably also enable heterogeneity by TI. I'm OK with either 1) leaving with only wd, and possibly expanding functionality in future, or 2) enabling wd, ws, ti heterogeneity. I don't love the idea of enabling ws without ti
- agreed, but see below
- agreed, but see below

I would say that, when passing in a `wind_data` object (`TimeSeries` or `WindRose`), the heterogeneous flow map should be contained on the `wind_data` object and the direct keyword argument to `FlorisInterface.set()` should be ignored, in much the same way as other keyword arguments for `FlorisInterface.set()` are ignored (e.g. `wind_speeds`).

paulf81:
> @paulf81, to address your questions:
> 
> * Overall, I agree with the approach, but I don't like that `FlorisInterface` now takes both `heterogeneous_inflow_config` and `heterogeneous_inflow_config_by_wd`. An option would be to allow only the `heterogeneous_inflow_config` version if the user is passing to the `FlorisInterface` directly, but have the other option if they pass through a `wind_data` object. See below.
> * I'm on the fence with wind speeds. It would be nice to allow users that flexibility; but if we're enabling heterogeneity by wind speed, we should probably also enable heterogeneity by TI. I'm OK with either 1) leaving with only wd, and possibly expanding functionality in future, or 2) enabling wd, ws, ti heterogeneity. I don't love the idea of enabling ws without ti
> * agreed, but see below
> * agreed, but see below
> 
> I would say that, when passing in a `wind_data` object (`TimeSeries` or `WindRose`), the heterogeneous flow map should be contained on the `wind_data` object and the direct keyword argument to `FlorisInterface.set()` should be ignored, in much the same way as other keyword arguments for `FlorisInterface.set()` are ignored (e.g. `wind_speeds`).

Good comments, let's meet up some time and go through,

paulf81:
Note this pull request is being super-ceded by #831 and will be closed without merging

Issue #516, opened 2022-10-31
Add times series option to UncertaintyInterface


paulf81:
**Is your feature request related to a problem? Please describe.**
It would be useful to follow FlorisInterface and allow the ws/wd to be passed in in time series form.

**Describe the solution you'd like**
UncertaintyInterface to include time_series flag just with IO/functionality matched to FlorisInterface

misi9170:
@paulf81 I believe the intended behavior is now incorporated in the `UncertainFlorisModel` in v4, and this issue can be closed?

paulf81:
Oh yeah, this is done, good catch

PR #821, opened 2024-02-22
Update uncertainty interface to 4d, new API

paulf81:
# Update uncertainty interface to 4d, new API

This pull request updates UncertaintyInterface to the v4 4d structure and the new set/run api currently in the fi_api branch.  In doing so it resolves Issue #516  since uncertainty will now work equivalently to time series or wrapped wind rose.

The UncertaintyInterface, unlike previous implementatons, inherits the FlorisInterface and overloads certain functions to implement wind direction uncertainty (in practice this was the only type used).  As before it uses gaussian averages of perturbations, but makes greater use of rounding and numpy reshaping and summing methods to achieve this result.  It reimplements set (leaving run as is), as well as get_turbine_powers, get_farm_power, get_farm_aep to return this uncertain result.  Parameters of uncertainty are passed in at initialization.

Pull request includes unit tests of some of the underlying functions, and an update of example 20_ to show the new usage.

## Related issue
#516 

uncertainty_interface.py


examples/20_calculate_farm_power_with_uncertainty.py
```diff
@@ -16,41 +15,106 @@
 """
 
 # Instantiate FLORIS using either the GCH or CC model
-fi = FlorisInterface("inputs/gch.yaml") # GCH model
-fi_unc = UncertaintyInterface("inputs/gch.yaml") # Add uncertainty with default settings
+fi = FlorisInterface("inputs/gch.yaml")  # GCH model
+fi_unc_3 = UncertaintyInterface(
+    "inputs/gch.yaml", verbose=True, wd_std=3
+)  # Add uncertainty with default settings
```
bayc, line 21: Is this default settings with the different `wd_std`?

examples/20_calculate_farm_power_with_uncertainty.py
```diff
@@ -16,41 +15,106 @@
...
+fi_unc_5 = UncertaintyInterface(
+    "inputs/gch.yaml", verbose=True, wd_std=5
+)  # Add uncertainty with default settings
```
bayc, line 24: Same here, default?

floris/tools/uncertainty_interface.py
```diff
@@ -1,460 +1,217 @@
...
+        configuration: dict | str | Path,
+        wd_resolution=1.0,  # Degree
+        ws_resolution=1.0,  # m/s
+        ti_resolution=0.01,
+        yaw_resolution=1.0,  # Degree
+        power_setpoint_resolution=100,  # kW
+        wd_std=3.0,
+        wd_sample_points=None,
+        verbose=False,
     ):
```
bayc, line 29: Need to add a docstring?

floris/tools/uncertainty_interface.py
```diff
@@ -1,460 +1,217 @@
...
+        self.n_sample_points = len(self.wd_sample_points)
+
+        # Get the weights
+        self.weights = self._get_weights(self.wd_std, self.wd_sample_points)
+
+
+        # Call base init function
+        super().__init__(configuration)  # Call the parent's __init__
+
+    def set(self,
```
bayc, line 54: Should bump self down. Odd that ruff didn't catch that.

floris/tools/uncertainty_interface.py
```diff
@@ -1,460 +1,217 @@
...
+        return result
 
-        Returns:
-            NDArrayFloat: Power production of all turbines in the wind farm.
-            This array has the shape (num_wind_directions, num_wind_speeds,
-            num_turbines).
+    def get_farm_power(
+        self,
+        turbine_weights=None,
+    ):
```
bayc, line 212: Could be all on one line.

floris/tools/uncertainty_interface.py
```diff
@@ -468,46 +225,39 @@ def get_farm_power(self, turbine_weights=None):
...
 
         if turbine_weights is None:
             # Default to equal weighing of all turbines when turbine_weights is None
             turbine_weights = np.ones(
                 (
-                    self.floris.flow_field.n_wind_directions,
-                    self.floris.flow_field.n_wind_speeds,
-                    self.floris.farm.n_turbines
+                    self.n_unexpanded,
+                    self.floris.farm.n_turbines,
```
bayc, line 240: Could be on one line.

floris/tools/uncertainty_interface.py
```diff
@@ -468,46 +225,39 @@ def get_farm_power(self, turbine_weights=None):
...
         elif len(np.shape(turbine_weights)) == 1:
             # Deal with situation when 1D array is provided
             turbine_weights = np.tile(
                 turbine_weights,
-                (
-                    self.floris.flow_field.n_wind_directions,
-                    self.floris.flow_field.n_wind_speeds,
-                    1
-                )
+                (self.n_unexpanded, 1),
```
bayc, line 247: Could be on one line.

examples/20_calculate_farm_power_with_uncertainty.py
paulf81: 3 is the default, but passing in to be explicit

paulf81: Ah I see what you mean, fixing

examples/20_calculate_farm_power_with_uncertainty.py
paulf81: fixed

floris/tools/uncertainty_interface.py
paulf81: fixed

floris/tools/uncertainty_interface.py
paulf81: somehow ruff set it this way?

floris/tools/uncertainty_interface.py
paulf81: another ruff decision...

floris/tools/uncertainty_interface.py
paulf81: ruff again...

floris/tools/uncertainty_interface.py
```diff
@@ -1,337 +1,103 @@
...
+            If not provided, defaults to [-2 * wd_std, -1 * wd_std, 0, wd_std, 2 * wd_std].
+        verbose (bool, optional): Verbosity flag for printing messages. Defaults to False.
+
+    """
+    def __init__(
         self,
-        unc_options=None,
-        unc_pmfs=None,
-        fix_yaw_in_relative_frame=None
+        configuration: dict | str | Path,
```
rafmudaf, line 46: Are `str` and `Path` supported?

floris/tools/uncertainty_interface.py
```diff
@@ -1,337 +1,103 @@
...
+        # Get the weights
+        self.weights = self._get_weights(self.wd_std, self.wd_sample_points)
 
-        if fix_yaw_in_relative_frame is not None:
-            self.fix_yaw_in_relative_frame = bool(fix_yaw_in_relative_frame)
+        # Call base init function
+        super().__init__(configuration)  # Call the parent's __init__
 
-    def reinitialize(
+    def set(
```
rafmudaf, line 79: A docstring would be helpful

floris/tools/uncertainty_interface.py
rafmudaf: It's the trailing comma. If you have a comma at the end of a sequence, ruff is configured to split that. If you delete the comma after "n_turbines", you can make it one line and satisfy ruff.

floris/tools/uncertainty_interface.py
```diff
@@ -1,337 +1,103 @@
...
-        unc_options=None,
-        unc_pmfs=None,
-        fix_yaw_in_relative_frame=False,
-    ):
-        """A wrapper around the nominal floris_interface class that adds
-        uncertainty to the floris evaluations. One can specify a probability
-        distribution function (pdf) for the ambient wind direction. Unless
-        the exact pdf is specified manually using the option 'unc_pmfs', a
-        Gaussian probability distribution function will be assumed.
+class UncertaintyInterface(FlorisInterface):
```
rafmudaf, line 16: Reading through this class, I wonder if inheritance is the right architecture over composition. Instead of managing the FlorisInterface through `self` you could create it in `__init__` here and call `.set()` as needed. This avoids raising NotImplementErrors at the end and doesn't necessarily tie this class to the other. Note that the two functions used from the super-class are `.set` and `.get_turbine_power`, and they're both overloaded. This is just to say that the benefit of inheritance might be outweighed by any complexities that come up when theres a change to `FlorisInterface` that shouldn't impact `UncertaintyInterface` (i.e. a new function to do an unrelated task will require adding the NotImplemented-version here if you want to maintain the pattern).

rafmudaf: This would require adding a `UncertaintyInterface.run()` function, fwiw

floris/tools/uncertainty_interface.py
paulf81: I'm not sure, I think I copied this from the type hint in FlorisInterface, and this is just getting passed downwards to there

floris/tools/uncertainty_interface.py
paulf81: Added

floris/tools/uncertainty_interface.py
paulf81: Added

floris/tools/uncertainty_interface.py
paulf81: We're also using the init() function from the super.  @misi9170 and I debated this at the beginning and thought the inheritance was slightly more elegant, but I think I was thinking if we did composition we'd have to pass in the "certain" fi into the uncertain, but I think your solution is really nice, if I understand it right, we declare the "certain" fi within the uncertain one and keep it around as a member to call on as needed.  Do I have that right?  I can code it up,

rafmudaf: Yea, I think you have it right. Here's a prototype of how this would connect to FlorisInterface with the current functionality:

```python
class UncertaintyInterface:

    def __init__(self, ...):
        self.floris_interface = FlorisInterface(...)

    def set(self, ...):
        self.floris_interface.set(...)
        self.set_uncertain()

    def run(self):
        self.floris_interface.run()

    def get_turbine_power(self):
        unique_turbine_powers = self.floris_interface.get_turbine_power()
        expanded...
        weights...
        return result
```

For what it's worth, I'm not sure if this is "right" either, but it's another option. What stood out is that usually a subclass is a superset of the parent-class and in this case it's a subset. That's more of a smell test than a rule, so I thought it would be worth considering composition.

paulf81: ok! last commit switches from inheritance to composition,

floris/tools/uncertainty_interface.py
```diff
@@ -1,337 +1,174 @@
...
-            FLORIS to historical data, in which a single measurement usually
-            represents a 10-minute average, and thus is often a mix of various
-            true wind directions. The inherent assumption then is that the turbine
-            perfectly tracks the wind direction changes within those 10 minutes.
-            Then, fix_yaw_in_relative_frame=False is typically used for robust
-            yaw angle optimization, in which we take into account that the turbine
-            often does not perfectly know the true wind direction, and that a
-            turbine often does not perfectly achieve its desired yaw angle offset.
-            Defaults to fix_yaw_in_relative_frame=False.
+        wd_resolution (float, optional): The resolution of wind direction, in degrees.
```
rafmudaf, line 32: Maybe not here, but in general we should pick either "wind_direction" or "wd" for wind direction named arguments.

rafmudaf: And same for the other related input arguments

floris/tools/uncertainty_interface.py
```diff
@@ -1,337 +1,174 @@
...
+                                    ws_resolution=self.ws_resolution,
+                                    ti_resolution=self.ti_resolution,
+                                    yaw_resolution=self.yaw_resolution,
+                                    power_setpoint_resolution=self.power_setpoint_resolution,
+                                    wd_std=self.wd_std,
+                                    wd_sample_points=self.wd_sample_points,
+                                    verbose=self.verbose)
+
+
+    def set(
```
rafmudaf, line 117: I wonder if we can pass `**args` directly through to `FlorisInterface.set()` and point to `FlorisInterface.set()` in the docstring here. It's a nitpick, but it seems like a lot of repeated code that might get out of sync easily.

rafmudaf:
Thanks for addressing my comments @paulf81. Overall, this is looking pretty good to me.

floris/tools/uncertainty_interface.py
paulf81: Yeah, I would be happy either way, with maybe a slight preference toward wd_ for succinctness in this case

paulf81:
@rafmudaf  and @misi9170 , just to close a loop here, @misi9170 added some tests of applying yaw angles to the uncertainty interface that threw some unexpected errors.  I tracked down two seperate issues:

1) One is a problem of passing lists, or integers, to yaw_angles in the set function to FlorisInterface.  I made a new pull request for handling that one (#828 )

2) I realized that having one saved internal floris_interface was a bug, because in some places I expected it to hold the un_expanded conditions and in others the expanded.  This works fine if you call set once, but falls apart after that.  However, the solution of holding two seperate fis, with the expanded created every time you run _set_uncertain, I think clears everything up.  This change is now in the code.

paulf81:
Hi @rafmudaf  and @misi9170 , we talked today about possibly removing the un-expanded FLORIS interface, I started making the changes tonight (storing the dict instead) but there did seem to be some trade-offs, mainly I think in the set function the operation would be to declare a new FLORIS object, and again only store the dict.  I think it clearly could be done, but if you wouldn't mind having a quick look and concurring it should be done (versus the current version's memory usage but clearer syntax), let me know what you think and I'll make the change

misi9170:
> Hi @rafmudaf and @misi9170 , we talked today about possibly removing the un-expanded FLORIS interface, I started making the changes tonight (storing the dict instead) but there did seem to be some trade-offs, mainly I think in the set function the operation would be to declare a new FLORIS object, and again only store the dict. I think it clearly could be done, but if you wouldn't mind having a quick look and concurring it should be done (versus the current version's memory usage but clearer syntax), let me know what you think and I'll make the change

I guess an option would be to create a dict direction rather than instantiating a new FLORIS object and then exporting it to a dict. The code for that (building a `floris_dict`) is in `FlorisInterface._reinitialize()`. It doesn't seem great to copy that code into `UncertaintyInterface`, but we could consider having a standalone function that builds a `floris_dict` that both `FlorisInterface._reinitialize()` and `UncertaintyInterface.set()` call. But, honestly, that might be overdoing it; I'd say that, for now, it's probably fine to just go ahead and instantiate a FLORIS object (and allocate memory to it); and then convert it to a dict with `.as_dict()` and (presumably) release the memory again. It's not super elegant, but at least it's easy with the methods we already have in place.

paulf81:
> > Hi @rafmudaf and @misi9170 , we talked today about possibly removing the un-expanded FLORIS interface, I started making the changes tonight (storing the dict instead) but there did seem to be some trade-offs, mainly I think in the set function the operation would be to declare a new FLORIS object, and again only store the dict. I think it clearly could be done, but if you wouldn't mind having a quick look and concurring it should be done (versus the current version's memory usage but clearer syntax), let me know what you think and I'll make the change
> 
> I guess an option would be to create a dict direction rather than instantiating a new FLORIS object and then exporting it to a dict. The code for that (building a `floris_dict`) is in `FlorisInterface._reinitialize()`. It doesn't seem great to copy that code into `UncertaintyInterface`, but we could consider having a standalone function that builds a `floris_dict` that both `FlorisInterface._reinitialize()` and `UncertaintyInterface.set()` call. But, honestly, that might be overdoing it; I'd say that, for now, it's probably fine to just go ahead and instantiate a FLORIS object (and allocate memory to it); and then convert it to a dict with `.as_dict()` and (presumably) release the memory again. It's not super elegant, but at least it's easy with the methods we already have in place.

Ok @misi9170 , I made the change, now the unexpanded FI is only created as needed and  only the dictionary version is saved.  One place we might just want to double check ourselves, does the dictionary save the values of yaw_angles/power_setpoints?  If not repeated calls to uncertainity_interface.set() might work a little differently in this version than the previous where we hold the entire fi in some cases, but I'm not sure,

paulf81:
ok @rafmudaf  and @misi9170 , tonight it:
 1) Reverted this back to using a seperate FlorisInterface (rather than dict, a little wasteful but safer)
 2) I merged v4 back in following the merge into v4 of the change to layout_visualization.
 3) Made sure I had addressed previous hanging comments

Think it's now ready,

paulf81:
Note the method outlined here will be adapted into #831 and this pull request will be closed

misi9170:
> Note the method outlined here will be adapted into #831 and this pull request will be closed

I don't follow?

misi9170 (approved):

paulf81:
@rafmudaf good to merge this one?

Issue #779, opened 2024-01-17
Drive Optimization off of WindData data structures


paulf81:
# Drive Optimization off of WindData (TimeSeries, WindRose) data structures

Convert optimization functions to be driven off of WindData structures, this issue will wait to be addressed until the optimizations have been updated to v4 without this change.   This will enable simple switching between aep-based optimization, and for example revenue-based optimization, since both of these will be held in WindData Structures

PR #822, opened 2024-02-22
Drive layout optimizations using WindData objects

paulf81:
# Drive layout optimizations using WindData objects

This pull request reorients the layout optimization classes to accept WindData objects (WindRose or TimeSeries), in place of freq tables, in order to facilitate allowing change of optimization from maximizing AEP to maximizing other quantities such as revenue.  Changing to passing the WindData objects facilitates this because these objects hold the value per bin (WindRose) or as a time series (TimeSeries).  

(Note this will appear to be many changes at first since built off of #819 ) but once that PR is merged, this will look much simpler

## Related issue
#779


floris/tools/optimization/layout_optimization/layout_optimization_boundary_grid.py
```diff
@@ -8,6 +8,12 @@
     Polygon,
 )
 
+from floris.tools import (
+    TimeSeries,
+    WindRose,
+)
+from floris.tools.wind_data import WindDataBase
+
```
ejsimley, line 16: It doesn't seem like these WindData classes are used in this file. Are they needed?

floris/tools/optimization/layout_optimization/layout_optimization_base.py
```diff
@@ -30,12 +35,19 @@ def __init__(self, fi, boundaries, min_dist=None, freq=None, enable_geometric_ya
...
+        # If wind_data is None, make a simple time series object off the fi
+        if wind_data is None:
+            wind_data = TimeSeries(wind_directions=fi.wind_directions,
+                                wind_speeds=fi.wind_speeds,
+                                turbulence_intensities=fi.turbulence_intensities)
         else:
-            self.freq = freq
+            if ((not isinstance(wind_data, WindDataBase)) and
+                 (not isinstance(wind_data, TimeSeries)) and
+                 (not isinstance(wind_data, WindRose))):
```
ejsimley, line 46: I think we only need to check that wind_data is an instance of WindDataBase

paulf81: Fixed!

floris/tools/optimization/layout_optimization/layout_optimization_boundary_grid.py
paulf81: You're right, deleting

floris/tools/optimization/layout_optimization/layout_optimization_pyoptsparse.py
```diff
@@ -4,6 +4,12 @@
 from scipy.spatial.distance import cdist
 from shapely.geometry import Point
 
+from floris.tools import (
+    TimeSeries,
+    WindRose,
+)
+from floris.tools.wind_data import WindDataBase
+
```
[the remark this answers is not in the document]
paulf81: You're right, deleting

ejsimley (approved):
Thanks @paulf81, the WindData additions all look good here.

bayc:
I merged `v4` in to make the changes easier to review.

floris/tools/__init__.py
```diff
@@ -19,7 +19,7 @@
     '__name__', '__package__', '__path__', '__spec__', 'cut_plane',
     'floris_interface',
     'layout_functions', 'optimization', 'plotting', 'power_rose',
-    'visualization']
+    'visualization', 'wind_rose']
```
rafmudaf, line 22: This file was removed #775 so it's no longer needed here. I removed this specific line in #819, so this might be a result of merge conflicts.

bayc: Ah, gotcha, I misread the merge conflict current vs incoming. I was merging v4 in and it was listed in the incoming, or so I thought. I will remove it.

bayc: @rafmudaf Yep, I got the two switched up, just went back to that commit and re-merged. Good catch!

rafmudaf: Got it, thanks for fixing that

paulf81:
Merged v4 back in again

floris/tools/optimization/layout_optimization/layout_optimization_base.py
```diff
@@ -3,15 +3,17 @@
...
 from floris.tools.optimization.yaw_optimization.yaw_optimizer_geometric import (
     YawOptimizationGeometric,
 )
+from floris.tools.wind_data import WindDataBase
 
 from ....logging_manager import LoggingManager
 
 
 class LayoutOptimization(LoggingManager):
-    def __init__(self, fi, boundaries, min_dist=None, freq=None, enable_geometric_yaw=False):
```
misi9170, line 14: Does this mean that, if the user would like to provide a weighting to different wind speed bins; they must provide a wind_data object (whereas if they want to use uniform weighting, just supplying fi will suffice)?

Could we instead leave the freq option, which will be ignored if wind_data is None, but which otherwise will be used in the creation of the TimeSeries object in LayoutOptimization.__init__()?

Same goes for the other layout optimization classes, obviously

paulf81: @misi9170 let's chat about this one tomorrow,

paulf81: Ok good point @misi9170 , as discussed changing wind_data to required input to avoid supplying a use case no one is likely to want but could get accidentally

PR #823, opened 2024-02-23
Move FlorisInterface .reinitialize() / .calculate_wake() to .set() / .run()

rafmudaf:
## Adopt .set() / .run() flow for FlorisInterface workflows

Currently, the typical flow for using FlorisInterface is to build a workflow around the `.reinitialize()` and `.calculate_wake()` functions. #801 suggests to add an additional step between these two for setting turbine operating conditions. In the subsequent discussion (see https://github.com/NREL/floris/discussions/801#discussioncomment-8468688), the idea to refactor this workflow resulted in the following suggestion:
```python
fi = FlorisInterface.from_dict(...)
fi.set(
    wind_directions=...,
    wind_speeds=...,
    any_other_input=...
)
fi.run()
```
This pull request implements the suggested workflow by adding `FlorisInterface.set()` and `FlorisInterface.run()` and removing `FlorisInterface.reinitialize()` and `FlorisInterface.calculate_wake()`.

The `FlorisInterface.set()` function allows for setting any of the supported inputs or operation settings individually or in batch:
```python
# These two are equivalent
fi.set(
    wind_directions=...,
    wind_speeds=...,
    any_other_input=...
)

fi.set(wind_directions=...)
fi.set(wind_speeds=...)
fi.set(any_other_input=...)
```
For any parameter that is part of the input (defined in the input file or input dictionary to `FlorisInterface` and `Floris`), the `.set()` function creates a new `Floris` object with the setting changed. This matches the functionality of `.reinitialize()`. Therefore, changes should be batched into one `.set()` when possible to save the overhead of creating the new objects. In either case, the parameters that aren't provided as arguments to the function are not changed in Floris including the operation settings.

If either `FlorisInterface.calculate_wake()` or `FlorisInterface.reinitialize()` are used from v3, an error is raised pointing the user to the upgrade guide.

RIP [calculate_wake()](https://github.com/NREL/floris/blob/v0.1.0/floris/flow_field.py#L142) 💀

## Related issue
Related to https://github.com/NREL/floris/discussions/801

`FlorisInterface` and any script or function that uses it.


rafmudaf:
Consider this flow for multiple calculations as implemented in #823:
```python
f = Floris.from_dict(...)

# Get power with non-zero yaw
f.set(wind_directions=..., wind_speeds=..., yaw_angles=...)
f.run()
yawed_power = f.get_farm_AEP()

# Get power with no yaw
f.reset_operation()    # Sets all operating set-points to their default
f.run()
baseline_power = f.get_farm_AEP()
```

This start to look like Floris is retaining memory, and this is the same paradigm we intended to remove in the shift from v2 to v3. In fact, `f.reset_operation()` resets the operating set-points and creates a new Floris object, so the memory is not retained. However, the user-facing API still implies that we're operating on a single instance.

@bayc I'm wondering whether you read this the same way, and if you have any feedback on this given our experience with v2 and the shift to v3. For my part, this flow feels like a backward step, but I think I'm stuck too much on the previous API's to give this a fair evaluation.

floris/tools/floris_interface.py
```diff
@@ -108,175 +111,134 @@ def copy(self):
...
-    def calculate_wake(
+    def set(
         self,
+        wind_speeds: list[float] | NDArrayFloat | None = None,
+        wind_directions: list[float] | NDArrayFloat | None = None,
+        wind_shear: float | None = None,
+        wind_veer: float | None = None,
+        reference_wind_height: float | None = None,
+        turbulence_intensities: list[float] | NDArrayFloat | None = None,
+        # turbulence_kinetic_energy=None,
```
paulf81, line 122: Should we go ahead and delete?

floris/tools/floris_interface.py
```diff
@@ -108,175 +111,134 @@ def copy(self):
...
-                )
-
-            # Must have first dimension = n_turbines
-            if disable_turbines.shape[1] != self.floris.farm.n_turbines:
-                raise ValueError(
-                    f"disable_turbines has a size of {disable_turbines.shape[1]} "
-                    f"in the 1th dimension, must be equal to "
-                    f"n_turbines={self.floris.farm.n_turbines}"
-                )
+    def reset_operation(self):
```
paulf81, line 203: If this function only calls another function do we still need it?  Is it a public private split?

floris/tools/floris_interface.py
```diff
@@ -477,6 +524,8 @@ def calculate_horizontal_plane(
         wd=None,
         ws=None,
         yaw_angles=None,
+        power_septoints=None,
+        disable_turbines=None,
```
paulf81, line 528: note the docstring args for this function is incomplete

floris/tools/floris_interface.py
```diff
@@ -560,6 +610,8 @@ def calculate_cross_plane(
         wd=None,
         ws=None,
         yaw_angles=None,
+        power_setpoints=None,
+        disable_turbines=None,
```
paulf81, line 614: docstring is incomplete

floris/tools/floris_interface.py
```diff
@@ -638,6 +692,8 @@ def calculate_y_plane(
...
```
paulf81, line 696: docstring is incomplete

floris/tools/parallel_computing_interface.py
```diff
@@ -279,7 +280,7 @@ def _postprocessing(self, output):
 
         return turbine_powers
 
-    def calculate_wake(self):
+    def calculate_wake(self): # TODO: Remove or update this function?
```
paulf81, line 283: Do we now know the resolution of this TODO item?

floris/tools/floris_interface.py
```diff
@@ -108,175 +111,134 @@ def copy(self):
...
-        self.floris.farm.power_setpoints = power_setpoints
-
         # Initialize solution space
         self.floris.initialize_domain()
 
         # Finalize values to user-supplied order
         self.floris.finalize()
 
-    def reinitialize(
+    def _reinitialize(
```
paulf81, line 241: Would it make sense for there to be functions like reininitialize() and calculate_wake() which solely exist to prompt users to read the v4 documentation?

floris/tools/floris_interface.py
rafmudaf: Good point. I'm not sure how to handle it. `_reinitialize` does set the operation set-points back to their defaults, but it's also very close to the v3 `reinitialize` so I think it's a bad idea to keep that as a public function if we want to move to a different workflow.

floris/tools/floris_interface.py
rafmudaf: Yes, that's a nice suggestion. I've added these two functions in https://github.com/NREL/floris/pull/823/commits/058930b053f2065469872bacb9f5ba69c2266b28.

paulf81:
@rafmudaf I noticed a test failing and traced it to a small issue in test_disabled_turbines and pushed up a fix

floris/tools/floris_interface.py
paulf81: I made a quick fix to both

floris/tools/floris_interface.py
paulf81: fixed

floris/tools/floris_interface.py
paulf81: I fixed both I think

paulf81:
These seems ready to merge from my point of view, only outstanding comment I see is a question on "TODO" item in the parellel_computing_interface, but that is really minor so I will approve

misi9170:
> @misi9170 It would be good to get your eyes here again when you have a chance. I think I've addressed the tests that we had discussed last week.

Thanks for setting those tests up @rafmudaf --- they look good to me.

PR #828, opened 2024-02-28
[Bugfix] Cast yaw angles to np.ndarray on set

paulf81:
# [Bugfix]  Fix yaw setting

In the course of some other work, I noticed that a few conditions that I would've expected to be ok throw an error:

- Passing a list to yaw_angles in fi.set
- If the values of yaw_angles are integers instead of floats

This is not so hard to fix but I wasn't positive the preferred solution (attrs?). So I thought I'd start just by adding the failing tests and then we can agree on how we want to fix it.

Other things to consider:

- [x] Does this also impact power setpoints? _No, tested_
- [x] Does this also impact disable turbines? _No, tested_
- [x] Does fixing this in FlorisInterface intrinsically fix the problem in Parallel and Uncertain interface? _It seems to, yes. I tried this on the #821 (implemented the fixed, and then altered the test to take yaw angles as a list), and it worked._

Note this issue is a little bit being worked-around in #821


rafmudaf:
See https://github.com/NREL/floris/blob/v4/floris/simulation/farm.py#L328

misi9170:
The fix was straightforward here---in `_set_operation()`, I've now wrapped the assignment of the `yaw_angles` onto `self.floris.farm.yaw_angles` in a `np.array()` statement. 

Incidentally, this is exactly what @rafmudaf warned me about---that reaching in and altering the attributes (rather than instantiating them properly, which would pass through `__attrs__`) can lead to unintended consequences. For now, this fix works; but we should possibly consider adding setter methods to `Farm` to avoid this happening again.

misi9170:
@paulf81 I can't add you as a reviewer, since you opened the PR, but please look through and see what you think.

paulf81:
This works great for me, thank you @misi9170 I "approve"

floris/tools/floris_interface.py
```diff
@@ -355,7 +355,7 @@ def _set_operation(
         """
         # Add operating conditions to the floris object
         if yaw_angles is not None:
-            self.floris.farm.yaw_angles = yaw_angles
+            self.floris.farm.yaw_angles = np.array(yaw_angles)
```
rafmudaf, line 358: Is this the only place where yaw angles are going to be set? If not, you might consider adding a setter method on `Farm` to do this always:

```python
class Farm:
    def set_yaw_angles(yaw_angles):
        # Do lots of checks
        ...
        # Cast to Numpy array
        self.yaw_angles = np.array(yaw_angles)
```

You could also use the `@property` decorator [the way we used to do](https://github.com/NREL/floris/blob/31fe1b69ff863f0a610aec5b22424382ec3cc933/floris/utilities.py#L96-L102), if you think maintaining the attribute-style use is important.

rafmudaf: Ah and I just read https://github.com/NREL/floris/pull/828#issuecomment-1979972559, sorry to say the same thing @misi9170

misi9170: No problem---I'm going to try to do this (use a setter method), it seems like a better approach

rafmudaf (approved):

Issue #791, opened 2024-02-03
Raise informative error if v3 intput yaml passed into v4


paulf81:
This issue is to catch the intention that before releasing v4 the following enhancement should be included, although it's final form will depend upon the final form of v4.  FLORIS v4 should raise an informative error when a v3 input yaml file is passed into v4.   Current issues to detect are:

- Cp values passed in in place of Power
- Tubulence Intensity instead of turbulence intensities

However recommend a full reassessment toward the end of the v4 development process.

Alternaitively or additionally, we can provide an automatic converter script from v3 to v4

misi9170:
Closed by #829

PR #829, opened 2024-03-03
Raise informative errors if v3 input files passed in

misi9170:
Addresses #791 

Checklist:
- [X] Raise informative errors if v3 floris input yaml used
- [ ] Raise informative errors if v3 turbine yaml used
- [ ] Provide converter for v3 floris input yaml
- [X] Provide conveter for v3 turbine yaml (see floris/tools/convert_turbine_v3_to_v4.py, introduced in #765)

### DIFF
--- a/floris/simulation/floris.py
+++ b/floris/simulation/floris.py
@@ -347,6 +347,7 @@ class Floris(BaseClass):
             Floris: The class object instance.
         """
         input_dict = load_yaml(Path(input_file_path).resolve())
+        check_input_file_for_v3_keys(input_dict)
         return Floris.from_dict(input_dict)
 
     def to_file(self, output_file_path: str) -> None:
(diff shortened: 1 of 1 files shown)


paulf81:
hi @misi9170 , I went through the code and thought it all looks great, thank you!  I approved.  I agree with your 2nd and 3rd comments that this is an ok way to leave things.  Then on pip install, there I don't know, maybe something to see in beta testing?

rafmudaf:
I like the spirit of this pull request, and I think it's a very helpful addition. Architecturally, this is mixing the physics model with the software infrastructure. I suggest to keep anything related to versioning, especially things you'll want to go back and change later as new versions come up, all in one place at the outer-most layer of the architecture. There was a place for this in the Floris class from the v2 to v3 shift (https://github.com/NREL/floris/blob/develop/floris/simulation/floris.py#L167). Essentially, when we read and deal with the Farm class in the future, we're not going to care about the input checking and messaging, so this adds more noise for our brains. Keeping it at the top allows us to forget about it until we need to go find it.

rafmudaf (approved):

misi9170:
To help address my third point above, I've now added an explanation that the user will need to do their own conversion of the multidimensional turbine csv file if they call convert_floris_input_v3_to_v4.py on a floris input file that specifies `multidim_cp_ct` as the velocity model. I'll do something similar for the turbine yaml converter.

rafmudaf:
Regarding https://github.com/NREL/floris/pull/829#issuecomment-1977010709, @misi9170 and I looked at the options for catching outdated turbine inputs at a higher level in the architecture of FLORIS. The ideal time is prior to initializing the `Floris` object so in `Floris.from_file` or `Floris.from_dict`.

`turbine_type` can be either a list of dictionaries or a list of strings for the turbine names. If it's a list of dictionaries, inspecting the data is straightforward. However, if it's a list of strings referring to turbine names, then the turbine data are not yet available for inspection. This is specifically a problem when using the external turbine library feature.

Rather than trying to hack something together that is more complicated, we agreed it's best to have a more robust check within the `Farm` class rather than at the `Floris` class level.

PR #830, opened 2024-03-05
Rename floris.simulation, floris.tools to floris.core, floris

rafmudaf:
# Rename FLORIS packages
This pull request restructures the packages within FLORIS.

## v3 Architecture

Currently, the software has the following high-level architecture:
```mermaid
classDiagram
    class tools {
        +FlorisInterface
    }
    class simulation {
        +Floris
    }
    class type_dec
    class utilities
    tools <-- type_dec
    simulation <-- type_dec
    tools <-- utilities
    simulation <-- utilities
    tools <-- simulation
```

The `floris.tools.FlorisInterface` class is the entry point for most use-cases, and it uses `floris.simulation` as a library for computing the wake calculation with the `Floris` class as the central data structure. This architecture has been around since at least v2, but it has some issues. One problem is that the word "floris" is used in these contexts:
- "FLORIS" as a acronym for the project
- "floris" as the Python package, this is what is installed from PyPI
- "floris" as the primary module within the `floris.simulation` package, i.e. `from floris.simulation.floris import Floris`. Note that the second "floris" in the import statement was hidden by the `__init__` imports at the top level
- "Floris" as the primary data structure in the `floris.simulation` package

`FlorisInterface` as the primary user entry was contextually relevant when FLORIS included interfaces to other simulators like SOWFA, but that functionality was left out in the upgrade to v3. The "Interface" portion remained in the name mostly as a relic but without the context of the other interfaces.

## v4 Module-level Redesign
As part of the transition to v4, the directory structure and package architecture of the FLORIS project is redesigned to the following:
```mermaid
classDiagram
    class Floris {
        +FlorisModel
    }
    class core {
        +Core
    }
    class type_dec
    class utilities
    Floris <-- type_dec
    core <-- type_dec
    Floris <-- utilities
    core <-- utilities
    Floris <-- core
```

Here, the user-facing component is `floris.FlorisModel`, and the former tools package is moved up to the top level.

## v3 / v4 API Compared
Here's a comparison of the old and new API's for a simple script:

```python

##### v3
import numpy as np

fi.reinitialize(wind_directions=..., wind_speeds=...)
fi.calculate_wake(yaw_angles=...)
turbine_powers = fi.get_turbine_powers()/1000.

horizontal_plane = fi.calculate_horizontal_plane(height=90.0)
wakeviz.visualize_cut_plane(
    horizontal_plane,
    ax=axarr[0],
    title="Initial setup",
    min_speed=MIN_WS,
    max_speed=MAX_WS
)

wakeviz.show_plots()


##### v4 at this PR

import numpy as np

import floris.visualization as wakeviz

fmodel.set(wind_directions=..., wind_speeds=..., yaw_angles=...)
fmodel.run()
turbine_powers = fmodel.get_turbine_powers()/1000.

horizontal_plane = fmodel.calculate_horizontal_plane(height=90.0)
wakeviz.visualize_cut_plane(
    horizontal_plane,
    ax=axarr[0],
    title="Initial setup",
    min_speed=MIN_WS,
    max_speed=MAX_WS
)

wakeviz.show_plots()
```

## Related issues, discussions, PRs
- https://github.com/NREL/floris/discussions/801
- https://github.com/NREL/floris/pull/823
- https://github.com/NREL/floris/discussions/825

This touches most files within the `floris/` directory. I've tried to keep this pull request scope to variable and class name changes, and I've tried to exclude other minor refactoring.

## Things to do
- [ ] First, I'd like feedback on the general design of the API. See examples updated in https://github.com/NREL/floris/commit/940b1b22894fc64ae97f0f901802576cc8c1f8ae. Agree to patterns for:
    - Variable name for the `FlorisModel` instance, currently `fmodel`
    - Naming conventions for `UncertaintyInterface` and `ParallelComputingInterface`, suggested `FlorisUncertaintyModel` / `umodel` and `FlorisParallelModel` / `pmodel`
    - Is `FlorisModel` as the main entry point good for everyone?
- [ ] Update API's in remaining packages and examples
    - Find / replace
        - FlorisInterface -> FlorisModel
        - fi = -> fmodel =
        - fi= -> fmodel=
        - fi. -> fmodel.
        - fi, -> fmodel,
        - fi: -> fmodel:
        - fmodel.floris -> fmodel.core
- [ ] Update docs:
    - README
    - Architecture & Design
    - Getting started


rafmudaf:
@paulf81 @misi9170 @bayc @ejsimley Please see this pull request, and any conceptual feedback would be helpful. There are a lot of files changed, but the content of the changes is relatively small. Rather than reviewing each file, it would be best to mostly focus on the following locations:
- The examples changed in https://github.com/NREL/floris/commit/940b1b22894fc64ae97f0f901802576cc8c1f8ae
- The [floris module containing FlorisModel](https://github.com/rafmudaf/floris/blob/package_name_change/floris/floris.py)
- The [visualization module](https://github.com/rafmudaf/floris/blob/package_name_change/floris/visualization.py) since that uses much of the components of the `FloridModel` class

misi9170:
@rafmudaf I agree with these changes, but I'm not a good data point because we already spoke about this. 

One thing though: is there any reason that the [floris module containing FlorisModel](https://github.com/rafmudaf/floris/blob/package_name_change/floris/floris.py) shouldn't just be called floris_model.py? that would stick to the convention of the file containing a class of the same name, and would also help get rid of this `from floris.floris ...` business

paulf81:
tests passing @rafmudaf

floris/flow_visualization.py
```diff
@@ -215,7 +215,7 @@ def visualize_heterogeneous_cut_plane(
     Args:
         cut_plane (:py:class:`~.tools.cut_plane.CutPlane`): 2D
             plane through wind plant.
-        fi (:py:class:`~.tools.floris_interface.FlorisInterface`): FlorisInterface object.
```
rafmudaf, line 218: @paulf81 just a heads up that this syntax, the `:py:class:` part, is good to keep since it links to the target class in the API docs. See [here](https://nrel.github.io/floris/_autosummary/floris.tools.visualization.visualize_heterogeneous_cut_plane.html#floris.tools.visualization.visualize_heterogeneous_cut_plane) for this particular example.

paulf81: That is good to know, it's a bit random right now the style of of the docstrings depending on by who and when they were written.  What if we made a separate issue/pull to go through and bring all the docstrings to a particular format?

paulf81:
hi @rafmudaf , one question for you, I also update the profiling folder, but noticed that timing.py and serial_vectorize.py are more out of date, but also that nothing is calling them, so maybe these are just to be deleted?

floris/flow_visualization.py
rafmudaf: No problem, I agree that the docstrings are very inconsistent. I have no issue with leaving them how they are, and I agree that updating the docstrings is outside the scope of this pull request.

rafmudaf:
> hi @rafmudaf , one question for you, I also update the profiling folder, but noticed that timing.py and serial_vectorize.py are more out of date, but also that nothing is calling them, so maybe these are just to be deleted?

Nothing in the `profiling/` directory is used by an automated system or many people outside of our group. Any of the scripts that we've developed in the past few years to profile FLORIS are dumped there. Unless there's a compelling reason to remove it, I'd prefer to update those scripts, or even leave them out of date, and keep them there as a "better than nothing" until there's a more comprehensive performance profiling system. There is some useful stuff there, but it's not well organized.

paulf81:
> Generally, it would be good to keep the scope of this pull request to changing the variable, class, and file names, and then we can do clean up etc in another pull request. This way, when we review the changes here, we can be on the lookout for only a specific set of things.

Sounds good, I think with that I've finished going through and updating all the class references and imports to the new style and all tests and examples are running

misi9170:
Working through the changes for `UncertaintyInterface` and `ParallelComputingInterface`; decided (with @paulf81) to go with:
- `UncertaintyInterface` -> `UncertainFlorisModel` (suggested instantiation `ufmodel`)
- 'ParallelComputingInterface` -> `ParallelFlorisModel` (suggested instantiation `pfmodel`)

misi9170:
Now addresses #834.

Note that support from `cut_in_wind_speed` and `cut_out_wind_speed` have been removed from `ParallelFlorisModel.get_farm_AEP()`. @paulf81 and I have discussed removing support from these from `FlorisModel.get_farm_AEP()` and `UncertainModel.get_farm_AEP()` also, but that will come in a separate pull request. These methods are counter to the paradigm that the underlying `FlorisModel` should contain only the data that is to be computed; instead, we can consider moving such functionality to the `WindData` objects.

PR #838, opened 2024-03-12
Add ti-mixing to EMG

paulf81:
# Add ti-mixing to EMG

This may not be an actual pull request, but may be convenient if it yields some changes.  The purpose of this is to look at making a version of EMG with ``atmospheric_ti_gain`` re-enabled.  To this end I started with a small pseudo-example that compares the power output of columns of a wind farm from the current default EMG to one which uses ``atmospheric_ti_gain``.

I had in mind a step by step criteria, ideally:

- [ ] The row averaged power ratios will be similar when 6% TI is used
- [ ] The row averaged power ratios will also be similar when IEC-reference TI per wind speed (provided using the Iref function in WindRose), or not, this might need something thinking
- [ ] The wakes will be deeper when TI < 6%
- [ ] The wakes will be shallower when TI > 6%

Note this is just a start but thought it a convenient way to collaborate.  Note also this is very messy until the namespace changes go into v4.  The only file worth checking is: examples/xx_add_ti_to_emg.py

Tagging: @misi9170  @jfrederik-nrel


paulf81:
Closing this one in favor of #841

PR #839, opened 2024-03-12
Add utilities for floris models

paulf81:
# Add utilities for floris models

(Note this will be easier to read when API changes merged to v4)

Previously, we added to flasc a set of utilities to make convenient the changing of FLORIS parameters within the nested dictionary structure.  This pull request moves them to FLORIS as they could have uses beyond SCADA comparisons, matching the pattern we set up in PR #805 

Originally I had planned to add these to utilities.py, but this creates a circular import since several of the functions accept and return a FlorisModel.  So created the files:

- floris_model_utils.py
- floris_model_utils_test.py

  Created tests to confirm correct behaviors.  Last item could be if we wanted to provide a convenience function that sets certain parameters without their parents in the case they appear only once in the dictionary.  But this a little bit violates the spirit of explicitness seen elsewhere.  

A function to print the dictionary in a nice, indented way is provided.


misi9170:
I'm going to wait to review until #830 is in to simplify the comparison?

paulf81:
It's funny @misi9170 , I think this should now be simpler following the merge of #830 but it hasn't refreshed or something?

rafmudaf:
@paulf81 I think v4 branch needs to be merged into this one.

paulf81:
> @paulf81 I think v4 branch needs to be merged into this one.

That's it, somehow I'm not managing to merge in v4, I keep hitting errors.  But since this PR is just adding two new files I just quit and started a new pull request off of a more current v4 to avoid the problems.  Closing this one in favor of #840

PR #842, opened 2024-03-13
Added helix model and examples

jfrederik-nrel:
- Added helix wake mixing model
- Added helix turbine power loss model
- Added examples to showcase new functionality

To do's:
- [x] Format code @misi9170 
- [x] Write tests
- [x] Format examples
- [x] Add default tuning parameters for different turbines


examples/inputs/emgauss.yaml
```diff
@@ -47,7 +47,8 @@ wake:
     velocity_model: empirical_gauss
 
   enable_secondary_steering: false
-  enable_yaw_added_recovery: true
+  enable_yaw_added_recovery: false
```
misi9170, line 50: Should `enable_yaw_added_recovery` be changed here to `false`? Ideally, we'd be able to leave this as `true`

examples/inputs/emgauss_iea_15mw.yaml
misi9170, line 1: I'd prefer not to have to add an extra input file---would it instead be possible to add the `helix_wake_exp` and `helix_wake_denominator` terms to examples/inputs/emgauss.yaml, load from that input instead in example 42, and then switch the `turbine_type` to the IEA 15MW?

examples/42_test_helix.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+"""
+Example to test out using helix wake mixing of upstream turbines.
+Helix wake mixing is turned on at turbine 1, off at turbines 2 to 4;
+Turbine 2 is in wake turbine 1, turbine 4 in wake of turbine 3.
+"""
+
+# Grab model of FLORIS and update to helix-enabled turbines
+fmodel = FlorisModel("inputs/emgauss_iea_15mw.yaml")
```
misi9170, line 30: Could we instead add the necessary Helix parameters to inputs/emgauss.yaml; load inputs/emgauss.yaml; and then `set` the `turbine_types` to `["iea_15mw"]`? That would save us adding another example input yaml to keep track of.

misi9170 (changes requested):
@jfrederik-nrel I've noted a handful of small updates that I think would clean this up. 

Then, a slightly bigger question that we have discussed before would be to change the names of `helix_amplitudes` and `helix_frequencies` to be more generic so that they can be used in future wake mixing models (or, if even more generic, perhaps other ways of operating the turbine that are not necessarily for the purpose of wake mixing).

A couple of options would be:
1. `wake_mixing_amplitude` and `wake_mixing_frequency`
2. `blade_pitch_amplitude` and `blade_pitch_frequency` 
3. `pitch_induced_mixing_amplitude` and `pitch_induced_mixing_frequency`

Others might have suggestions for this too!

Changing the names will be a little annoying as they appear in many places throughout Floris, but hopefully find and replace will be pretty effective once we decide on the names...

misi9170:
Another note is that we'll also need to figure out how to handle the helix setpoints in the `UncertainFlorisModel` and `ParallelFlorisModel`. I'd say that for now, it's appropriate for those to simply raise `NotImplementedError`s if helix setpoints are passed in (which I can implement), but it would be good to get others' opinions on that.

examples/42_test_helix.py
jfrederik-nrel: I get where you're coming from, but for now I've left it in for the following reason: the helix parameters are tuned specifically for 15MW simulations, so it feels weird to me to define these helix expansion parameters in a file defined for a turbine for which they aren't calibrated.

misi9170: That makes sense. Perhaps, then, we could instead call this file emgauss_helix.yaml, and have it specify the IEA_15MW as it's turbine (as it already does)? We could then put in the description (or just a comment at the top of the file) that the helix parameters specified are tuned to work with the IEA 15MW turbine.

floris/turbine_library/nrel_5MW.yaml
```diff
@@ -39,6 +39,12 @@ power_thrust_table:
   cosine_loss_exponent_tilt: 1.88
   # Cosine exponent for power loss due to yaw misalignment.
   cosine_loss_exponent_yaw: 1.88
+  # Helix parameters
+  helix_a: 1.802
+  helix_power_b: 4.568e-03
+  helix_power_c: 1.629e-10
```
rafmudaf, line 45: This is a very small number - is the helix model sensitive enough to changes at that order of magnitude? In other words, could this be 0?

tests/conftest.py
```diff
@@ -214,6 +214,11 @@ def __init__(self):
                 "cosine_loss_exponent_tilt": 1.88,
                 "ref_air_density": 1.225,
                 "ref_tilt": 5.0,
+                "helix_a": 1.809,
```
rafmudaf, line 217: These are different from the values in the NREL 5MW file. Should they be consistent?

floris/turbine_library/nrel_5MW.yaml
jfrederik-nrel: I agree that it's small, but it is multiplied by `base_powers`, which is in the order of 1e6. As a result, this sum is more or less the same magnitude as `helix_power_b`.

floris/core/solver.py
```diff
@@ -1290,6 +1335,17 @@ def empirical_gauss_solver(
                 downstream_distance_D[:,:,i],
                 model_manager.deflection_model.yaw_added_mixing_gain
             )
+        # if model_manager.enable_active_wake_mixing:
```
paulf81, line 1338: Delete commented code?

misi9170 (approved):

rafmudaf:
This pull request was squashed into https://github.com/NREL/floris/commit/9da4cbb39d06a7c7d2fb6af1c8f2c5b5192bacc2.

PR #843, opened 2024-03-13
Refactor examples

paulf81:
# Refactor examples for v4

This pull request refactors the examples around the v4 structures and style and in general cleans up the examples folder, which had become a sort of unsorted list of examples.  Specifically, 

- Examples are now organized into top level over-arching examples in the root examples folders, and then examples grouped by topic in subfolders
- The top level examples are completely redone to incrementally introduce FLORIS, in a way consistent with v4
- Each example now includes a self-describing docstring at the top of each
- Every example is refactored for v4 code and style, as well as overall improvements, style, spelling, ruff, etc

In addition to these changes, some supporting changes were made to the code

- Spelling errors of the word heterogeneous are fixed throughout the code
- A bug in 'mixed' operation model (which masked yaw_angles > 0, when it should've been != ) is fixed, which was identified in making one of the examples
- several additional properties were added to FlorisModel, adding to layout_x and layout_y, these are wind_directions, wind_speeds, turbulence_intensities, n_findex and n_turbines.  This change makes the examples more readable without the core.flowfield intermediates
- A bug in ParallelFlorisModel is fixed (sum -> nansum, fixed concatenating parallel results)

Finally, changes to CI include:

- CI running of examples update to include checking examples in subfolders

@misi9170  and @rafmudaf  and @ejsimley  I think this is now ready to review.  We talked about automatically generating html versions of the examples and so far this is proving difficult, so maybe that could still be done, but as part of another pull request


examples/002_visualizations.py
```diff
@@ -0,0 +1,95 @@
+"""Example 2: Visualizations
+
+This example demonstrates the use of the flow and layout visualizations in FLORIS.
```
ejsimley, line 3: Here it would be helpful to explain a little more what the example looks at. After the above sentence, something like: "First, an example wind farm layout is plotted, with the turbine names and the directions and distances between turbines shown. Next, the horizontal flow field at hub height is plotted for a single wind condition." I think this level of detail (a few sentences) would be good for all the examples.

ejsimley:
Overall, I think the organization of the examples folder looks good. And I like the standard documentation at the top of the files, though I think some could use more detail, as I mentioned in the comment. I'm wondering if the "examples_control_settings" folder should be renamed though. Is the point to demonstrate all the non-yaw or "beyond yaw" control examples there? Would a name like "examples_different_control_types" or just "examples_control_types" better describe the scope of those examples?

paulf81:
> Overall, I think the organization of the examples folder looks good. And I like the standard documentation at the top of the files, though I think some could use more detail, as I mentioned in the comment. I'm wondering if the "examples_control_settings" folder should be renamed though. Is the point to demonstrate all the non-yaw or "beyond yaw" control examples there? Would a name like "examples_different_control_types" or just "examples_control_types" better describe the scope of those examples?

Nice suggestion, making the change...

examples/004_set.py
misi9170, line 1: This example doesn't produce any outputs. I know that it isn't meant to, but that might be a bit confusing to users?

examples/006_get_farm_aep.py
```diff
@@ -0,0 +1,103 @@
...
+# the total number of wind directions and wind speed combinations
+print(f"Total number of wind direction and wind speed combination: {n_wd * n_ws}")
+print(f"Number of 0 frequency bins: {n_zeros}")
+print(f"n_findex: {fmodel.n_findex}")
+
+# Get the AEP
+aep = fmodel.get_farm_AEP()
+
+# Print the AEP
+print(f"AEP from wind rose: {aep/1E9:.3f} (GW-h)")
```
misi9170, line 91: We usually just format this as `GWh`, rather than `GW-h`? not a biggie, but could be good to be consistent

examples/examples_control_optimization/007_Optimize_Yaw_with_neighbor_farm.py
misi9170, line 1: capitals in file name are inconsistent?

examples/examples_control_types/002_setting_yaw_and_disabling.py
misi9170, line 1: I'm wondering whether this example could be combined with the 001 example? This one feels like more of a reality check for us internally than actually demonstrating intended usage?

examples/examples_control_types/000_derating_control.py
misi9170, line 1: numbering starts at 000 in this folder, which is inconsistent with other folders

examples/examples_uncertain/002_approx_floris.py
misi9170, line 1: Do we want to remove this, and the 003 example, for merging to v4?

examples/examples_visualizations/002_visualize_y_cut_plane.py
misi9170, line 1: Aren't examples 002--004 here covered by the 002 example in the main examples/ directory?

misi9170 (changes requested):
Hi @paulf81 , I have made a bunch of minor comments, but overall this looks good. 

I think the changes outside of the examples (`heterogeneous`, bugfixes to `ParallelFlorisModel`, `MixedOperationTurbine` bugfix, and properties available on `FlorisModel` look good).

examples/002_visualizations.py
paulf81: Done!  I'll make a pass through

examples/003_wind_data_objects.py
```diff
@@ -0,0 +1,217 @@
...
+# compute_zero_freq_occurrence parameter to True.
+wind_directions = np.array([200.0, 300.0])
+wind_speeds = np.array([5.0, 1.00])
+freq_table = np.array([[0.5, 0], [0.5, 0]])
+wind_rose = WindRose(
+    wind_directions=wind_directions, wind_speeds=wind_speeds, ti_table=0.06, freq_table=freq_table
+)
+fmodel.set(wind_data=wind_rose)
+
+print(
```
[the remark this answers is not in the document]
paulf81: Fixed

examples/004_set.py
paulf81: That is true but adding run, and get_power... after each set might clutter things a bit?  Not sure what's best here

examples/006_get_farm_aep.py
paulf81: Got it

examples/examples_control_types/000_derating_control.py
paulf81: fixed

examples/examples_control_types/002_setting_yaw_and_disabling.py
paulf81: Just flagging I fixed the numbering so this comment is now with the 002.  Here again I'm not sure, this is the only example to really use the "mixed" capability for now.  And then also, if I add it back to 002, I think it gets a little rambling, but, could do it I guess.  We can discuss more

examples/examples_uncertain/002_approx_floris.py
paulf81: Yes of course, that's my, deleting now...

examples/examples_visualizations/002_visualize_y_cut_plane.py
paulf81: Technically, yes, there a little more blocked out, but it's true not strictly necessary, but maybe convenient to have such a single-purpose example?  Also up for discussion...

examples/002_visualizations.py
paulf81: @ejsimley I went through the examples and tried to beef up the docstrings at the top of each

PR #848, opened 2024-03-18
Add CSV reader for WindRose

paulf81:
# Add CSV reader for WindRose

Working through the examples made me realize that a convenient method for reading a wind rose given as a CSV file (something that has to be done in many examples connected to AEP) leads to a lot of code clutter in examples.  This pull request aims to add this feature as a semi-member function of WindRose, that returns a WindRose.  Included in the pull request is a test of functionality using a minimal csv input file.

I think this is 90% ready but two questions I'd like feedback on:

1) If the CSV file is incomplete (missing possible combinations of wd/ws), the value of the freq table is clear (0) but the ti_table is ambiguous.  If the user prefers a fixed TI, could just back-fill this.  But if these are provided, ?  Right now going for np.nan, but not sure.

2) Is this type of function, a member function that doesn't accept "self" ok?  I wasn't sure where to put this function.  It's nicely organized under the WindRose class, but it doesn't make sense to have to first instantiate a dummy wind_rose object just to run this function, so thought this solution ok, but wasn't sure if I was breaking a taboo @rafmudaf ?


Bartdoekemeijer:
Maybe side-tracking but maybe not: would you like to support .tab files in FLORIS? .tab files are the common file format for wind roses internally here. I have a function that reads a .tab file and pours it into the FLORIS v3 wind rose format. It would be incredibly helpful on my end.

If so, then perhaps it's worth considering if you want to centralize I/O scripts anywhere in FLORIS. Could be a nice opportunity to move FLORIS further towards industry application. This is also in line with adding compatibility with .WRG wind resource files. Additionally, I also have a script to read WAsP .wtg turbine files and pour them into a `turbine_type`  dictionary. Happy to share these publicly and make them a part of FLORIS.

misi9170:
@paulf81 , regarding your second question, I would suggest putting a `@staticmethod` decorator on the new method, see https://www.geeksforgeeks.org/class-method-vs-static-method-python/

misi9170:
Just to tag it, this PR will enable cleaning up of some of the examples that will take place in #843

misi9170 (approved):

floris/wind_data.py
```diff
@@ -538,6 +538,90 @@ def plot_ti_over_ws(
...
+        wd_step = unique_wd[1] - unique_wd[0]
+        ws_step = unique_ws[1] - unique_ws[0]
+
+        # Now use TimeSeries to create a wind rose
+        time_series = TimeSeries(wind_directions, wind_speeds, turbulence_intensities)
+
+        # Now build a new wind rose using the new steps
+        return time_series.to_wind_rose(
+            wd_step=wd_step, ws_step=ws_step, bin_weights=freq_values
+        )
```
ejsimley, line 624: If a TI column is provided, and there are multiple TI values per wind speed/wind direction bin, then this could return a `WindTIRose`, which has ws, wd, ti as the frequency table dimensions. What do you think about adding that option?

paulf81: Sure I can do that

paulf81: Actually, I realized that adding it as a static member of that class is simpler

paulf81:
@rafmudaf @ejsimley @misi9170 ok to merge this afternoon?

floris/wind_data.py
ejsimley: good call

floris/wind_data.py
```diff
@@ -901,6 +985,87 @@ def plot_ti_over_ws(
...
+        is that the wind speed, wind direction combination is given for each row in the
+        CSV file. The wind speed, wind direction, are
+        given in separate columns, and the frequency of occurrence of each combination
+        is given in a separate column. The frequency column is optional, and if not
+        provided, uniform frequency of all bins is assumed.
+
+        The value of ti_col_or_value can be either a string or a float. If it is a string,
+        it is assumed to be the name of the column in the CSV file that contains the
+        turbulence intensity values. If it is a float, it is assumed to be a constant
+        turbulence intensity value for all wind speed and direction combinations.
```
ejsimley, line 1007: The above summary part of the doctoring should be updated for the WindTIRose version

ejsimley (approved):
After the doctstring summary for the WindTIRose version gets updated, this'll be good to merge.

floris/wind_data.py
paulf81: Fixed, thank you

PR #844, opened 2024-03-14
Add functions to return turbine and farm power in wind rose form

paulf81:
# Add functions to return turbine and farm power in wind rose form

This small pull request adds to functions to return the turbine and farm power in the shape (num_wd, num_ws, n_turbines), and relevent testing.  These functions may not be needed but could be useful when driving FLORIS off of WindRose wind_data objects.  Note these functions are not needed when computing AEP.

PR #849, opened 2024-03-20
Save WindData onto FlorisModel and simplify post-`run()` calls

misi9170:
This pull request simplifies calls to methods that extract the power after a `FlorisModel` has been `run()`. To do so, any `WindData` object that is passed in to `FlorisModel.set()` is saved onto the `FlorisModel`. Setting of the `FlorisModel._wind_data` attribute can only be achieved using `set()`, and a getter method is added to expose the hidden `FlorisModel._wind_data`.

These changes enable the following calls to be simplified considerably, and allow the output to match the "shape" of the `WindData` object, if one was used:
- `get_turbine_powers()`
- `get_farm_power()`
- `get_farm_AEP()`

A few important changes are also implemented:
- `get_farm_AEP()` is now simply a wrapper around a new method, `get_expected_farm_power()`. The `get_expected_farm_power()` is then clearer. We have also discussed raising an error if `get_farm_AEP()` is called _without_ a `WindRose` data type on `FlorisModel._wind_data`---@paulf81, what do you think, should `get_farm_AEP()` raise an error if called without a `WindRose`?
- None of the `get_...power` methods call `run()`. To use these methods, the user _must_ call `run()` first. This clarifies the paradigm, which used to differ between different methods (`get_farm_AEP()` would call `run()`, whereas `get_turbine_powers()` and `get_farm_power()` would not).
- The above means that the `no_wake` flag can be removed form the the `get_farm_AEP()` method---if the user would like a version without wakes, they will now execute `run_no_wake()` instead of `run()` prior to calling the `get_...power` methods.

### Other notes
- Saving the `WindData` object onto the `FlorisModel` essentially supersedes #822 
- the `cut_in_wind_speed` and `cut_out_wind_speed` arguments to `get_farm_AEP()` have been removed. These were inconsistent with the paradigm that all findices on the `FlorisModel` will be run. This also mirrors the removal of these terms on the `ParallelComputingInterface` (#830, although that removal is not complete). If users would like to remove certain wind conditions from AEP calculations, they will need to `set()` the conditions they intend to include.
- Supersedes and was motivated by #844 
- `get_farm_AEP()` now allows users to specify how many hours there are in a year (defaults to 365 * 24=8760, but some users may like to use 8766 to account for leap years)

### To do
- [x] Update `UncertainFlorisModel` to track these changes.
- [x] Consider switching to warning, rather than error, if layout optimization is run without a `WindData` object available
- [x] Consider raising error or warning if `get_farm_AEP()` is called without a `WindData` object available


misi9170:
@rafmudaf , before you review fully, are you happy with the main idea of this PR?
If so, @paulf81 , perhaps you could map these changes onto `UncertainFlorisModel` (which would be a good way of getting to know the changes I've made here...)

floris/floris_model.py
```diff
@@ -256,30 +263,31 @@ def _reinitialize(
...
                     "wind_speeds, turbulence_intensities or "
                     "heterogenous_inflow_config as this is redundant"
                 )
-            (
-                wind_directions,
-                wind_speeds,
-                turbulence_intensities,
-                heterogenous_inflow_config,
-            ) = wind_data.unpack_for_reinitialize()
+            elif self.wind_data is not None:
```
ejsimley, line 279: I think self.wind_data should be self._wind_data throughout floris_model

floris/floris_model.py
```diff
@@ -973,98 +1064,36 @@ def get_farm_AEP(
...
-            self.set(
-                wind_speeds=wind_speeds_subset,
-                wind_directions=wind_directions_subset,
-                turbulence_intensities=turbulence_intensities_subset,
-            )
-            if no_wake:
-                self.run_no_wake()
+        if freq is None:
+            if self.wind_data is None:
+                freq = np.array([1.0])
```
ejsimley, line 1073: to return the mean farm power, should this be freq = np.array([1.0/self.core.flow_field.n_findex])?

floris/floris_model.py
misi9170: I'm actually not sure about this---might need @rafmudaf 's input. My logic is that anywhere that the `wind_data` is modified (i.e. when it is set), it will set directly onto `self._wind_data`, but anywhere it is just called or used then it can use the getter version `self.wind_data`. But perhaps `self.wind_data` should be _only_ for external access, and all internal access (like the case you've pointed out here) should be `self._wind_data`.

ejsimley:
Thanks @misi9170. I think this is a good way to simplify getting farm power, expected power, AEP and saving/accessing the wind rose information in the floris model.

paulf81:
hi @misi9170 , one thought is you might discard the changes to the examples because they will only make the eventual merge to refactor_examples harder, if you leave the core changes in place I can update all the relavent examples in refactor_examples

misi9170:
> hi @misi9170 , one thought is you might discard the changes to the examples because they will only make the eventual merge to refactor_examples harder, if you leave the core changes in place I can update all the relavent examples in refactor_examples

That's true... but the checks will fail and the examples on the v4 branch won't run in the meantime. Is that a problem?

paulf81:
Maybe just delete those examples then?  I think the merge will be easier, I have everything I need in the refactor branch,

Paul

From: misi9170 ***@***.***>
Reply-To: NREL/floris ***@***.***>
Date: Wednesday, March 20, 2024 at 1:00 PM
To: NREL/floris ***@***.***>
Cc: Paul Fleming ***@***.***>, Mention ***@***.***>
Subject: Re: [NREL/floris] Save WindData onto FlorisModel and simplify post-`run()` calls (PR #849)


CAUTION: This email originated from outside of NREL. Do not click links or open attachments unless you recognize the sender and know the content is safe.

hi @misi9170<https://github.com/misi9170> , one thought is you might discard the changes to the examples because they will only make the eventual merge to refactor_examples harder, if you leave the core changes in place I can update all the relavent examples in refactor_examples

That's true... but the checks will fail and the examples on the v4 branch won't run in the meantime. Is that a problem?

misi9170:
Hi @paulf81, to be honest, I'd rather keep them in to keep a record of the changes necessary for the examples to run that will be tied to this PR's merge commit. Assuming this PR gets merged, when you go to merge v4 back into the refactor_examples branch at some point, you could just keep all current changes to the example files (and then make updates to the examples as necessary). I don't think that should be too much of a headache, but I'm certainly happy to help deal with the merge conflict if it's messier than I'm imagining!

paulf81:
> Hi @paulf81, to be honest, I'd rather keep them in to keep a record of the changes necessary for the examples to run that will be tied to this PR's merge commit. Assuming this PR gets merged, when you go to merge v4 back into the refactor_examples branch at some point, you could just keep all "current" changes (i.e. the versions on the refactor_example branch) to the example files (and then make updates to the examples as necessary). I don't think that should be too much of a headache, but I'm certainly happy to help deal with the merge conflict if it's messier than I'm imagining!

Ok, yeah, I get that.  Let's do it that way then, thanks!

floris/floris_model.py
misi9170: Confirmed with @rafmudaf that the current implementation is appropriate (no strong opinion either way, but this way if people read through the code and want to replicate parts, they'll see that they should be accessing the getter version

misi9170:
For future reference, note a small new capability that tests if a warning is logged (e.g [on this file](https://github.com/NREL/floris/blob/e56f6a11009c2759f4d1ebe07b7c1ef3798ad3b4/tests/layout_optimization_integration_test.py))

misi9170:
@aclerc, just wanted to tag you here so that you see this, since this change is partly in response to your feedback :)

PR #850, opened 2024-03-20
Allow to fix yaw to nominal wind direction

paulf81:
# Add fix_yaw_to_nominal_direction

This pull request re-includes the option in the UncertainFlorisModel to fix the yaws into the frame of the nominal wind direction.  Currently for example, if yaw angle is 0, and wind direction is 270, uncertainty is computed over a range of wind direction while holding yaw = 0.  This pull request adds the flag fix_yaw_to_nominal_direction, which changes the computations such that the yaw is modified to always be 0 relative to the nominal 270 degrees (so it would be +3 when checking 267 for nominal 270)

Specific changes:

- Incorporates the changes and bugfixes to UncertainFlorisModel already in #843 
- Adds the fix_yaw_to_nominal_direction functionality
- Adds tests for the new functionality

### DIFF
--- a/floris/uncertain_floris_model.py
+++ b/floris/uncertain_floris_model.py
@@ -11,7 +11,7 @@ from floris.type_dec import (
     NDArrayBool,
     NDArrayFloat,
 )
-from floris.utilities import wrap_360
+from floris.utilities import wrap_180
 from floris.wind_data import WindDataBase
 
 
@@ -610,9 +610,11 @@ class UncertainFlorisModel(LoggingManager):
             # If fix_yaw_to_nominal_direction is True, set the yaw angle to relative
             # to the nominal wind direction
             if fix_yaw_to_nominal_direction:
-                output_array[start_idx:end_idx, 3 : 3 + n_turbines] = (
-                    output_array[start_idx:end_idx, 3 : 3 + n_turbines] - wd_sample_points[i]
-                ) % 360
+
+                # Wrap between -180 and 180
+                output_array[start_idx:end_idx, 3 : 3 + n_turbines] = wrap_180(
+                    output_array[start_idx:end_idx, 3 : 3 + n_turbines] + wd_sample_points[i]
+                )
 
         return output_array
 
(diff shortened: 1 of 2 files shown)


floris/uncertain_floris_model.py
```diff
@@ -29,20 +29,27 @@ class UncertainFlorisModel(LoggingManager):
...
         power_setpoint_resolution (int, optional): The resolution of power setpoints, in kW.
             Defaults to 100.
         wd_std (float, optional): The standard deviation of wind direction. Defaults to 3.0.
         wd_sample_points (list[float], optional): The sample points for wind direction.
             If not provided, defaults to [-2 * wd_std, -1 * wd_std, 0, wd_std, 2 * wd_std].
+        fix_yaw_to_nominal_direction (bool, optional): Fix the yaw angle to the nominal
+            direction?   When False,
+            the yaw angle is the same across the sampled wind directions.  When True, the yaw
+            angle is fixed to the nominal wind direction and so of offset by the negative of
+            wind direction offset. Defaults to False.
```
ejsimley, line 49: This description is a little confusing to me. Could this say "When False, the yaw misalignment is the same across the sampled wind directions. When True, the turbine orientation is fixed to the nominal wind direction such that the yaw misalignment changes depending on the sampled wind direction."

ejsimley (commented):
This isn't related to the changes in the current PR, but I'm having a hard time understanding what the UncertainFlorisModel does. For example, how are wind speed and TI resolution used? Does uncertainty get modeled for those variables in addition to wind direction? Also, do Gaussian weights get applied to the power at the wd_sample_points? I'd suggest some additional explanation in the class docstring.

rafmudaf (approved):

floris/uncertain_floris_model.py
paulf81: Done!

ejsimley:
Sorry, accidentally rerequested review from @rafmudaf, who had approved already.

ejsimley (approved):
Extra documentation looks good

Issue #851, opened 2024-03-21
Calculation in complex terrain


Zhiming-Cai:
Hello! May I ask whether the latest FLORIS supports the calculation of wind farms in complex terrain?

misi9170:
Hi @CaiZhiming-HHU ! 

FLORIS does not have capability of modeling complex terrain directly. The best method for approximating the effects of complex terrain is to use the `heterogenous_inflow_config`. This allows you to specify different ambient wind speeds at different locations within the domain, which is one of the main effects that complex terrain causes. See [example 16](https://github.com/NREL/floris/blob/main/examples/16_heterogeneous_inflow.py) for an example of how to use the heterogeneous flow option (in particular, see lines 111--122, where the heterogeneous inflow is defined and passed to FLORIS). 

Alternatively, the hub heights of the turbines can be set individually by setting different `hub_height`s for each turbine in the farm. This will allow you to "raise" and "lower" some turbines. You can use the following code snippet as an example of how to do this.
```python
from floris.tools import FlorisInterface

fi = FlorisInterface("inputs/gch.yaml")
fi.reinitialize(layout_x=[0, 500.], layout_y=[0., 0.])
fi.calculate_wake()
print(fi.get_turbine_powers())
print("Power with original hub heights: {:.2f} kW".format(fi.get_farm_power()[0,0]/1000))

hub_height_T1 = 120.0

turbine_type = fi.floris.farm.turbine_definitions[0]
turbine_0 = turbine_type.copy()
turbine_1 = turbine_type.copy()

turbine_1["hub_height"] = hub_height_T1
turbine_1["turbine_type"] = "nrel_5MW_tall"

fi.reinitialize(
    turbine_type=[turbine_0, turbine_1],
    layout_x=[0, 500.],
    layout_y=[0., 0.]
)
fi.calculate_wake()

print(fi.get_turbine_powers())
print("Power with updated hub heights: {:.2f} kW".format(fi.get_farm_power()[0,0]/1000))
```

You should be cautious using this approach, because it will cause the wakes to pass over or under turbines at different heights, which may not be a great approximation of what happens to wakes in complex terrain. For example, if you model a hill by simply raising the hub height of the turbine on the hill, wakes from upstream turbines will pass below the turbine (through the hill), which is obviously not very realistic. Still, a combination of setting different `hub_height`s and using the `heterogenous_inflow_config` may get the flow patterns you need to sufficiently model the complex terrain.

PR #856, opened 2024-03-27
[BUGFIX] hardcoded reset to first turbine type in set_operation_model

misi9170:
@Bartdoekemeijer pointed out that `FlorisModel.set_power_thrust_model()`, introduced in #840, seems to introduce a bug that sets all turbines to the same type.

This PR fixes that bug, and also allows users to specify different `power_thrust_model`s for different turbines in the farm.

However, in order to do so, it _must change the name of the `turbine_type` field on the `turbine_definitions` entry dictionaries_. It is not immediately clear to me why this should be the case---the current operation is that if multiple entries of the `turbine_definitions` list all have the same `turbine_type` key, then _only the first list entry is used_. 

To see this, you can comment out the lines 
```
            turbine_type_list[tindex]["turbine_type"] = (
                turbine_type_list[tindex]["turbine_type"]+"_"+power_thrust_model[tindex]
            )
```
in this PR, which will cause the test to fail because both turbines will be `"simple-derating"`.
@rafmudaf , @bayc , do you know why it's the case that the `turbine_type` keys need to be unique? This also came up in the example code I wrote for #851, where I had to rename the `turbine_type` in order for FLORIS to recognize different hub heights.


rafmudaf (approved):

bayc:
@misi9170 I believe we set it up that way to facilitate grabbing the turbine definitions from the turbine library. You can see here (https://github.com/NREL/floris/blob/v4/floris/core/farm.py#L113) that we build a turbine definition cache, and each turbine definition needs a unique name so that it can be mapped across all the turbines that share that type.

PR #857, opened 2024-03-27
Improvements to WindRose resampling

misi9170:
@Bartdoekemeijer  noted a few place for improvement in the `WindRose` class, in particular:
1. Potentially unexpected behavior of the `resample_wind_rose()` method, which returns a new instantiation rather than modifying the current object
2. Unexpected default behavior of `plot_wind_rose()`, which resamples the data provided by default so makes it ambiguous which bins were specified (also, it turns out that providing `wd_step=None` to avoid resampling causes `plot_wind_rose()` to fail).
3. Minimal support for resampling to a higher bin resolution

This PR addresses those issues.
- [x] `inplace` option provided for `resample_wind_rose()` to modify the current object (defaults to `False` at the moment to retain former operation, but we could consider changing the default to `True`, which may be more intuitive. However, defaulting to `False` is more pandas-like).
- [x] Default for `plot_wind_rose()` is now not to resample
- [x] Rename `plot_wind_rose()` and `resample_wind_rose()` to `plot()` and `resample()` for consistency across `WindRose` and `WindTIRose()` (and for simplicity)
- [x] Enhanced resampling to a higher bin resolution (@paulf81 , I'm not sure that this is going to work with the current method of creating a `TimeSeries` and calling `to_WindRose`---any ideas here?)\*
- [x] Map changes onto `WindTIRose` @paulf81 

\* We may need to be careful here, as there are several ways we could interpolate the data, so we should take a conservative/simple approach (but less conservative than the current approach, which is simply to assign extra 0 frequency bins)


paulf81:
@misi9170  and @Bartdoekemeijer (and @ejsimley  if you time?) could you check quick what you think of the commit I just pushed up to address the point above on enhanced resampling for upsampling?  You'll see I made the following changes:

 - I changed the name of `resampling` to `aggregating` which I think more accurately captures what it does
 - I set a check that aggregating can only go to larger bins
 - I added a new function `resample_by_interpolation`, which works like the `interpolate` function in v3 FLORIS.

I'll need to next follow-through on the code if you like it in principle, or can change things.  But next steps will be:

- [ ] Make matching changes to WindTIRose
- [ ] Update existing tests and maybe add a couple tests
- [ ] Update examples and perhaps documentation (although that is in another pull request)

floris/wind_data.py
```diff
@@ -368,6 +376,20 @@ def resample(self, wd_step=None, ws_step=None, inplace=False):
             - Uses the bin weights feature in TimeSeries to resample the wind rose.
             - If `ws_step` or `wd_step` is not specified, it uses the current values.
         """
+
+        # If ws_step is passed in, confirm is it at least as large as the current step
+        if ws_step is not None:
+            if len(self.wind_speeds) >= 2:
+                if ws_step < self.wind_speeds[1] - self.wind_speeds[0]:
+                    raise ValueError("ws_step must be at least as large as the current step")
```
misi9170, line 384: @paulf81 maybe here, we should print what the current step is, which might help them set it correctly? 
something like:
```python
current_ws_step = self.wind_speeds[1] - self.wind_speeds[0]
if ws_step < current_ws_step:
    raise ValueError("ws_step provided must be at least as large as the current ws_step ({0:.2f} m/s)".format(current_ws_step)
```

floris/wind_data.py
```diff
@@ -405,6 +427,110 @@ def resample(self, wd_step=None, ws_step=None, inplace=False):
...
+        Returns:
+            WindRose: Resampled wind rose based on the provided or default step sizes.
+
+        """
+        if method == "linear":
+            interpolator = LinearNDInterpolator
+        elif method == "nearest":
+            interpolator = NearestNDInterpolator
+        else:
+            UserWarning("Unknown interpolation method: '{:s}'".format(method))
```
misi9170, line 452:
```python
UserWarning("Unknown interpolation method: '{:s}'. Available methods are 'linear' and 'nearest'".format(method))
```

misi9170:
@paulf81 I like your suggestion above!

floris/wind_data.py
paulf81: Sounds good, making the change...

floris/wind_data.py
paulf81: fixed as well

misi9170: Actually what is a "UserWarning"? does it cause an error? If not, what happens next?

paulf81: Oof, this should be a ValueError, switching

paulf81:
Ok @misi9170 , assuming all green checks above I think this is all good to go.  The one hedge is I did update the TiRose resample -> aggregate (mostly a name change but added checking that we aren't upsampling) but didn't add an interpolate method here, fearing the 3D interpolation and being short on time, can come back later for this?

PR #862, opened 2024-04-03
Enabling layout optimization for value

ejsimley:
# Enabling layout optimization for value

This PR adds the capability to perform layout optimization to maximize the expected *value* of the energy produced (e.g., the electricity market value), using the "value" array stored in a `WindData` object. In addition to the layout optimization enhancement, methods are also added to compute expected value and annual value production (AVP), and to  allow value to be added to a `WindData` object as a function of wind speed and/or direction. Note that the option to optimize for value was not added to the yaw optimization classes because optimizing for value would be equivalent to optimizing for energy - since yaw offsets are optimized for each wind condition independently, maximizing the value of the energy for a particular wind condition would be the same as maximizing the energy produced (assuming value is greater than or equal to zero).

Specific enhancements include:
- In floris_model.py, there are two new methods `get_expected_farm_value` and `get_farm_AVP` for getting the expected *value* produced by the wind farm and the annual *value* production, similar to `get_expected_farm_power` and `get_farm_AEP`.
- All of the `WindData` classes now include the method `assign_value_using_wd_ws_function` for assigning value using an arbitrary function and the method `assign_value_piecewise_linear` for specifying value using a piecewise linear function of wind speed. The piecewise linear approach was inspired by the mean electricity price vs. wind speed plots from Simley et al. "The value of wake steering wind farm flow control in US energy markets," Wind Energy Science, 2024. The default parameters are meant to approximate the curve for the SPP market region. The `WindRose` and `WindTIRose` classes also have a function for plotting value vs. TI.
- In "layout_optimization_base.py", the `LayoutOptimization` class now has an argument `use_value` to determine whether the objective is maximum energy or value. 
- In "layout_optimization_pyoptsparse.py" and "layout_optimization_scipy.py", the optimization objective functions will call `get_farm_AEP` or `get_farm_AVP` depending on the boolean `use_value` argument.
- Tests are added for `get_expected_farm_value`, `get_farm_AVP`, and `LayoutOptimizationScipy`, for the case when `use_value` is True.

I have an example ready too, but was thinking I'd wait to add it to PR #843 

As discussed with @paulf81 and @misi9170, I added the value-based layout optimization regression test to the existing "scipy_layout_opt_regression.py" file. However, we noticed that the tests in this file weren't getting run because the file name was missing the "_test" suffix. I renamed the file so it gets run now, but the original `test_scipy_layout_opt` test fails. I'll mark this as a draft PR until we resolve this. @bayc, are you familiar with this test? Do you think we should just update the baseline results (although they don't match perfectly, they are very close to the actual results that get computed), or was there a reason for removing this test from the regression test list maybe?


floris/optimization/layout_optimization/layout_optimization_scipy.py
```diff
@@ -18,9 +18,10 @@ def __init__(
         solver='SLSQP',
         optOptions=None,
         enable_geometric_yaw=False,
+        use_value=False,
     ):
         """
-        _summary_
+        _summary_ TODO: write summary
```
paulf81, line 24: @ejsimley if you want I'm happy to take a crack at this and commit to your branch.  Also, elsewhere in FLORIS we have the connection that the class docstring comes before the init function, in the same commit I can move this above?

floris/optimization/layout_optimization/layout_optimization_pyoptsparse.py
```diff
@@ -19,9 +19,35 @@ def __init__(
         storeHistory='hist.hist',
         hotStart=None,
         enable_geometric_yaw=False,
+        use_value=False,
     ):
-        super().__init__(fmodel, boundaries, min_dist=min_dist,
-                         enable_geometric_yaw=enable_geometric_yaw)
+        """
+        _summary_ TODO: write summary and describe the remaining arguments
```
paulf81, line 25: @ejsimley if you want I'm happy to take a crack at this and commit to your branch.  Also, elsewhere in FLORIS we have the connection that the class docstring comes before the init function, in the same commit I can move this above?

floris/optimization/layout_optimization/layout_optimization_scipy.py
ejsimley: Yes, please take care of that if you'd like, thanks. This is how the docstrings were before I started editing, so didn't feel like changing too much, but probably a good time to update it!

paulf81 (approved):
@ejsimley , this is nicely done and really elegant!  I agree with deferring adding the example until after the examples refactor PR is in.  Only comments:

1) On the scipy reg test, my vote would be we just go ahead and fix it, I'm assuming this was just an oversight (but maybe @rafmudaf you remember differently?)
2) I think I can write the two missing docstrings if you like (see comments)

floris/optimization/layout_optimization/layout_optimization_scipy.py
paulf81: will do!

ejsimley:
> @ejsimley , this is nicely done and really elegant! I agree with deferring adding the example until after the examples refactor PR is in. Only comments:
> 
> 1. On the scipy reg test, my vote would be we just go ahead and fix it, I'm assuming this was just an oversight (but maybe @rafmudaf you remember differently?)
> 2. I think I can write the two missing docstrings if you like (see comments)

@paulf81, thanks for working on the docstrings. For comment 1, I just updated the test results with the values I got locally. We'll see if the test passes here too. Hopefully we'll get some more input on those results, but I agree, probably fine to just fix it now. The small changes in the results could be relate to a dependency version.

rafmudaf:
>  However, we noticed that the tests in this file weren't getting run because the file name was missing the "_test" suffix.

Good catch @ejsimley. This file was added in https://github.com/NREL/floris/pull/819, and looking back at that PR I do think it was a typo @misi9170.

bayc:
> PR I

Yea, that was my bad. I added those tests and didn't catch that they needed the _test.

floris/optimization/layout_optimization/layout_optimization_pyoptsparse.py
```diff
@@ -42,7 +71,7 @@ def __init__(
             self.logger.error(err_msg, stack_info=True)
             raise ImportError(err_msg)
 
-        # Insantiate ptOptSparse optimization object with name and objective function
+        # Instantiate ptOptSparse optimization object with name and objective function
```
bayc, line 74: Also a typo in "ptOptSparse"

bayc (approved):

floris/optimization/layout_optimization/layout_optimization_pyoptsparse.py
ejsimley: There are a few arguments that aren't described in the docstring yet (timeLimit, storeHistory, hotStart). @bayc, it's not a big issue, but since we're updating this now, would you mind adding the descriptions? Or if you just want to write them here I can update the code.

PR #866, opened 2024-04-04
Add merge/reduce FLORIS objects

paulf81:
Just a rough draft for now @misi9170  and @Bartdoekemeijer  of sketching in merging/reducing FLORIS objects (taking these functions out of FLASC and into FLORIS), we'll look a mess until we merge refactor...

Then also, first decision, is this the right place?  Do we prefer a static member function?


misi9170:
I think this way (as a static function in floris_model.py) is reasonable. Will we want to do the same for uncertain models and parallel models? perhaps we can figure out a convenient function that could handle any of them? In that case, we may want this function on a separate file.

I'm more into the merge function than the reduce function. For reducing, we already have the following options:
- Updating `layout_x` and `layout_y` in `set()`
- Using `disable_turbines` for "temporarily" reducing (although the computation will still be "large").
I guess I feel that things could easily go wrong with the reduction function

paulf81:
> I think this way (as a static function in floris_model.py) is reasonable. Will we want to do the same for uncertain models and parallel models? perhaps we can figure out a convenient function that could handle any of them? In that case, we may want this function on a separate file.
> 
> I'm more into the merge function than the reduce function. For reducing, we already have the following options:
> 
> * Updating `layout_x` and `layout_y` in `set()`
> * Using `disable_turbines` for "temporarily" reducing (although the computation will still be "large").
>   I guess I feel that things could easily go wrong with the reduction function

Ok sounds good, I'll set it as a static member of the FlorisModel and remove the reduce function.  I think for uncertain/parallel these should be defined specifically for those, rather than trying to make a catch-all for now, which we could do later, but that moment catching all the special cases could be something.  At any rate I think I will just make the change in FlorisModel and then pause and wait for refactor merge just to simplify a little

paulf81:
@misi9170  the merge for this one is a little tedious given that I'm really just so far adding one function,

paulf81:
Ok @misi9170 , I think this one is ready, I think for now just adding the function is good.  I tested it out a bit with some test code, but I think for now we won't call it in the examples and can add that later (or just make greater use in flasc).  This is ready from my perspective

floris/floris_model.py
```diff
@@ -1722,3 +1722,72 @@ def reinitialize(self, **_):
...
+        reference_wind_heights = []
+        for fmodel in fi_list:
+            x_list.extend(fmodel.layout_x)
+            y_list.extend(fmodel.layout_y)
+
+            fi_turbine_type = fmodel.core.farm.turbine_type
+            if len(fi_turbine_type) == 1:
+                fi_turbine_type = fi_turbine_type * len(fmodel.layout_x)
+            elif not len(fi_turbine_type) == len(fmodel.layout_x):
+                raise UserWarning("Incompatible format of turbine_type in fmodel.")
```
misi9170, line 1769: Should these UserWarnings be errors?

floris/floris_model.py
```diff
@@ -1722,3 +1722,72 @@ def reinitialize(self, **_):
...
+
+            turbine_type_list.extend(fi_turbine_type)
+            reference_wind_heights.append(fmodel.core.flow_field.reference_wind_height)
+
+        # Derive reference wind height, if unspecified by the user
+        if reference_wind_height is None:
+            reference_wind_height = np.mean(reference_wind_heights)
+            if np.any(np.abs(np.array(reference_wind_heights) - reference_wind_height) > 1.0e-3):
+                raise UserWarning(
```
misi9170, line 1778: Same comment as above

floris/floris_model.py
```diff
@@ -1722,3 +1722,72 @@ def reinitialize(self, **_):
...
+                "Incompatible input specified. Please merge FlorisInterface objects "
+                " before inserting them into ParallelComputingInterface and UncertaintyInterface."
+            )
+
+        # Get the turbine locations and specifications for each subset and save as a list
+        x_list = []
+        y_list = []
+        turbine_type_list = []
+        reference_wind_heights = []
+        for fmodel in fi_list:
```
misi9170, line 1761: fmodel.reset_operation()

misi9170 (changes requested):

misi9170:
@paulf81 I made a few changes in d127c70
- Changed UserWarning to ValueError in a few places (please feel free to change back if you intended to raise UserWarnings)
- Changed `fi` and `FlorisInterface` to `fmodel` and `FlorisModel` throughout, and made warning about not providing `FlorisModel`s more general
- Included a `reset_opeartion()` call so that any control setpoints, which are specified for a specific number of turbines, are dropped.

If you're happy with those changes, feel free to merge this in with Squash and Merge.

misi9170 (approved):

floris/floris_model.py
```diff
@@ -1722,3 +1722,72 @@ def reinitialize(self, **_):
...
+                for their reference_wind_height.
+
+        Returns:
+            fmodel_merged (FlorisModel): The merged FlorisModel object,
+                merged in the same order as fmodel_list. The objects are merged
+                on the turbine locations and turbine types, but not on the wake parameters
+                or general solver settings.
+        """
+
+        if not isinstance(fmodel_list[0], FlorisModel):
```
rafmudaf, line 1748: @misi9170 @paulf81 I just came across this and I think it's not validating `fmodel_list` correctly. If I understand correctly that it's a 1-D list, then it's only validating the first element. You can check all elements with something like:
```python
# does `isinstance` on each element of the list through `map`
# then checks whether all the elements are True.
# If any elements are not True, then it goes into the if-statement
if not all( list( map(isinstance, fmodel_list, len(fmodel_list)*[FlorisModel]) ) ):

# or the same thing but more simply
if not all( type(fm) == FlorisModel for fm in fmodel_list ):
```

It may or may not be worth fixing right now, but I wanted to give you a heads up in case you see anything funny in the future with this function.

PR #868, opened 2024-04-05
Remove setpoints and wind condition specifics from calculate_XX_plane methods

misi9170:
Various methods used for visualization on `FlorisModel` (`calculate_horizontal_plane()`, `calculate_cross_plane()`, and `calculate_y_plane()`) still accepted `yaw_angles` as an input, but not other control setpoints.

If `yaw_angles` were passed, they _were_ applied; and if `yaw_angles` were not passed, the stored yaw angles were used (which could lead to findex length clashes, if the `fmodel` had `n_findex` > 1). 

Simply removing `yaw_angles` as an input and allowing those that had already been `set` to be used could create a findex clash. Instead, I decided to:
- remove `yaw_angles` as an input
- also remove `wd`, `ws`, and `ti` as inputs
- give the option to the user of which `findex` they would like to visualize
- Added some infrastructure to enable this: new method `set_for_viz`; deepcopies of `fmodel` to get around issue with setpoints not being stored on the dictionary (which has come up a couple of times now---we may need a better fix for this issue).

The result works, but may be somewhat slower than before. However, I'm not sure this is the path we'd like to go down, so I'd like some feedback here. If we are OK with this, I can also make some changes to simplify `calculate_horizontal_plane_with_turbines()` on flow_visualization.py. Once those changes are in, the `check_wind_condition_for_viz()` method on `FlorisModel` can likely be removed. For now, I've simply removed the erroneous type hints on `check_wind_condition_for_viz()`.

I'll also add some tests for all of this if we do want to go with this approach.


floris/floris_model.py
```diff
@@ -885,51 +886,64 @@ def get_turbine_TIs(self) -> NDArrayFloat:
 
     ### Methods for sampling and visualization
 
+    def set_for_viz(self, findex: int, solver_settings: dict) -> None:
```
paulf81, line 889: Could/should the findex default to 0?  Does the user have to supply a findex if n_findex=0?

paulf81: Should have read more before commenting, this seems to be default in the higher-level functions

misi9170:
> > may be somewhat slower
> 
> What's the cause of this being slower?

@rafmudaf , well, I'm not sure that it's slower, but I changed from exporting the existing `fmodel` to a dictionary and then reinstantiating to simply making a `deepcopy` of the exisiting `fmodel`. I think that's more robust (I did it in order to save the control setpoints), but I think more memory is needed for the deep copy than a configuration dictionary.

misi9170:
I've now gone ahead and made the commensurate changes to `calculate_horizontal_plane_with_turbines()` and removed `FlorisModel.check_wind_condition_for_viz()`, which was made redundant.

floris/floris_model.py
```diff
@@ -942,54 +943,65 @@ def get_turbine_TIs(self) -> NDArrayFloat:
...
+        """
+        Set the floris object to a single findex for visualization.
+
+        Args:
+            findex (int): The findex to set the floris object to.
+            solver_settings (dict): The solver settings to use for visualization.
+        """
+        # TODO: can get from properties once #843 is merged
```
rafmudaf, line 954: Is this resolved or still relevant?

misi9170: Good catch, thanks---I've addressed this now, as well as a couple of other TODOs.

Issue #865, opened 2024-04-04
FlorisModel set method is slow to call if we don't need to update flow field parameters


achenry:
I'm using FLORIS in a MPC context, where at each time-step I need to set the wind speeds and directions once, and then iterate for many values of the yaw offsets. My optimization procedure slowed significantly with the update from `reinitialize` to. `set`, and on profiling the code I found that virtually all of the time was spent in the `_reinitialize` method. My workaround has been to call `set` once for each time step, and then during the optimizer iterations to call:

```
self.fi.env.core.farm.set_yaw_angles(self.fi.env.core.farm.yaw_angles)
self.fi.env._set_operation(
    yaw_angles=yaw_offsets,
    disable_turbines=self.offline_status,
)
self.run()
```
to update the turbine powers for new yaw angles. It may be that the `run` method could also be optimized for yaw only changes, I haven't investigated.

achenry:
I’m not sure if that line is necessary, I just included it because it was in the set method.

PR #869, opened 2024-04-05
Make set_operation method public

misi9170:
As mentioned in #865, there are some cases where users may want to set _only_ the control setpoints. Currently, the method for doing this, `FlorisModel._set_operation()`, is private. This PR makes that method public so that it may be used.

We intend that only power users will use this functionality---most will simply use the `set` method to alter the control setpoints.

In making this change I found that there was no check that the `yaw_angles` or `power_setpoints` second dimension matched the number of turbines, and in fact it was possible to set and run `yaw_angles` with more than the number of turbines. This is now checked for.

I spoke to @rafmudaf about a possible issue created by the following workflow:
```python
fmodel.set_operation(yaw_angles=some_yaw_angles)
fmodel.run()
fmodel.set_operation(yaw_angles=some_other_yaw_angles)
powers = fmodel.get_turbine_powers()
```
This would be bad, because the `power` would have the wake caused by `some_yaw_angles` but the power behavior caused by `some_other_yaw_angles`.

To make sure this does not happen, I have set the `core`'s `state` back to `UNINITIALIZED` if `set_operation()` is called. This then throws a `RuntimeError` if `get_turbine_powers()` is called before calling `run()` again. I test for this in the tests.

### DIFF
--- a/floris/floris_model.py
+++ b/floris/floris_model.py
@@ -251,9 +251,19 @@ class FlorisModel(LoggingManager):
         """
         # Add operating conditions to the floris object
         if yaw_angles is not None:
+            if np.array(yaw_angles).shape[1] != self.core.farm.n_turbines:
+                raise ValueError(
+                    f"yaw_angles has a size of {np.array(yaw_angles).shape[1]} in the 1st "
+                    f"dimension, must be equal to n_turbines={self.core.farm.n_turbines}"
+                )
             self.core.farm.set_yaw_angles(yaw_angles)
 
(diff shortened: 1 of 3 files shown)


paulf81 (approved):
This looks solid @misi9170 thanks!

misi9170:
I'm going to wait to merge until I get response on #865---that should be fine though, this doesn't really affect any other open PRs.

misi9170:
Thanks @achenry for calling this out!

PR #872, opened 2024-04-05
Fixing datatype of value in WindData objects

ejsimley:
# Bug fix: setting value as a float in WindData objects and removing a warning when computing AVP

This PR fixes two small issues related to the WindData value array. 

- First, when wind speeds were defined as integers when creating a WindData object, the values array resulting from calling the WindData `assign_value_piecewise_linear` method would also be integers. Since we want values to be able to take non-integer values, `dtype=float` is now added when assigning values.
- Second, in `get_farm_AVP` in floris_model.py, a warning about values being treated as 1 was printed when the WindData attribute in the floris model was a TimeSeries object. However, TimeSeries objects have a value array, so this warning isn't necessary. Now it only prints the warning when the floris model has no WindData attribute.

### DIFF
--- a/floris/floris_model.py
+++ b/floris/floris_model.py
@@ -843,11 +843,7 @@ class FlorisModel(LoggingManager):
                 "operation."
             )
 
-        if (
-            values is None
-            and not isinstance(self.wind_data, WindRose)
-            and not isinstance(self.wind_data, WindTIRose)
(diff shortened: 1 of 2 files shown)


paulf81 (approved):
Great catch, thanks Eric!

PR #878, opened 2024-04-10
[BUGFIX] Add enable_active_wake_mixing field to v3 to v4 documentation and conversion

misi9170:
As part of the merge of #842 , a new field `enable_active_wake_mixing` was added to the `wake` field of the main floris input yaml.

The documentation describing the switch from v3 to v4 missed this detail; additionally, the converter utility convert_floris_input_v3_to_v4.py does not currently add this necessary field, so converted floris input files would currently not actually run.

In this PR, I've fixed the documentation added the `enable_active_wake_mixing` field to the converter, and tested it by converting a [v3 GCH floris input file](https://github.com/NREL/floris/blob/b4d9d1eddca9f4fa197f3e76a295d2349a63d584/examples/inputs/gch.yaml) to v4 and running the first example with it.

I'd also like to fix the [Helix operation model documentation](https://nrel.github.io/floris/operation_models_user.html#awc-model), which I plan to add to this PR before we merge it..

### DIFF
--- a/docs/v3_to_v4.md
+++ b/docs/v3_to_v4.md
@@ -108,11 +108,15 @@ in v3. However, there are a few changes to the fields on each that mean that exi
 will not run in v4 as is.
 
 #### Main FLORIS input yaml
-The only change in fields on the main FLORIS input file is that the `turbulence_intensity` field,
+On the main FLORIS input file, the `turbulence_intensity` field,
 which was specified as a scalar in FLORIS v3, has been changed to `turbulence_intensities`, and
 should now contain a list of turbulence intensities that is of the same length as `wind_directions`
 and `wind_speeds`. Additionally, the length of the lists for `wind_directions` and `wind_speeds`
 _must_ now be of equal length.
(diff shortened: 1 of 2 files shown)


rafmudaf:
@misi9170 since develop and main are at the same commit now, you might consider merging this one to main. That way, you can also fix the broken links in the README here.

misi9170:
> @misi9170 since develop and main are at the same commit now, you might consider merging this one to main. That way, you can also fix the broken links in the README here.

That's a good idea. I'll add those fixes later today. Will we need a new release tag, or is it fine to merge directly to main (and then back to develop) without a release tag in that case?

rafmudaf:
I'd bet that there will be more things to fix in the next couple of weeks, so I would wait to aggregate them into v4.0.1.

jfrederik-nrel (approved):
Looks good, Misha!

paulf81 (approved):
This looks good, thank you @misi9170

PR #880, opened 2024-04-12
Add test for v3_to_v4 input file converters

paulf81:
# Add CI testing of v3 to v4 conversions

I was thinking as we add features to v4 it'll be good to have some testing that the converter from v3 to v4 is keeping up if we we add new aspects to v4.  The test I added works by:

- Keeping a copy in tests of input files from the v3 branch of FLORIS
- Making a copy into that folder of the current converter scripts
- Running those scripts on the floris and turbine inputs
- Opening a FLORIS model on the converted script

One special case handled is that the v3 yaml points to a turbine_type "XXXXX" which is replaced after conversion with a !include pointing to the newly created v4 turbine file.  Originally I had the !include in the v3 input but this gets merged to early and the whole turbine file ends up in the v4 floris yaml.

## Related issue
Inspired by #878


misi9170:
Thanks @paulf81 ! Just to tag it, this will hopefully avoid situations like #878 , where I missed updating the v3 to v4 converter because we weren't testing it.

tests/convert_v3_to_v4_test.py
```diff
@@ -0,0 +1,49 @@
...
+    # Go through the file filename_v4_turbine and where the place-holder string "XXXXX" is found
+    # replace it with the string f"!include {filename_v4_turbine}"
+    with open(filename_v4_floris, "r") as file:
+        filedata = file.read()
+    filedata = filedata.replace("XXXXX", f"!include {filename_v4_turbine}")
+    with open(filename_v4_floris, "w") as file:
+        file.write(filedata)
+
+    # Now confirm that the converted file can be loaded by FLORIS
+    _ = FlorisModel(filename_v4_floris)
```
misi9170, line 43: I think we should also confirm that it runs, because I'm not 100% sure that just loading it catches all possible missing fields? I don't think we should `set` anything, just instantiate to `fmodel` and `run`.

misi9170 (changes requested):
Just requesting we also run the loaded v4 `fmodel` to be sure; otherwise, this looks great, thanks

tests/convert_v3_to_v4_test.py
paulf81: Good suggestion!  Adding now

misi9170 (approved):
Thanks!

tests/convert_v3_to_v4_test.py
```diff
@@ -0,0 +1,52 @@
...
+    # Print the current directory
+    print(os.getcwd())
+
+    # Run the converter on the turbine file
+    os.system(f"python convert_turbine_v3_to_v4.py {filename_v3_turbine}")
+
+    # Run the converter on the floris file
+    os.system(f"python convert_floris_input_v3_to_v4.py {filename_v3_floris}")
+
+    # Go through the file filename_v4_turbine and where the place-holder string "XXXXX" is found
```
misi9170, line 34: Sorry, just noticed maybe a type here: should this comment be referring to `filename_v4_floris` rather than `filename_v4_turbine`?

paulf81: yes, thanks!

Issue #709, opened 2023-09-08
Bug report: cubature interpolation does not work with CC model


Bartdoekemeijer:
# Cubature interpolation throws error for CC wake velocity deficit model

## How to reproduce
Here is a minimal code that reproduces the error:
```
# Instantiate FLORIS using either the GCH or CC model
from floris.tools import FlorisInterface
fi = FlorisInterface("inputs/cc.yaml")

# Use turbine cubature grid for rotor interpolation
fi_dict = fi.floris.as_dict()
fi_dict["solver"] = {
    "type": "turbine_cubature_grid",
    "turbine_grid_points": 3
}
fi = FlorisInterface(fi_dict)

# Calculate wakes and retrieve turbine powers
fi.calculate_wake()
turbine_powers = fi.get_turbine_powers()
print(turbine_powers.flatten())
```

## Relevant output
The example code above results in the following error first:
```
Exception has occurred: AttributeError
'TurbineCubatureGrid' object has no attribute 'x'
  File "/mnt/c/Users/Bart.Doekemeijer/OneDrive - Shell/Documents/python_scripts/floris/floris/simulation/grid.py", line 381, in set_grid
    self.x = np.take_along_axis(self.x_sorted, self.unsorted_indices, axis=2)
```

If we then remove those lines from the code (is that correct?), the example script runs fine when you replace `cc.yaml` with `gch.yaml`. With `cc.yaml`, we have a second error come up:
```
Exception has occurred: ValueError
non-broadcastable output operand with shape (1,1,1,1,1) doesn't match the broadcast shape (1,1,1,1,1,9,3)
  File "/mnt/c/Users/Bart.Doekemeijer/Documents/python_scripts/floris/floris/simulation/solver.py", line 564, in cc_solver
    effective_yaw_i += added_yaw
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

What I also found is that the `velocities` variable being passed into the `Ct(...)` function in `turbine.py` seems to not be the right dimensions when using the CC model.

- With CC, the shape of `velocities` is `(1, 1, 3)`,
- With GCH, the shape of `velocities` is `(1, 1, 1, 9, 1)`.

## Floris version
Using the current `develop` branch with latest commit `8ece0f`.

## System Information

 - OS: Ubuntu 22.04
 - Python version: Python 3.11
 - Library versions:
   - `numpy==1.24.3`
   - `numexpr==2.8.4`
   - `scipy==1.10.1`

[...]

PR #881, opened 2024-04-17
[BUGFIX] Update CubatureGrid for 4-dimensional data structures

rafmudaf:
# Update CubatureGrid for 4-dimensional data structures
In #764, the core data structures in FLORIS were changed from 5-dimensional to 4-dimensional by collapsing the two left-most dimensions. Some parts of the CubatureGrid class were updated then, but it was not fully changed. Since this grid-type is not included in the automated tests, the error was not caught.

Separately, the check for a valid grid resolution was incorrect, so that has also been updated here.

This change can be verified to work with the following script. Before this change, the script raises an error.

```python
solver_settings = {
    "type": "turbine_cubature_grid",
    "turbine_grid_points": 3
}
fmodel.set(solver_settings=solver_settings)
fmodel.run()

turbine_powers = fmodel.get_turbine_powers()
print(turbine_powers.flatten())
```

## Related issue
Tangentially related to #709.

CubatureGrid class

### DIFF
--- a/floris/core/grid.py
+++ b/floris/core/grid.py
@@ -317,21 +317,28 @@ class TurbineCubatureGrid(Grid):
             ),
             dtype=floris_float_type
         )
-        _x = x[:, :, :, None, None] * template_grid
-        _y = y[:, :, :, None, None] * template_grid
-        _z = z[:, :, :, None, None] * template_grid
+        _x = x[:, :, None, None] * template_grid
+        _y = y[:, :, None, None] * template_grid
+        _z = z[:, :, None, None] * template_grid
+
+        n_coordinates = len(yv)
+        yv = np.broadcast_to(yv, (self.n_findex, self.n_turbines, n_coordinates))
+        yv = np.expand_dims(yv, axis=-1)
+        zv = np.broadcast_to(zv, (self.n_findex, self.n_turbines, n_coordinates))
+        zv = np.expand_dims(zv, axis=-1)
+
         for ti in range(self.n_turbines):
-            _y[:, :, ti, :, :] += yv[None, None, :, None]*self.turbine_diameters[ti] / 2.0
-            _z[:, :, ti, :, :] += zv[None, None, :, None]*self.turbine_diameters[ti] / 2.0
+            _y[:, ti, :, :] += yv[:, ti] * self.turbine_diameters[ti] / 2.0
+            _z[:, ti, :, :] += zv[:, ti] * self.turbine_diameters[ti] / 2.0
 
         # Sort the turbines at each wind direction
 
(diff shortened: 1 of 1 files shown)


rafmudaf:
@Bartdoekemeijer heads up that this was an issue and will be fixed. However, it does not address #709.

Bartdoekemeijer:
Thanks @rafmudaf. Also, the cubature scheme generates a quadratic number of points similar to the default square turbine grid method, so 2x2, 3x3, 4x4. However, in the 3x3 setting for example, there are only 7 _unique_ points. I spoke with Misha before about this, and we argued that perhaps a further simplification in FLORIS could be

`(n_wind_directions, n_wind_speeds, n_rotor_points)`

rather than
`(n_wind_directions, n_wind_speeds, turbine_grid_n, turbine_grid_n)`

This would allow an easy speed-up of almost 30% using the cubature method without any loss in accuracy.

rafmudaf:
@Bartdoekemeijer It's a good point, and @misi9170 also mentioned it to me. I agree that this change is worthwhile. However, for the cubature grid, we keep one dimension flat and the other contains the points. For N=3, it's `(n findex, n turbines, 9, 1)` in the v4 form, so collapsing the dimensions wouldn't solve the problem. Instead, we can remove any redundant points at the center and combines their weights. This pull request is only to fix the bug so that it will at least run, but I've also opened #883 where we can continue to talk through the remaining items to finalize this.

misi9170:
Hi @rafmudaf , I think this looks good and I've confirmed that your example script runs on this branch but fails on develop, as you said. 

This does beg the question though---should we be adding a test for the cubature grid, perhaps just something very simple based on your example script?

rafmudaf:
Hi @misi9170,

Thanks for reviewing these changes. I have the infrastructure for some tests in my fork's [cubature_testing](https://github.com/rafmudaf/floris/tree/cubature_testing) branch, but I haven't taken the time to think through good tests for this feature. My idea was to fix the bug in the code that exists in v4.0, and then add tests and validate for v4.1 as described in #883. However, I can add tests to this pull request before merging, if you'd rather.

And no worries on the target branch, I'll change it to develop.

misi9170:
> Hi @misi9170,
> 
> Thanks for reviewing these changes. I have the infrastructure for some tests in my fork's [cubature_testing](https://github.com/rafmudaf/floris/tree/cubature_testing) branch, but I haven't taken the time to think through good tests for this feature. My idea was to fix the bug in the code that exists in v4.0, and then add tests and validate for v4.1 as described in #883. However, I can add tests to this pull request before merging, if you'd rather.
> 
> And no worries on the target branch, I'll change it to develop.

Sounds good, thanks!

Issue #893, opened 2024-04-25
Bug report: None Turbulence model returns the wrong values


Bartdoekemeijer:
# The 'None' turbulence model is incorrect in its current implementation.

## How to reproduce
The turbulence model is a wake-added turbulence model, which is added on top of the ambient turbulence intensity. Currently, the 'None' turbulence model returns the ambient turbulence values, while it should return zeros.

## Relevant output

The turbulence model should return zeros.

## Floris version
FLORIS v3.6 but it's also in v4:
https://github.com/NREL/floris/blob/6181d78d9c9bd52a3a23f97f6d37ef3ebd84e11e/floris/core/wake_turbulence/none.py#L32C20-L32C44

## System Information
N/A

Bartdoekemeijer:
Sorry for the short bug report. Jasper Kreeft found this while cross-validating wake modeling tools and wanted to flag this.

misi9170:
Thanks @Bartdoekemeijer ! I can see the issue that you are talking about. When running the GCH model (solved using `sequential_solver()`) an atmospheric TI of 0.06, with the `NoneWakeTurbulence` model selected, the `wake_added_turbulence_intensity`, calculated [here](https://github.com/NREL/floris/blob/main/floris/core/solver.py#L220) is 0.06, rather than the expected 0. 

This is a direct result of [this line](https://github.com/NREL/floris/blob/main/floris/core/wake_turbulence/none.py#L32), which appears to be incorrect. Since the TI returned by this model (as well as any other turbulence intensity model) will be added to the ambient TI, this line should simply be changed to `return np.zeros_like(x)`. 

I will check to see if there is anywhere where the turbulence intensity model is being used to represent the "full" turbulence intensity (combined ambient and wake), to ensure that changing this to 0 does not conflict with the usage in other solvers.

misi9170:
Addressed in #894

PR #894, opened 2024-04-26
[BUGFIX] `NoneWakeTurbulence` returns zero (added) turbulence

misi9170:
As pointed out by @Bartdoekemeijer in #893, the `NoneWakeTurbulence` model currently returns the ambient turbulence. However, this is then _added_ (as wake added turbulence) to the existing ambient turbulence, meaning that the "final" turbine turbulence is above the ambient TI level.

The expected behavior is that `NoneWakeTurbulence` makes no change to the turbulence intensity in turbines' wakes. This PR implements that change.

To see the issue, and how this fixes it, change the gch.yaml input field `turbulence_model` to `none` (replacing `crespo-hernandez`, and switch off all secondary effects (`enable_secondary_steering`, `enable_yaw_added_recovery`, `enable_active_wake_mixing`, and `enable_transverse_velocities` all set to `false`). Then, add the following line to example 001_opening_floris_and_computing_power.py somewhere after `fmodel.run()`:

```python
print(fmodel.core.flow_field.turbulence_intensity_field_sorted[0,:,0,0])
```

Running the modified 001 example on the develop branches produces
```
[0.06       0.08485281]
```
where the second turbine clearly has higher turbulence due to the presence of the first turbine's wake, even though the `turbulence_model` is `none`.

After the fix, the example produces
```
[0.06 0.06]
```
as expected.

I'm also using this opportunity to update one of the comments in example 002_visualizations.py, which was noted to be incorrect in #892.

### DIFF
--- a/examples/002_visualizations.py
+++ b/examples/002_visualizations.py
@@ -69,8 +69,10 @@ ax.set_title("Use Provided Turbine Names")
 # we show the horizontal plane at hub height, further examples are provided within
 # the examples_visualizations folder
 
-# For flow visualizations, the FlorisModel must be set to run a single condition
-# (n_findex = 1)
+# For flow visualizations, the FlorisModel may be set to run a single condition
+# (n_findex = 1). Otherwise, the user may set multiple conditions and then use
+# the findex_for_viz keyword argument to calculate_horizontal_plane to specify which
+# flow condition to visualize.
 fmodel.set(wind_speeds=[8.0], wind_directions=[290.0], turbulence_intensities=[0.06])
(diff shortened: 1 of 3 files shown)


misi9170:
@paulf81 Thanks for the suggestion. I've gone ahead and added a test for `NoneWakeTurbulence`. I've denoted this as a regression test, because although this particular wake turbulence model is more of a unit or integration test, I'm imagining that if we want to add tests for the other turbulence models in future they'll look more like regression tests (i.e., results will be checked against stored values for the turbulence intensity at turbine locations).

Bartdoekemeijer:
Great stuff, thanks for the quick bugfix!

PR #903, opened 2024-05-07
correct placement of a in gauss exponent

jaredthomas68:
# Correct placement of `a` in the gauss model exponent

It came to my attention that the placement of `a` in the gauss exponential equation does not match the paper (see eq. 10): https://wes.copernicus.org/articles/3/819/2018/wes-3-819-2018.pdf
image

## Related issue

None

Line 222 in `gauss.py`

All tests pass, which I find interesting. I did not change the tests, only the placement of `a` in line 222 of `gauss.py`

### DIFF
--- a/floris/core/wake_velocity/gauss.py
+++ b/floris/core/wake_velocity/gauss.py
@@ -219,7 +219,7 @@ def rC(wind_veer, sigma_y, sigma_z, y, y_i, delta, z, HH, Ct, yaw, D):
         "sin(wind_veer) ** 2 / (2 * sigma_y ** 2) + cos(wind_veer) ** 2 / (2 * sigma_z ** 2)"
     )
     r = ne.evaluate(
-        "a * ((y - y_i - delta) ** 2) - 2 * b * (y - y_i - delta) * (z - HH) + c * ((z - HH) ** 2)"
+        "(a * (y - y_i - delta) ** 2) - 2 * b * (y - y_i - delta) * (z - HH) + c * ((z - HH) ** 2)"
     )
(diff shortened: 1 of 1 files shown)


rafmudaf:
Guys isn't it the same thing?

```
a * ((y - y_i - delta) ** 2)
vs
(a * (y - y_i - delta) ** 2)
```
The exponent is applied to the thing in parenthesis and then a is multiplied to all of that.

misi9170:
> Guys isn't it the same thing?
> 
> ```
> a * ((y - y_i - delta) ** 2)
> vs
> (a * (y - y_i - delta) ** 2)
> ```
> 
> The exponent is applied to the thing in parenthesis and then a is multiplied to all of that.

:) Yeah, that looks like the same thing

rafmudaf:
In addition to checking the order of operations, we can test it in a Python interpreter:
```
(floris) >>mbp@~ $ python
Python 3.12.1 | packaged by conda-forge | (main, Dec 23 2023, 08:01:35) [Clang 16.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> a = 5
>>> y = 4
>>> y_i = 3
>>> delta = 2
>>> a * ((y - y_i - delta) ** 2)
5
>>> (a * (y - y_i - delta) ** 2)
5
>>> 
```

paulf81:
Not at all @jaredthomas68 , thanks very much for taking the time to put in an issue when you saw it, and thanks team for debugging!

rafmudaf:
It's not a problem @jaredthomas68, and thank for taking the time to submit a change. In my experience, this sort of situation comes up when there's some other odd but yet unknown thing happening. If that's the case for you, let me know and I'm happy to help, if needed.

jaredthomas68:
Thanks, I'm just working on adding a feature that requires changing a few lines in this part of the code and I noticed what I thought was an error.

PR #904, opened 2024-05-07
Ignore includes in v3->v4 conversions

paulf81:
# Ignore includes in v3->v4 conversions

Currently when you convert an input yaml from v3 -> v4 using the converter script, any !include statements to turbine files get unpacked within the yaml file and then dumped into the file.  While this can work, it means that the yaml file can get quite long if there were a number of turbines listed, and also the yaml is no longer tied to the turbine file, and changes made to the turbine file will have to be made to each instance within the yaml.

This pull pull request modifies the yaml load/dump routines within the converter to pass through those include statements directly.  It further updates the convert script test to confirm the new methods are working.  Previously, to avoid this issue a sentinal "XXXXX" string was used but this can now go back to the turbine yaml name

### DIFF
--- a/floris/convert_floris_input_v3_to_v4.py
+++ b/floris/convert_floris_input_v3_to_v4.py
@@ -1,10 +1,7 @@
-
 import sys
 from pathlib import Path
 
 import yaml
-
-from floris.utilities import load_yaml
 
 
 """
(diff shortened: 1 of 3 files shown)


floris/convert_floris_input_v3_to_v4.py
```diff
@@ -64,10 +70,20 @@
...
-    yaml.dump(
-            v4_floris_input_dict,
-            open(output_path, "w"),
-            sort_keys=False
-        )
+    yaml.dump(v4_floris_input_dict, open(output_path, "w"), sort_keys=False)
+
+    # Open the output file and loop through line by line
+    # if a line contains the substring !include, then strip all
+    # occurrences of ' from the line
```
misi9170, line 77: This comment could perhaps state that the yaml load / dump sequence adds the string marks ' ' to the !include statements, which then need to be removed from the final output file. I didn't follow that at first. Otherwise, this looks great!

misi9170 (approved):
Very minor suggestion for an improved comment; otherwise, looks great and runs as expected, thanks @paulf81 !

PR #911, opened 2024-05-27
Change docs logo

rafmudaf:
# Change the documentation logo
The logo currently in the docs site was created quickly and without much group consensus a few years ago. At that time, I felt that it was important to have some very easily identifiable indication of what the software does as a logo, so I created something intentionally on the nose. However, this was not intended to be a long term decision, and I think that current logo has overstayed its welcome. Primarily, it's not really a logo since it's a gif. Secondly, it's not possible to embed it in many types of messaging, and it is too complex.

@misi9170 recently shared an image that I think could be ultimately adopted as the next iteration of the FLORIS logo. The one he shared was square, and I cropped it to have a 2:3 height-to-width ratio. This pull request included the change to the docs site to use this logo.

Ultimately, I think we should discuss a long term logo and general branding. There are some issues with even this logo:
- It doesn't say "floris"
- The colors are not generally accessible
- A black and white version may not have much contrast depending on the calculation of brightness per color channel

However, I think it's a step in the right direction, and I suggest to adopt it even as a temporary logo.

## Related issue
None

Docs site and other associated branding for FLORIS

Nice one @misi9170!


rafmudaf:
Also, I used it [here](https://rafmudaf.github.io/wcomp/) and I think it looks nice.

christiannvaughn:
We need to run this by our WETO TM and Comms before making it official.

rafmudaf:
> We need to run this by our WETO TM and Comms before making it official. 

Did I say logo? I meant docs site image.

christiannvaughn:
Well the official logo is on our to-do list. I will add it to the agenda for our next meeting. Maybe we can provide some FLORIS 4.0 release stats in that meeting too.

Issue #910, opened 2024-05-23
Add 3rd dimension to HeterogeneousMap


paulf81:
# Add 3rd dimension to HeterogeneousMap

On original submission of HeterogeneousMap, I neglected to include a capability for 3rd (z) dimension, although this is already supported by the underlying heterogeneous_inflow_config dictionary.  Starting an issue to track this issue and list out intended improvements

## Proposed solution
Proposed solution should include:

  1) Z-term optionally included in HeterogeneousMap and routed through each member function
  2) Function to build a new 2D HeterogeneousMap via composition and return it
  3) Function to visualize 3D heterogeneity (maybe, if not too complicated)
  4) Function to show apparent shear layer
  5) Updated testing
  6) Update documentation

And then I plan to fold-in some additional improvements I've realized I should have originally included:

  7) Functions to provide statistics of HeterogeneousMap, including outliers, to help screen potential issues
  8) Functions to return subset HeterogeneousMap by wind direction and wind speed, also by composition

@misi9170 , just wanted to flag this to you, I'll do the work but just making sure you agree with this list, or if you think of any additions or subtractions

misi9170:
Addressed by #915

PR #915, opened 2024-06-03
Add z/3d to HeterogeneousMap

paulf81:
# Add z/3D capabilities to HeterogeneousMap

On original submission of HeterogeneousMap, I neglected to include a capability for 3rd (z) dimension, although this is already supported by the underlying heterogeneous_inflow_config dictionary.  Opened #910 to document the issue and this pull request is meant to address it.  In #910 I proposed the following list of items to include to fully take advantage of the underlying 3D capability, placing that list here as a checklist to track progress to completion.

Update, you'll see in list below but as I got into it I found some of the ideas I had thought to put in here belong in an outer calling code, which has access to more data before the HeterogeneousMap is built

## Items to address

- [x] Z-term optionally included in HeterogeneousMap and routed through each member function
- [x] Function to build a new 2D HeterogeneousMap via composition and return it
- [x] Added a check in FlorisModel that not both z in het and shear are defined, as this must yield strange results
- [x] Added a test that computing shear via het and via the shear exponent produce expected results
- [ ] ~~Function to visualize 3D heterogeneity (maybe, if not too complicated)~~
- [ ] ~~Function to show apparent shear layer~~
- [x] Updated testing
- [x] Update documentation
- [ ] ~~Functions to provide statistics of HeterogeneousMap, including outliers, to help screen potential issues~~
- [ ] ~~Functions to return subset HeterogeneousMap by wind direction and wind speed, also by composition~~

## Open for discussion

I added a ValueError for the case that fmodel.set() indicates that both `heterogeneous_inflow_config` contains a `z` term and `wind_shear` is a value other than 0.0.  This could alternatively be a warning, but thought an error is safer since this seems like it should be an error of double counting.

## Related issue
#910 

heterogeneous_map.py


paulf81:
ok @misi9170 this is now ready for review

floris/floris_model.py
```diff
@@ -210,6 +210,22 @@ def _reinitialize(
...
+
+            # If both heterogeneous_inflow_config contains a z term,
+            # and flow_field_dict["wind_shear"] is
+            # not 0, raise and error the user that there is potential
+            # for double-counting of shear
+            if (
+                "z" in heterogeneous_inflow_config
+                and flow_field_dict["wind_shear"] != 0.0
+            ):
+                if heterogeneous_inflow_config['z'] is not None:
```
misi9170, line 222: I don't think there need to be nested `if` statements here? Can we just do:
IF "z" in heterogeneous_inflow_config AND h_i_c["z"] is not None AND 0 shear:
    raise ValueError()
    
?

floris/floris_model.py
```diff
@@ -210,6 +210,22 @@ def _reinitialize(
...
+                    raise ValueError(
```
misi9170, line 223: We could consider using logger warning here, rather than a ValueError, in case the user would like to have both wind shear and a z component. It's probably an edge case, but I could imagine there might be times when it could be useful.

floris/heterogeneous_map.py
```diff
@@ -39,6 +41,7 @@ def __init__(
         x: NDArrayFloat,
         y: NDArrayFloat,
         speed_multipliers: NDArrayFloat,
+        z: NDArrayFloat = None,
```
misi9170, line 44: Is the reason this is after `speed_multipliers` to allow it to take a default (while `speed_multipliers` has no default)?

What if we reorder it (to `x`, `y`, `z`, `speed_multipliers`) and give `speed_multipliers` a default of 1?

floris/heterogeneous_map.py
```diff
@@ -50,20 +53,40 @@ def __init__(
         if not isinstance(speed_multipliers, (list, np.ndarray)):
             raise TypeError("speed_multipliers must be a numpy array or list")
 
+        # If z is provided, check that it is a list or numpy array
+        if z is not None:
+            if not isinstance(z, (list, np.ndarray)):
```
misi9170, line 58: Again, I don't think the nested `if` is necessary here... I would go for 

`if z is not None and not isinstance():`

floris/heterogeneous_map.py
```diff
@@ -50,20 +53,40 @@ def __init__(
...
         # Save the values
         self.x = np.array(x)
         self.y = np.array(y)
         self.speed_multipliers = np.array(speed_multipliers)
 
+        # If z is provided, save it as an np array
+        if z is not None:
+            self.z = np.array(z)
+        else:
+            self.z = None
```
misi9170, line 70: This is fine, but might be cleaner to have a single liner
`self.z = z if z is None else np.array(z)`

floris/heterogeneous_map.py
```diff
@@ -195,12 +235,49 @@ def get_heterogeneous_inflow_config(
...
             )
 
-        # Return heterogeneous_inflow_config
-        return {
-            "x": self.x,
-            "y": self.y,
-            "speed_multipliers": speed_multipliers_by_findex,
-        }
+        # Return heterogeneous_inflow_config with only x and y is z is not defined
+        if self.z is None:
```
misi9170, line 239: I think here, we could just have a single return statement that always returns the `"z": self.z` field, even if it is None. Would be more consistent and a simpler output.

misi9170 (commented):
Handful of small comments @paulf81 mostly on the error catches.

misi9170:
Added documentation looks good, thanks for adding that!

floris/floris_model.py
paulf81: I think if z is not in heterogeneous_inflow_config, then heterogeneous_inflow_config['z'] raises an error

floris/floris_model.py
paulf81: I started with a warning, but I couldn't think of a case where you'd want both, also I think the result is a little ambiguous, for example, say you apply a normal shear coefficient, but then use heterogeneity to have the shear inverted and go slower with height.  You will get some result, but then you don't have the shear layer implied by the exponent and that feels more like an error to me.

 It was really easy to imagine accidentally forgetting to turn off shear (I had done it!)

floris/heterogeneous_map.py
paulf81: I think it would work, but I think most of the code I've looked at expects that a 2d heterogeneous_inflow_config has no 'z' field

floris/heterogeneous_map.py
paulf81: I feel like I'm arguing every point, and I don't mean to, appreciate the suggestions!  I just somehow prefer the multi-line

floris/heterogeneous_map.py
paulf81: Yeah, that's it.  Admit it isn't a perfect solution but I think it is somehow correct that speed_multipliers has no default, but willing to discuss more

floris/floris_model.py
misi9170: Yeah, definitely easy to forget, but that's what the warning would be for right? Or is your feeling that the warnings are too easy to ignore?

floris/floris_model.py
misi9170: I'd have to check, but I think it will check the statements in order---it doesn't execute the second clause of the AND if the first is already false

floris/heterogeneous_map.py
paulf81: Done!

floris/floris_model.py
paulf81: I think I like it a shade better this way, but not so attached if you have a strong preference to one line

misi9170: Just checked, and it will do them in order, for instance:
```
x = {"a": 2}

if "b" in x and x["b"] == 2:
    print("Hello, world")
```
runs as expected (no printout, and no error raised)

floris/floris_model.py
paulf81: I guess it's more I'm wondering if it isn't an error, what I mean is, the shear layer will be a little unpredictable to me if both are applied.  I think of it like applying yaw and derating at the same time, it could work but we don't allow it because the correct result is unclear

floris/floris_model.py
paulf81: Done!

floris/heterogeneous_map.py
misi9170: Yeah, maybe that's right. Ok, hopefully people just use keyword arguments then and then can specify them in whatever order they please.

floris/floris_model.py
misi9170: Oh, I see. Yeah, Ok, that makes sense. Maybe we should just add to the error statement something like: "please include shear effects in `speed_multipliers`". At the moment, the wording of the error sounds like a warning ("may result in double-counting"). Since it won't run, there's no risk of it double-counting.

misi9170 (approved):

Issue #916, opened 2024-06-05
Add WRG file reader to FLORIS


paulf81:
# Add WRG file reader to FLORIS

WRG files, wind resource grids, are used in industrial practice for describing how wind resources vary over a given site.  WRGs represent wind resources by the parameters of Weibull distributions for given grid points and given wind directions sectors.  This implies that different points in the site will have different wind rose distributions.

### Information on WRG files:

https://backend.orbit.dtu.dk/ws/portalfiles/portal/116352660/WAsP_10_Help_Facility.pdf

## Proposed solution

We propose to develop new modules that:

- Read and hold the information embedded in a WRG file
- Develop a new WindRoseByTurbine class that holds a separate wind rose per turbine in a farm
- Modify FlorisModel (and children) to accept WindRoseByTurbine and provide  `get_expected_turbine_power()` and change how `get_expected_farm_power()` operates when WindRoseByTurbine is provided
- Investigate how best to perform layout optimizations under these conditions

@misi9170 @pjstanle

FYI: @ejsimley @rafmudaf @bayc

misi9170:
#919 addresses this

PR #919, opened 2024-06-05
Add WRG file support to FLORIS

paulf81:
# Add WRG file support to FLORIS

WRG files, wind resource grids, are used in industrial practice for describing how wind resources vary over a given site.  WRGs represent wind resources by the parameters of Weibull distributions for given grid points and given wind directions sectors.  This implies that different points in the site will have different wind rose distributions.

This (draft) pull request will add these capabilities and is being opened at the start of work to facilitate collaboration.

Envisioned steps (this will be updated as progress continues):

- [x] Develop script to make an example wrg file
- [x] Add `WindRoseWRG` class
- [x] Integration tests for `WindRoseWRG`
- [x] Update `FlorisModel` and children, to accept `WindRoseWRG`
- [x] Implement `get_expected_turbine_power` and update `get_expected_farm_power`
- [x] Tests of  `get_expected_turbine_power` and `get_expected_farm_power`
- [x] Update layout optimization to use `WindRoseWRG`?
- [x] Layout example
- [x] Develop methods to build `HeterogeneousMap` from `WrgReader`?
- [x] New examples of all of the above
- [x] Documentation of `WindRoseWRG`
- [x] Need to address @misi9170's point about the return_weighted_turbine_powers in the function `_get_farm_power`

## Related issue
#916


paulf81:
Hi @misi9170 and @pjstanle , I still have to update a few docs but on the whole most of the main work of this is done so I'm marking it ready for review if you'd like to take a look.

tests/wind_data_integration_test.py
```diff
@@ -693,9 +693,7 @@ def test_gen_heterogeneous_inflow_config():
 
     expected_result = np.array([[1.0, 1.0], [1.0, 1.0], [1.0, 1.0], [1.0, 1.0], [1.1, 1.2]])
     np.testing.assert_allclose(heterogeneous_inflow_config["speed_multipliers"], expected_result)
-    np.testing.assert_allclose(
-        heterogeneous_inflow_config["x"], heterogeneous_inflow_config["x"]
-    )
+    np.testing.assert_allclose(heterogeneous_inflow_config["x"], heterogeneous_inflow_config["x"])
```
misi9170, line 696: I don't think this line was added in this PR, just altered for formatting, but what is this line supposed to be testing?

tests/wind_data_integration_test.py
```diff
@@ -867,3 +853,70 @@ def test_read_csv_long_ti():
...
+    # Finally make sure that if set correctly no errors are raised
+    wind_speeds = np.array([6, 7])
+    wind_rose_1 = WindRose(wind_directions, wind_speeds, 0.06)
+    wind_roses = [wind_rose_0, wind_rose_1]
+    wind_rose_by_turbine = WindRoseByTurbine(
+        layout_x=[0, 0], layout_y=[0, 0], wind_roses=wind_roses
+    )
+
+    # Test the unpack method
+    (
```
misi9170, line 893: It seems more elegant to me to use `wind_rose_by_turbine.unpack()[0]` for this (rather than assigning many `_` variables)

floris/optimization/layout_optimization/layout_optimization_random_search.py
```diff
@@ -53,13 +53,26 @@ def _get_objective(
         layout_y,
         fmodel,
         yaw_angles=None,
-        use_value=False
+        use_value=False,
+        wind_resource_grid=None,
+        wind_resource_grid_wind_speeds = np.arange(0.0,25.0,2.0)
```
misi9170, line 58: There seems to be a lot of piping that has to happen to have this work; just for us to recommend using the HeterogeneousMap? Perhaps this and other comments around the two methods for layout optimization warrant some further discussion

floris/floris_model.py
```diff
@@ -558,10 +559,47 @@ def get_turbine_powers(self):
...
+        if len(np.shape(freq)) == 2:
+            return np.nansum(np.multiply(freq, turbine_powers), axis=0)
+        else:
+            return np.nansum(np.multiply(freq.reshape(-1, 1), turbine_powers), axis=0)
+
     def _get_farm_power(
         self,
         turbine_weights=None,
         use_turbulence_correction=False,
+        return_weighted_turbine_powers=False,
```
misi9170, line 602: I think I'd prefer if this was just a separate method---this flag pretty well directly alters the meaning of the method, in my opinion.

tests/wind_resource_grid_integration_test.py
misi9170, line 1: This seems to be more of a unit test than an integration test, in my opinion.

examples/examples_wind_resource_grid/001_generate_example_wrg.py
misi9170, line 1: Perhaps we should number this example differently, seeing as it's not actually an example that really demonstrates any FLORIS usage---it's just needed to set up the following examples (and not strictly needed, since the output wrg_example.wrg is included in the repo).

I'm not exactly sure what to label it, but possibly 000_ is fine (and then renumber the others starting at 001)?

misi9170 (changes requested):

examples/examples_wind_resource_grid/001_generate_example_wrg.py
```diff
@@ -0,0 +1,230 @@
...
+            # Get the A and k parameters for this sector
+            A, k = estimate_weibull(wind_speeds, freq_table[wd_idx, :])
+
+            # Write the A and k parameters
+            line = line + f"{int(A*10):4d}{int(k*100):5d}"
+
+        # Write the line to the file
+        f.write(line + "\n")
+
+# Echo the file contents in this cell
```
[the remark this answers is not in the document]
paulf81: Yeah, I agree, removing now

examples/examples_wind_resource_grid/001_generate_example_wrg.py
paulf81: Starting with 000 works for me, I'll make the change

tests/wind_resource_grid_integration_test.py
paulf81: Do I then just rename the file you think?

floris/optimization/layout_optimization/layout_optimization_random_search.py
paulf81: Definitely, do you have time today?

examples/examples_wind_resource_grid/001_generate_example_wrg.py
```diff
@@ -0,0 +1,230 @@
...
+with open("wrg_example.wrg", "r") as f:
+    print(f.read())
+
+# Show the plots
+plt.show()
```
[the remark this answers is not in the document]
paulf81: Done!

floris/wind_resource_grid.py
```diff
@@ -0,0 +1,407 @@
...
+        # Loop through the turbines and get the wind rose at each location
+        for i in range(len(layout_x)):
+            wind_rose = self.get_wind_rose_at_point(
+                layout_x[i], layout_y[i], wind_speeds=wind_speeds, fixed_ti_value=fixed_ti_value
+            )
+            wind_roses.append(wind_rose)
+
+        return WindRoseByTurbine(layout_x=layout_x, layout_y=layout_y, wind_roses=wind_roses)
+
+    def get_heterogeneous_map(
```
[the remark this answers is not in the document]
paulf81: Check the new version?

tests/wind_data_integration_test.py
paulf81: This line is no longer there

tests/wind_resource_grid_integration_test.py
paulf81: File is gone

paulf81:
@misi9170 @Bartdoekemeijer @pjstanle , this PR is getting closer to done, following some bigger revisions with @misi9170 , there are still a few last issues to go but at this point would welcome some feedback if you could have a look!

floris/optimization/layout_optimization/layout_optimization_random_search.py
paulf81: This is all removed

paulf81:
Note I updated the comments above, their is a failing test, it is the `test_upsample` function, but here I have changed the expected behavior.  The reason is in the old version, if say the original wind directions were 0, 30, 60 ... 330, this does in fact cover everything from 0-360, however if in the past you upsampled to 1 deg steps, the highest wind direction included would be 330.5. (and then 359.5 is covered from above), this is not desired.  I think I have a fix in but it changes the expected behavior some I think

paulf81:
ok @misi9170 , this is now ready to review, but could be a little easier after merging #943

misi9170:
@paulf81 , this looks good overall. One part I wasn't convinced by was adding the `subset_wind_speeds` method to `WindRose`, with the only current use case being for selecting a single wind speed in `WindRoseWRG.get_heterogeneous_wind_rose()`. I decided to try removing that new method on `WindRose` and instead implementing the steps directly in `WindRoseWRG.get_heterogeneous_wind_rose()`. 

It would be good to check that the examples are still running as you intend (the 003 example, in particular). If you'd like to revert the change I made, the commit prior to the change is f179558

floris/wind_data.py
```diff
@@ -2651,3 +2672,672 @@ def to_WindTIRose(
...
+        # The frequency is the difference in the cumulative distribution function
+        # at the edges
+        freq = cdf_edges[1:] - cdf_edges[:-1]
+
+        # Normalize the frequency
+        # TODO: This is perhaps not quite right, if the user only asks
+        # for wind speeds at say 8 and 10 m/s
+        # then the bins would be from 7 to 11 and so fairly there is
+        # distribution outside the range, but
+        # I think this is on the whole right -- PF
```
misi9170, line 2978: Does this need to be addressed still?

paulf81: I guess I think now it's right enough.  When I wrote this comment I was thinking renormalzing when we know certain bands are not included is wrong, but now I think it's just another interpretation.  Like, the Weibull distribution in theory goes on forever but it seems ok to say that we want the bins which we sampled to sum to 1.0.  I think that changes slightly the meaning but in a way which FLORIS expects, we don't really have a mechanism for wind speed bin weights that don't sum to 1.0.   So I think we could delete this TODO if you agree @misi9170

paulf81:
> @paulf81 , this looks good overall. One part I wasn't convinced by was adding the `subset_wind_speeds` method to `WindRose`, with the only current use case being for selecting a single wind speed in `WindRoseWRG.get_heterogeneous_wind_rose()`. I decided to try removing that new method on `WindRose` and instead implementing the steps directly in `WindRoseWRG.get_heterogeneous_wind_rose()`.
> 
> It would be good to check that the examples are still running as you intend (the 003 example, in particular). If you'd like to revert the change I made, the commit prior to the change is [f179558](https://github.com/NREL/floris/commit/f179558483605639eda90404f699aaaa908aac94).
> 
> I also tried moving the `WindRoseWRG` class to a separate file, to prevent wind_data.py from getting too long. I found that that was indeed possible without circular import issues, but in the end I decided it's best to stay on wind_data.py for now, since that is where all of the `WindDataBase` children are. However, we might want to consider creating a separate file for each in future (which I think may be possible without breaking changes if we place them all in a wind_data/ subdirectory of floris/, and create an appropriate **init**.py file with imports?)

This all works for me @misi9170 , I had originally thought to put in seperate files but indeed hit those import issues but nice they can be avoided.  This could be a good future fix?

floris/wind_data.py
misi9170: I think the current approach is correct, if I'm understanding everything correctly. However, I do think we could add a more detailed comment to avoid any confusion in future, something along the lines of:
```
# NOTE: The probability mass associated to each discrete wind speed (ws) is taken as the cumulative mass under the continuous Weibull distribution from ws-ws_step to ws+ws_step, where ws_step is the step between the provided wind_speeds.
```

In looking more carefully at this,

paulf81: I think enforcing even step sizes is good approach for now, and matches the requirements put in for wind direction in the upsample pull request.  I think even spacing is implicitly assumed so good to made it explicit and if we decide to relax it to allow un-even spacing we need to handle/test a number of cases, if there's a good reason to support it

misi9170: Done.

misi9170 (approved):

PR #924, opened 2024-06-18
Add Unified Momentum Model to operation models

jaimeliew1:
# Add MIT yaw correction model (3rd pass)

This pull request implements an interpretation of the yawed actuator disk as described in Heck _et al._ (2023):

The method assumes that $C_T'$ is invariant to yaw misalignment (an assumption still to be verified), and preforms a wind speed reduction based on the model equations. 

## Related issue

The structure of the code changes resembles those in the previous pull request (#638) , but now compatible with Floris V4. 

Another older pull request can be found here (#624)
 
A new operation model is added `mit_loss` (name to be decided).

I added the code used to calculate the MIT operational model to `floris/core/mit_models`, but please move this to wherever you think is best.
 
An example has been modified to show the operational models at different controller operation points. The added model is able to transition from below-rated to above-rated wind speeds in a realistic manner as shown in the following figure. I.e. The power output is "clipped" near rated wind speed, and is invariant to yaw far above rated wind speed. 
floris_power_loss

Relevant tests are still to be added.


paulf81:
Thank you @jaimeliew1 for this exciting contribution!  We'll start working on the review/merge soon!

rafmudaf:
A summary of changes I've made:
- Consolidated all of the model implementation to one file at `floris/core/turbine/mit_turbine.py`
- Arranged that module in the following order:
  1. Turbine operation module functions; these are the hooks to incorporate this model into the `power`, `thrust`, and `axial_induction` functions defined in the `BaseOperationModel` interface
  2. Iteration functions
  3. The `BaseOperationModel` subclass for the MIT Turbine model
  4. Functions implementing the model as described in the paper
- Added a brief description of the model, some equivalent figures from the paper, and a link to the paper in the docs

Notes / suggestions:
- I propose to change the name from "mit-loss" / "MITTurbine" to something like "YawCorrectedADM" or "YawAwareADM"; to a user who has not seen this pull request, the option for a "mit-loss" model will not have much meaning, so it would be helpful to introduce a more descriptive name
- The implementation is elegantly implemented and easy to map back to the model described in the paper. However, the programming paradigm is different than the rest of FLORIS and it may be challenging to maintain in the future. I don't recommend to change the implementation, but this difference is worth noting for our future reference. If there are any maintenance challenges, it might be worthwhile to refactor to match the rest of the FLORIS paradigm.

rafmudaf:
@jaimeliew1 I wasn't able to understand where [mit_rotor_axial_induction](https://github.com/jaimeliew1/floris/blob/jaime/floris/core/turbine/mit_turbine.py#L29) and [mit_rotor_velocity_yaw_correction](https://github.com/jaimeliew1/floris/blob/jaime/floris/core/turbine/mit_turbine.py#L47) are described in the paper. Could you explain this portion a bit more?

misi9170:
After discussion with the developers of the model, the model will be referred to as the Unified Momentum Model going forward. This is the simpler implementation based on [Heck et al., 2023](https://www.cambridge.org/core/journals/journal-of-fluid-mechanics/article/modelling-the-induction-thrust-and-power-of-a-yawmisaligned-actuator-disk/3A34FC48A6BC52A78B6D221C13F4FC3A) (which makes simplifications assuming thrust coefficients lower than approximately 0.9) rather than the full implementation of [Liew et al., 2024a](https://iopscience.iop.org/article/10.1088/1742-6596/2767/3/032018) and [Liew et al., 2024b](https://www.nature.com/articles/s41467-024-50756-5).

misi9170 (approved):
This looks good to me. The only thing we should do is possibly clear up the location of the example once we merge in #832 . So, I'd say we merge #832 first; align the example here; and then merge this PR.

misi9170:
@rafmudaf I've now aligned this with the develop branch so it's ready to go in. In doing so, I:
- moved the tests for the `UnifiedMomentumModel` to their own file (rather than being on turbine_operation_models_unit_test.py)
- Moved the code from the added examples_turbine/004..py to instead be in examples_operation_models/001..py. Running that example now produces 
image
which includes both the `ControllerDependentTurbine` (#832) and `UnifiedMomentumModelTurbine`.

PR #930, opened 2024-06-25
[BUGFIX] Fix wind rose plots

paulf81:
# Bugfix for WindRose plot

The `plot` function of the WindRose class had a bug that plotted the percentage of wind speeds from 0: ws_idx, when this should've been include (:ws_idx+1).  This pull request fixes that error.

Further, a change allows there to be only one wind direction, although this creates an odd looking polar plot, it's probably better to allow that case.

A temporary example (plan to delete before merging), test_plot.py is inclued to show the behaviors is improved.


paulf81:
@ejsimley , maybe we should make a matching change within WindTIRose?

ejsimley:
> @ejsimley , maybe we should make a matching change within WindTIRose?

Yes, it looks like that same line is in WindTIRose too.

paulf81:
> > @ejsimley , maybe we should make a matching change within WindTIRose?
> 
> Yes, it looks like that same line is in WindTIRose too.

Ok done I think!  @misi9170 if you wouldn't mind to check my work I wouldn't mind getting this into develop, pretty small change in the end,

misi9170 (approved):
This looks good to me.
@paulf81 just remember to remove the test_plot.py example before merging

paulf81:
> This looks good to me. @paulf81 just remember to remove the test_plot.py example before merging

Done!  Just going to check for all the green checks then merge

Issue #938, opened 2024-07-08
Allow more recent versions of `coloredlogs` as a dependency to FLORIS


pjireland:
# Allow more recent versions of `coloredlogs` as a dependency to FLORIS
In `floris/setup.py`, `coloredlogs` is currently pinned to version `~=10.0`. This version is quite old now (release was May 2018). 

## Proposed solution
Recommend either switching to a newer version (the latest is `15.0.1`, released in June 2021) or allowing any version greater than or equal to `10.0`.

## Additional context
I'm trying to install FLORIS within an environment that requires `coloredlogs==15.0.1`. I therefore need to create my own branch of FLORIS (with the `10.0` `coloredlogs` dependency relaxed) to be able to do so.

paulf81:
Thanks for raising this issue @pjireland , I've opened a pull request (#939) to look at updating FLORIS' requirements

pjireland:
> Thanks for raising this issue @pjireland , I've opened a pull request (#939) to look at updating FLORIS' requirements

Great - thanks, @paulf81!

paulf81:
Resolved in #939

PR #939, opened 2024-07-08
Update coloredlogs requirement

paulf81:
# Update requirements

FLORIS uses the compatible release clause (`~=`) when specificying versioning (c.f. https://peps.python.org/pep-0440/#compatible-release).  This PR updates to newer versions of certain requirements that have had major releases to catch up with: numpy and colored-logs.

~It further adds macos to list of OS to try in CI~

## Related issue
Issue #938

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ REQUIRED = [
     "attrs",
     "pyyaml~=6.0",
     "numexpr~=2.0",
-    "numpy~=1.20",
+    "numpy~=2.0",
     "scipy~=1.1",
 
     # tools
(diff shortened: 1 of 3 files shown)


misi9170:
Linking #748

setup.py
```diff
@@ -18,7 +18,7 @@
     "attrs",
     "pyyaml~=6.0",
     "numexpr~=2.0",
-    "numpy~=1.20",
+    "numpy~=2.0",
```
pjireland, line 21: Since `numpy` 2.0 is a very new release (June 16, 2024), it would be helpful for me (and I suspect many other users) to not yet require version 2.

misi9170 (approved):

Issue #936, opened 2024-07-04
`test_random_search_layout_opt` is failing locally on my machine but not on github


paulf81:
# `test_random_search_layout_opt` is failing locally on my machine but not on github

I think this must be a difference in package version, or os, but on the version of develop (current July 4), running `test_random_search_layout_opt` produces the error:

```
/Users/pfleming/Projects/FLORIS/floris/tests/reg_tests/random_search_layout_opt_regression_test.py::test_random_search_layout_opt failed: sample_inputs_fixture = <tests.conftest.SampleInputs object at 0x7fefe38b8640>

    def test_random_search_layout_opt(sample_inputs_fixture):
        """
        The SciPy optimization method optimizes turbine layout using SciPy's minimize method. This test
        compares the optimization results from the SciPy layout optimization for a simple farm with a
        simple wind rose to stored baseline results.
        """
        sample_inputs_fixture.core["wake"]["model_strings"]["velocity_model"] = VELOCITY_MODEL
        sample_inputs_fixture.core["wake"]["model_strings"]["deflection_model"] = DEFLECTION_MODEL
    
        boundaries = [(0.0, 0.0), (0.0, 1000.0), (1000.0, 1000.0), (1000.0, 0.0), (0.0, 0.0)]
    
        fmodel = FlorisModel(sample_inputs_fixture.core)
        wd_array = np.arange(0, 360.0, 5.0)
        ws_array = 8.0 * np.ones_like(wd_array)
    
        wind_rose = WindRose(
            wind_directions=wd_array,
            wind_speeds=ws_array,
            ti_table=0.1,
        )
        D = 126.0 # Rotor diameter for the NREL 5 MW
        fmodel.set(
            layout_x=[0.0, 5 * D, 10 * D],
            layout_y=[0.0, 0.0, 0.0],
            wind_data=wind_rose
        )
    
        layout_opt = LayoutOptimizationRandomSearch(
            fmodel=fmodel,
            boundaries=boundaries,
            min_dist_D=5,
            seconds_per_iteration=1,
            total_optimization_seconds=1,
            use_dist_based_init=False,
            random_seed=0,
        )
        sol = layout_opt.optimize()
        optimized_aep = sol[0]
        locations_opt = np.array([sol[1], sol[2]])
    
        if DEBUG:
            print(locations_opt)
            print(optimized_aep)
    
>       assert_results_arrays(locations_opt, locations_baseline_aep)

[...]

PR #940, opened 2024-07-09
[BUGFIX] Create reproducible version of random search layout optimizer for testing

misi9170:
As reported in #936 , the regression tests for the random search layout optimizer can _sometimes_ fail. I believe that the cause for this is that the optimizer is run for one second during the test; however, depending on the machine, it may take more or less steps over that period.

This PR alters the `LayoutOptimizationRandomSearch` code to make it modular, and then adds a debugging version of the optimization routine that will take a fixed number of steps (in particular, 100 random perturbations per generation, 1 individual per generation, and 2 generations).

This should enable reproducible regression test results.


misi9170:
Note that after the changes, I still get the same result (as on the develop branch) running examples/examples_layout_optimization/003_genetic_random_search.py if I fix the random seed to the `LayoutOptimizationRandomSearch`.

PR #943, opened 2024-07-10
[BUGFIX] Circular upsampling across wind directions

paulf81:
# Fix upsample

(Note this is cleaned up version of #942 from a git history perspective, closing that one in favor of this one)

The upsample function in wind_data.py (both `WindRose` and `WindTIRose`) has a subtle issue.  Because the current implementation assumes the same range of wind directions and wind speeds in the resampled rose as in the original, there can be a gap in the range implied by the combination of the original wind direction steps and bin size.  The issue is illustrated by the code snippet:

```
"""Example X: Demo Wind Sample Update Issue (Temporary)

Demonstrate the issue with upsampling a wind rose

"""


import numpy as np

from floris import (
    WindRose,
)

# Say that you start with a wind rose with 10-degree wind direction bins and 2 m/s wind speed bins
wind_directions = np.arange(0, 360, 10.0)
wind_speeds = np.arange(4, 20, 2.0)

# Uniform frequency
freq_table = np.ones((len(wind_directions), len(wind_speeds)))
freq_table = freq_table / np.sum(freq_table)

wind_rose = WindRose(
    wind_speeds=wind_speeds,
    ti_table=0.06,
    freq_table=freq_table,
)

# Upsample the wind rose to 1-degree bins
wind_rose_upsampled = wind_rose.upsample(wd_step=2.0)

# Show the original and resample wind roses
fig, axarr = plt.subplots(1, 2, figsize=(12, 5), subplot_kw={"polar": True})
wind_rose.plot(ax=axarr[0])
axarr[0].set_title("Original Wind Rose")

wind_rose_upsampled.plot(ax=axarr[1])
axarr[1].set_title("Upsampled Wind Rose")

plt.show()

```

Which produces:

image

This pull request fixes the issue by setting the new wind directions to cover the complete span of the covered directions.  This change is not breaking per se, but does require changing some of the tests to match the new expected behavior.  For example, a previous test was:

```
# Test interpolating TI along the wind direction axis
wind_directions = np.array([270, 280])
wind_speeds = np.array([6, 7])
ti_table = np.array([[0.06, 0.06], [0.07, 0.07]])
wind_rose = WindRose(wind_directions, wind_speeds, ti_table=ti_table)

wind_rose_resample = wind_rose.upsample(wd_step=5.0, ws_step=1.0, inplace=False)

np.testing.assert_allclose(wind_ti_rose_resample.wind_directions, [270, 275, 280])
```

But note this result only extends to 267.5, not the 265 of the original version.  The new test concludes with:

```
np.testing.assert_allclose(wind_rose_resample.wind_directions, [267.5, 272.5, 277.5, 282.5])
```

With the improvements, above figure is now:

image

## Related issue
This pull request takes into more focus work originally done within #919 but we thought should be its own pull request to make the changes clearer

wind_data.py

### DIFF
--- a/examples/examples_layout_optimization/002_optimize_layout_with_heterogeneity.py
+++ b/examples/examples_layout_optimization/002_optimize_layout_with_heterogeneity.py
@@ -26,10 +26,10 @@ fmodel = FlorisModel("../inputs/gch.yaml")
 
 # Setup 2 wind directions (due east and due west)
 # and 1 wind speed with uniform probability
-wind_directions = np.array([270.0, 90.0])
+wind_directions = np.array([90.0, 270.0])
 n_wds = len(wind_directions)
-wind_speeds = [8.0] * np.ones_like(wind_directions)
-turbulence_intensities = 0.06 * np.ones_like(wind_directions)
+wind_speeds = np.array([8.0])
+
 # Shape frequency distribution to match number of wind directions and wind speeds
 freq_table = np.ones((len(wind_directions), len(wind_speeds)))
 freq_table = freq_table / freq_table.sum()
@@ -106,7 +106,7 @@ ax = plt.gca()
 fig = plt.gcf()
 sm = ax.tricontourf(x_locs, y_locs, speed_multipliers[0], cmap="coolwarm")
 fig.colorbar(sm, ax=ax, label="Speed multiplier")
-ax.legend(["Initial layout", "Optimized layout", "Optimization boundary"])
+ax.legend(["Optimization boundary", "Initial layout", "Optimized layout" ])
 ax.set_title("Geometric yaw disabled")
 
 
@@ -144,7 +144,7 @@ ax = plt.gca()
 fig = plt.gcf()
 sm = ax.tricontourf(x_locs, y_locs, speed_multipliers[0], cmap="coolwarm")
 fig.colorbar(sm, ax=ax, label="Speed multiplier")
-ax.legend(["Initial layout", "Optimized layout", "Optimization boundary"])
+ax.legend(["Optimization boundary", "Initial layout", "Optimized layout"])
 ax.set_title("Geometric yaw enabled")
 
 print(
--- a/examples/examples_layout_optimization/002_optimize_layout_with_heterogeneity.py
+++ b/examples/examples_layout_optimization/002_optimize_layout_with_heterogeneity.py
@@ -106,7 +106,7 @@ ax = plt.gca()
 fig = plt.gcf()
 sm = ax.tricontourf(x_locs, y_locs, speed_multipliers[0], cmap="coolwarm")
 fig.colorbar(sm, ax=ax, label="Speed multiplier")
-ax.legend(["Optimization boundary", "Initial layout", "Optimized layout" ])
+ax.legend(["_Optimization boundary", "Initial layout", "Optimized layout" ])
 ax.set_title("Geometric yaw disabled")
 
 
@@ -144,7 +144,7 @@ ax = plt.gca()
 fig = plt.gcf()
 sm = ax.tricontourf(x_locs, y_locs, speed_multipliers[0], cmap="coolwarm")
 fig.colorbar(sm, ax=ax, label="Speed multiplier")
-ax.legend(["Optimization boundary", "Initial layout", "Optimized layout"])
+ax.legend(["_Optimization boundary", "Initial layout", "Optimized layout"])
 ax.set_title("Geometric yaw enabled")
 
 print(
--- a/floris/wind_data.py
+++ b/floris/wind_data.py
@@ -452,25 +452,50 @@ class WindRose(WindDataBase):
                 "Available methods are 'linear' and 'nearest'"
             )
 
+        # First establish the current ws_step and wd_step
+        if len(self.wind_speeds) >= 2:
+            ws_step_current = self.wind_speeds[1] - self.wind_speeds[0]
+        else:  # wind rose will have only a single wind speed, and we assume a ws_step of 1
+            ws_step_current = 1.0
+
+        if len(self.wind_directions) >= 2:
+            wd_step_current = self.wind_directions[1] - self.wind_directions[0]
+        else:  # wind rose will have only a single wind direction, and we assume a wd_step of 1
+            wd_step_current = 1.0
+
         # If either ws_step or wd_step is None, set it to the current step
         if ws_step is None:
-            if len(self.wind_speeds) >= 2:
-                ws_step = self.wind_speeds[1] - self.wind_speeds[0]
-            else:  # wind rose will have only a single wind speed, and we assume a ws_step of 1
-                ws_step = 1.0
+            ws_step = ws_step_current
         if wd_step is None:
-            if len(self.wind_directions) >= 2:
-                wd_step = self.wind_directions[1] - self.wind_directions[0]
-            else:  # wind rose will have only a single wind direction, and we assume a wd_step of 1
-                wd_step = 1.0
+            wd_step = wd_step_current
 
-        # Set up the new wind direction and wind speed bins
-        new_wind_directions = np.arange(
-            self.wind_directions[0], self.wind_directions[-1] + wd_step / 2.0, wd_step
-        )
-        new_wind_speeds = np.arange(
-            self.wind_speeds[0], self.wind_speeds[-1] + ws_step / 2.0, ws_step
-        )
+        # Set up the new wind directions
+        wd_range_min_current = np.min(self.wind_directions) - wd_step_current / 2.0
+        wd_range_max_current = np.max(self.wind_directions) + wd_step_current / 2.0
+
+        if wd_range_max_current > 360:
+            # TODO: This could probably be more clear
+            raise ValueError(
+                "Cannot upsample wind rose for case when wind directions are defined"
+                " such that 0 degrees is included by bins to the left of 0 degrees. "
+            )
+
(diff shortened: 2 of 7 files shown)


floris/wind_data.py
```diff
@@ -1551,7 +1638,7 @@ def plot(
             rects = []
             freq_table_sub = freq_table[wd_idx, :].flatten()
             for var_idx, ws in reversed(list(enumerate(var_bins))):
-                plot_val = freq_table_sub[:var_idx + 1].sum()
+                plot_val = freq_table_sub[: var_idx + 1].sum()
```
misi9170, line 1641: I presume this is something from ruff-format? It's a strange one, I'm not a fan, but so be it...

floris/wind_data.py
```diff
@@ -452,25 +452,51 @@ def upsample(self, wd_step=None, ws_step=None, method="linear", inplace=False):
                 "Available methods are 'linear' and 'nearest'"
             )
 
+        # First establish the current ws_step and wd_step
+        if len(self.wind_speeds) >= 2:
+            ws_step_current = self.wind_speeds[1] - self.wind_speeds[0]
```
misi9170, line 457: Do we need something that checks that the `wind_speeds` and `wind_directions` are actually evenly spaced? It seems possible that a user could specify them with irregular intervals (probably an edge case, but could happen)? Perhaps we could just add a check to the initialization of the roses to ensure even spacing?

floris/wind_data.py
```diff
@@ -452,25 +452,51 @@ def upsample(self, wd_step=None, ws_step=None, method="linear", inplace=False):
...
+        # Identify the covered range of wind directions
+        wd_range_min_current = np.min(self.wind_directions) - wd_step_current / 2.0
+        wd_range_max_current = np.max(self.wind_directions) + wd_step_current / 2.0
+
+        # Look for unlikely case where for example wind directions are 8, 28, ... 358
+        if wd_range_max_current > 360:
+            raise ValueError(
+                "Cannot upsample wind rose for case when wind directions are defined"
+                " such that 0 degrees is included by bins to the left of 0 degrees. "
+            )
```
misi9170, line 481: Should we add a TODO note here to add handling for this case? It does seem pretty unusual though, unlikely to come up very often.

floris/wind_data.py
```diff
@@ -452,25 +452,51 @@ def upsample(self, wd_step=None, ws_step=None, method="linear", inplace=False):
...
+
+        # Identify the new minimum wind direction
+        wd_min_new = wd_range_min_current + wd_step / 2.0
+        wd_max_new = wd_range_max_current - wd_step / 2.0
```
misi9170, line 485: Might be worth checking here that `wd_step <= wd_step_current` and same with `ws_step`? I.e., enforce actual upsampling and not downsampling? We could raise an error that points users to the `downsample` method if needed.

floris/wind_data.py
```diff
@@ -452,25 +452,51 @@ def upsample(self, wd_step=None, ws_step=None, method="linear", inplace=False):
...
+        ws_range_min_current = np.min(self.wind_speeds) - ws_step_current / 2.0
+        ws_range_max_current = np.max(self.wind_speeds) + ws_step_current / 2.0
+        ws_min_new = ws_range_min_current + ws_step / 2.0
+        ws_max_new = ws_range_max_current - ws_step / 2.0
+
+        # Force the new ws_min to 0 if negative
+        if ws_min_new < 0:
+            ws_min_new = 0.0
+
+        new_wind_speeds = np.arange(ws_min_new, ws_max_new + ws_step / 2.0, ws_step)
```
misi9170, line 499: Could this be strange if `ws_min_new` gets forced to zero in the lines above? All bins would be shifted? Perhaps we need to create `new_wind_speeds` by stepping backwards from `ws_max_new` rather than forwards from `ws_min_new`?

floris/wind_data.py
```diff
@@ -482,48 +508,61 @@ def upsample(self, wd_step=None, ws_step=None, method="linear", inplace=False):
...
-        # and expand the matrices accordingly (this avoids interpolation errors)
-        if len(wind_speed_column) == 1:
-            wind_speed_column = np.array(
-                [wind_speed_column[0] - 1, wind_speed_column[0], wind_speed_column[0] + 1]
-            )
-            ti_matrix = np.hstack((ti_matrix, ti_matrix[:, 0][:, None], ti_matrix[:, 0][:, None]))
-            freq_matrix = np.hstack(
-                (freq_matrix, freq_matrix[:, 0][:, None], freq_matrix[:, 0][:, None])
+        # Make sure everything sorted by wind direction and wind speed
+        # This should be the case but it's possible user can pass in unsorted data
```
misi9170, line 512: Again, the user could also pass in data with irregular spacing, but I don't think we currently can handle that?

floris/wind_data.py
```diff
@@ -482,48 +508,61 @@ def upsample(self, wd_step=None, ws_step=None, method="linear", inplace=False):
...
+            value_matrix = np.vstack((value_matrix[0, :], value_matrix))
+
+        # Add the max range to the wind direction column
+        wind_direction_column = np.append(wind_direction_column, wd_range_max_current)
+        ti_matrix = np.vstack((ti_matrix, ti_matrix[-1, :]))
+        freq_matrix = np.vstack((freq_matrix, freq_matrix[-1, :]))
+        if self.value_table is not None:
+            value_matrix = np.vstack((value_matrix, value_matrix[-1, :]))
+
+        # Pad out the wind speeds
```
misi9170, line 540: This comment is redundant to the comment on line 525 (maybe just mention wind directions on line 525)

floris/wind_data.py
```diff
@@ -482,48 +508,61 @@ def upsample(self, wd_step=None, ws_step=None, method="linear", inplace=False):
...
+        if self.value_table is not None:
+            value_matrix = np.vstack((value_matrix[0, :], value_matrix))
+
+        # Add the max range to the wind direction column
+        wind_direction_column = np.append(wind_direction_column, wd_range_max_current)
+        ti_matrix = np.vstack((ti_matrix, ti_matrix[-1, :]))
+        freq_matrix = np.vstack((freq_matrix, freq_matrix[-1, :]))
+        if self.value_table is not None:
+            value_matrix = np.vstack((value_matrix, value_matrix[-1, :]))
+
```
misi9170, line 539: Would it be cleaner to use np.concatenate() rather than np.append so that the min and the max can be added in the same line?

misi9170 (approved):
@paulf81 thanks for the thorough PR description; that really helps make it clear what the bug was and how this works!

I've put in a few comments for you to consider, but none of them is critical I think, so I'm going ahead and approving this to ensure your other workflows can proceed. Feel free to address my comments as you see fit and merge when you're ready.

floris/wind_data.py
paulf81: Done

floris/wind_data.py
paulf81: Added above and to TI rose, only doubt I have is if we need such strictness for TI, but probably it will help avoid bugs where we use the first elements for step size

floris/wind_data.py
paulf81: Added below

floris/wind_data.py
paulf81: I think it's ok, there's not that many ways to set the wind speeds, and less that include 0 so I wrote a few test cases into tests and couldn't see how this gets us into trouble

floris/wind_data.py
paulf81: Fixed now

floris/wind_data.py
paulf81: Yeah, I could see that, I guess I'll see if you see this as should do, or fine to leave as is?

floris/wind_data.py
paulf81: Exact, we can take it out if you want

floris/wind_data.py
ejsimley: The requirement of being evenly spaced should probably be added to the docstring argument descriptions.

paulf81: Adding now

ejsimley (approved):
This looks like the right approach to me, and I agree, very helpful description of the problem and solution.

paulf81:
@misi9170  and @ejsimley , sorry to do this, I had to make a few more medium-substantial changes, dealing with the wind direction range and sorting, I kept finding small bugs, and finally decided to pull out a few smaller functions to do the trickier bits so I can test a lot of possibilities directly.  I think it was helpful but changed a bit the code so would appreciate a re-review, feel free to merge if happy!

examples/examples_layout_optimization/002_optimize_layout_with_heterogeneity.py
```diff
@@ -26,10 +26,10 @@
 
 # Setup 2 wind directions (due east and due west)
 # and 1 wind speed with uniform probability
-wind_directions = np.array([270.0, 90.0])
+wind_directions = np.array([90.0, 270.0])
 n_wds = len(wind_directions)
-wind_speeds = [8.0] * np.ones_like(wind_directions)
-turbulence_intensities = 0.06 * np.ones_like(wind_directions)
```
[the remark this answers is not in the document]
misi9170: I've convinced myself that this was essentially a bug before, and that it's fine to remove these. This is still a small change to the user interface, but necessary for the bugfix.

misi9170 (approved):
This looks OK to me. @ejsimley , if you're happy, I can get this merged.

PR #946, opened 2024-07-15
Wind direction resampling 2

misi9170:
I recently approved and merged #943, but then noticed after the merger that, when I switched double uses of `np.append()` to `np.concatenate()`, I hadn't propagated the change throughout the file. 

This small PR addresses that issue and makes the use of `np.concatenate()` consistent throughout. It's not a bugfix, as the code runs fine, but is a very minor enhancement.

@paulf81 , I have left comment that it would be good to get your opinion on, on a small inconsistency between the `upsample()` methods of `WindRose` and `WindTIRose` (again, something I should probably have caught when reviewing #943):

In `WindTIRose.upsample()`, there is a code block:
```python
        # If wd_range_min_current is less than 0, then pad the wind_direction column with
        # that value + 360 and expand the matrices accordingly (this avoids interpolation errors)
        if wd_range_min_current < 0:
            # Pad wind direction column with min_wd + 360
            wind_direction_column = np.append(
                wind_direction_column, np.min(self.wind_directions) + 360.0
            )

            # Pad the remaining with the appropriate value
            freq_matrix = np.concatenate((freq_matrix, freq_matrix[0, :, :][None, :, :]), axis=0)
            if self.value_table is not None:
                value_matrix = np.concatenate(
                    (value_matrix, value_matrix[0, :, :][None, :, :]), axis=0
                )
```
but this code does not exist in `WindRose.upsample()`. Is that expected, and to do with the difference between the two classes? If not, is it redundant in `WindTIRose.upsample()` or missing in `WindRose.upsample()`?

### DIFF
--- a/floris/wind_data.py
+++ b/floris/wind_data.py
@@ -1656,7 +1656,7 @@ class WindTIRose(WindDataBase):
                     (value_matrix, value_matrix[0, :, :][None, :, :]), axis=0
                 )
 
-        # Grid wind directions, wind speeds and turbulence intensities to match the 
+        # Grid wind directions, wind speeds and turbulence intensities to match the
         # freq_matrix when flattened
         wd_grid, ws_grid, ti_grid = np.meshgrid(
             wind_direction_column, wind_speed_column, turbulence_intensity_column, indexing="ij"
--- a/floris/wind_data.py
+++ b/floris/wind_data.py
@@ -237,7 +237,7 @@ class WindRose(WindDataBase):
         # If heterogeneous_inflow_config_by_wd is not None, then create a HeterogeneousMap object
         # using the dictionary
         if heterogeneous_inflow_config_by_wd is not None:
-            # TODO: In future, add deprectation warning for this parameter here
+            # TODO: In future, add deprecation warning for this parameter here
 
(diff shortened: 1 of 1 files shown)


paulf81:
This is redundant in the WindTIRose case, it is now implemented within this block (starting at 1585):

```
  # In the alternative case, where the wind directions cover the full range
        # ie, 0, 10, 20 30, ...350, then need to place 0 at 360 and 350 at -10
        # to cover all interpolations
        else:
            # Pad wind direction column with min_wd + 360
            wind_direction_column = np.concatenate(
                (
                    [np.max(self.wind_directions) - 360.0],
                    wind_direction_column,
                    [np.min(self.wind_directions) + 360.0],
                )
            )

            # Pad the remaining with the appropriate value
            freq_matrix = np.vstack(
                (freq_matrix[-1, :, :][None, :, :], freq_matrix, freq_matrix[0, :, :][None, :, :])
            )
            if self.value_table is not None:
                value_matrix = np.vstack(
                    (
                        value_matrix[-1, :, :][None, :, :],
                        value_matrix,
                        value_matrix[0, :, :][None, :, :],
                    )
                )

```

Nice catch and sorry for the mistake ( I think mostly harmless, it's basically adding endpoints twice to the interpolation matrix) but the tests will confirm, I will delete and retest now

Issue #951, opened 2024-07-26
Issues when running FLORIS 4.1.1


misi9170:
### Discussed in https://github.com/NREL/floris/discussions/950

<div type='discussions-op-text'>

<sup>Originally posted by **prith-gs** July 26, 2024</sup>
Hello,

I'm trying to run example 01 with FLORIS 4.1.1 and I don't get this error in FLORIS 3.5 and I'm not sure why I get this error.

Here's the code I'm running where I am reading the input yaml file using FlorisModel function in line 9 and I'm using a function I found online in line 4 and 5.
![image](https://github.com/user-attachments/assets/0611306b-942f-49a4-b70e-d73f9309849f)

Line 4 and 5 executes properly and has no issues in reading the yaml file and I can see the variable 'data' as well
![image](https://github.com/user-attachments/assets/bc85e0b0-94de-4d08-bac9-0d7ead5790e4)

Here's what's stored in the variable data
![image](https://github.com/user-attachments/assets/5b85766b-9c78-43f0-bcc1-1a4ee7f8ac85)

But line 9 gives me the following error and I'm not clear why I'm getting this error
![image](https://github.com/user-attachments/assets/61b0cee3-00ba-4d9d-9944-89ea28406b58)

Any help is much appreciated.
Thanks!</div>

misi9170:
Tagging @prith-gs and @HD7812

misi9170:
Bringing over comments from discussion:

Comment by @HDW7812:
Hi, I'm also experiencing this problem, although I don't know exactly what's causing it, but I have a feeling it's a conflict with a previously installed package, so the turbine_library_path is incorrect. I create a new envs (without pip install floris), then run the demo, it works. I hope it can help you!
If you are in a hurry to use it, you can try this method. Wish someone could answer you more professionally.

Comment from @misi9170:
Hi @prith-gs , 

If you are using this within a conda environment, could you share the list of packages you have installed? As @HDW7812 pointed out, this could be something to do with package versions in your environment, especially if you are using the same environment you were running v3.5 with.

As also pointed out by @HDW7812 , starting with a fresh conda environment and a fresh pip install of FLORIS may solve your issue. 

I'm going to convert this discussion to an issue in the meantime.

Cheers,
Misha

prith-gs:
Hi @misi9170,

I tried creating a new environment and install floris again and I'm still getting the same error. 
I'm using miniconda and here's the list of packages I have installed in the new environment I created.

[...]

PR #953, opened 2024-07-30
Add back in explicit windows and mac testing

paulf81:
This is probably only a draft pull request, I was curious to see if it could help shed light on on #951.  Adding windows os (and mac) back into CI OS testing matrix

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yaml
+++ b/.github/workflows/continuous-integration-workflow.yaml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.8", "3.9", "3.10", "3.11"]
-        os: [ubuntu-latest] #, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
       fail-fast: False
     env:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
--- a/.github/workflows/continuous-integration-workflow.yaml
+++ b/.github/workflows/continuous-integration-workflow.yaml
@@ -10,6 +10,7 @@ jobs:
       matrix:
         python-version: ["3.8", "3.9", "3.10", "3.11"]
         os: [ubuntu-latest, macos-latest, windows-latest]
+        # Empty change to make the tests run
       fail-fast: False
     env:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
--- a/floris/type_dec.py
+++ b/floris/type_dec.py
@@ -158,6 +158,7 @@ def convert_to_path(fn: str | Path) -> Path:
     print("inspect.stack()[0] ", inspect.stack()[0])
     print("inspect.stack()[1] ", inspect.stack()[1])
     print("inspect.stack()[-1] ", inspect.stack()[-1])
+    print("inspect.stack() ", inspect.stack())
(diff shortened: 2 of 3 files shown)


misi9170:
@paulf81 , looks like you're right that this found the same issue as raised in #951 ...

paulf81:
Closing this one as seems like over-responding to one particular incident and greatly increasing our test count

rafmudaf:
fwiw I think it's worth including Mac and Windows in the test matrix even if it's only for a subset. For example, you could run the example suite on all three systems but run the tests only on ubuntu. It would just be good to know that something is running on each OS.

misi9170:
> fwiw I think it's worth including Mac and Windows in the test matrix even if it's only for a subset. For example, you could run the example suite on all three systems but run the tests only on ubuntu. It would just be good to know that something is running on each OS.

This feels like a reasonable compromise (examples only on ubuntu, since those take a while, and tests on all three OSs, since those are faster). I'll reopen the PR and make those updates.

misi9170:
After readding macOS and windows tests, it appears that the only failing tests are on Windows using python 3.9; and moreover, they are all related to using parallel operations (most are explicit, but note that `LayoutOptimizationRandomSearch` also defaults to parallelized operation). Evidently, everything works fine on python 3.10 and higher on Windows. 

So, there are a couple of options here: 
- Work on creating a fix so that the tests pass with python 3.9 and windows
- Drop support for python 3.9 (which will be at end of life later this year)

rafmudaf:
Just a heads up @misi9170 that a description of the issue is at https://github.com/NREL/floris/issues/951#issuecomment-2484057737. Based on their response, it looks like like @RHammond2 agrees with requiring Python 3.10+. I could go either way on that.

In addition to the suggestions above, here are some more ways around this:
- Somehow require Python 3.10+ for Windows only
- Leave #951 open with a "won't fix" tag so that other users will be able to understand the problem if it comes up (in this case, we'd skip or remove the test)

paulf81:
I prefer dropping 3.9, it's kind of a pain for reasons beyond those in this issue.  Sometimes tests crash just for 3.9 and it's because you need to add that `from __future__ import annotations` bit for 3.9 even though all other versions don't require it.

RHammond2:
@rafmudaf, yes, I fully recommend moving to 3.10+. I've already moved a couple projects to 3.10+ at this point because 3.9 is out the door this year anyway, and it was causing some nuisance issues that were hard to resolve. Also, 3.10 introduces `match` statements, which can be a bit more robust than `if/elif/else` blocks: https://docs.python.org/3/tutorial/controlflow.html#match-statements and I haven't looked back.

misi9170:
Thanks all. Since this issue is not critical, but 3.9 is not yet at end-of-life, I propose that we:
- Add a warning to the documentation that support for python 3.9 will be dropped within the next 6 months
- Open a separate PR to drop 3.9 support, which removes 3.9 from the testing matrix but also removes the `from __future__ import annotations` line that @paulf81 mentioned and possibly moves to the `match`/`case` structure that @RHammond2 mentioned for model selections and other appropriate areas.
- Wait to merge both the new PR dropping 3.9 support and this PR until roughly August 2025.

I've also added this comment to #951 .

misi9170:
Closing this in favor of https://github.com/NREL/floris/pull/1079

PR #954, opened 2024-08-01
Wind direction heterogeneity

misi9170:
==This pull request is a work in progress==

## (re)Introducing wind direction heterogeneity in FLORIS
In FLORIS v2, there was handling for wind direction heterogeneity. Wind direction heterogeneity was not straightforward to continue supporting in the vectorization effort for v3, and was dropped. This pull request reintroduces a new method for handling wind direction heterogeneity by "warping" the turbine layout mimic the effects of a wind direction change.

This is currently a work in progress, and is missing several crucial steps. I will update this description as those steps are addressed.

### Critical pieces still needed:
- [ ] Handling different wind directions (currently, works only for 270 degree winds)
- [ ] Setting main wind speed and wind direction according to specified `u`, `v` values (i.e., reconciling the turbine ordering based on the `wind_direction` with that based on `u` and `v`)
- [ ] Setting wind speed heterogeneity using `u`, `v` values

### Other TODO items:
- [ ] Expand to 3D case (`u`, `v` as a function of `z`; _possibly_ allow `w` specification)
- [ ] Add supporting code to all solvers (currently works with `sequential_solver` and `full_flow_sequential_solver`)
- [ ] Add supporting code to all (necessary) grid types
- [ ] Add supporting code to `HeterogeneousMap`
- [ ] Check that solve speed is not significantly affected

### Housekeeping
- [ ] Existing tests pass
- [ ] New functionality tests
- [ ] Clean up examples
- [ ] Formatting (ruff, isort, etc)
- [ ] Docstrings
- [ ] Documentation

## Methodology
The core idea here for modeling wind direction heterogeneity is that, rather than move the wakes to new locations based on local wind directions, it is easier to move the turbine locations a commensurate amount (warping the turbine layout). After testing several prototypes, the current methodology essentially works by assigning an equivalent wind farm layout _for each turbine_, based on the wind direction changes occurring in the wake of that turbine. That is, when solving for the wakes, the wind farm has a different layout depending on which turbine in the farm is being evaluated.

Users specify these wind direction changes by specifying the "u" (x-directional, i.e. west to east) and "v" (y-directional, i.e. south to north) components of the wind (that is, by specifying a 2D discrete velocity field). Streamlines are propagated from the rotor centers, which define the path of the wake center; these streamlines are then used to determine movement distances for the turbines. That is, the layout is warped according to the streamline to achieve the equivalent effect of a curved wake.

## Examples
Examples, which are a work in progress, can be found in examples/examples_heterogeneity/. The new examples begin with 00x_, for the time being until I have cleaned them up and renamed them.

Running 00x_wind_direction_heterogeneity.py produces
```
Difference (MW): [[0.         0.         0.27390606 0.61041914]]
```
(that is, the downstream turbines' powers are higher in this specific heterogeneous wind condition than in the aligned case)
and
wd_het_1

Running 00x_wind_direction_heterogeneity_visualization.py with `viz_by_sweep = False` produces
wd_het_2
and with `viz_by_sweep = True` produces
wd_het_3

Note that even once (if) this pull request is completed and merged, wind direction heterogeneity capabilities will likely be a beta feature for some time as we iron out any wrinkles that may arise.

## Affected area of the code
The main areas affected by the code are:
- floris/core/grid.py (as layout grids are modified to equivalently reflect wind direction heterogeneity)
- floris/utilities.py (where the main code that computes the streamlines and then warps the layouts appropriately is stored)
There are also other small changes to:
- floris/core/core.py
- floris/floris_model.py


misi9170:
Note that, finally, the streamlines are used to determine a turbine movement, not to actually calculate the wake profile along the streamline. For this reason, I do not think that this fully addresses the suggestion (for wind speed heterogeneity) by @CaiZhiming-HHU in #851. However, it should expand our ability to model complex terrain by proxy.

pjireland:
@misi9170 Very interesting concept! This would be very useful functionality for me, as I've still been using FLORIS version 2 due to its ability to handle wind direction heterogeneity. Do you have an idea when it will make its way into a FLORIS release?

misi9170:
> @misi9170 Very interesting concept! This would be very useful functionality for me, as I've still been using FLORIS version 2 due to its ability to handle wind direction heterogeneity. Do you have an idea when it will make its way into a FLORIS release?

@pjireland , I'm sorry that I didn't reply to this sooner---I've posted a (non)-answer to this over on #1099

PR #956, opened 2024-08-05
Improve examples

aclerc:
# Improve examples

Simple changes to fix ruff issues in the example files and get the plots to display when I run the example scripts.


misi9170:
Hi @aclerc ,

I'm just doing a bit of house-keeping and realized we hadn't ever responded to you on this. Overall, I don't think we will take this into FLORIS---as you point out, many of the changes are due to your use of a stricter set of formatting rules than we require for FLORIS, which, although they don't violate what we officially support, don't seem like a needed change.

Also, the addition of `plt.show()` and `plt.close()` throughout the example scripts changes how they run for me---it means the scripts will stop and wait for the user to close the figure generated by a midway `plt.show()` call before continuing with the script, which in my experience isn't standard and is not how the examples have been set up to run so far.

With that being said, some of the typo improvements etc are welcome. If it works for you, perhaps I could revert most of the changes implemented here and just keep those changes for merger?

aclerc:
Hi @misi9170 yes of course, go for it!

misi9170:
As this is an older PR, merging develop in is bringing in a lot of commits, so just documenting here that the last commit before I merged in develop and started making changes is a76dd11

misi9170:
@aclerc I've now gone through and merged in the develop branch, reverted the files where only formatting changes were happening and removed the `plt.show(); plt.close()` pairs. What's left now are files with functional changes, but there are still some fairly insignificant formatting changes left in those files that I may still revert while leaving the functional changes. 

I now plan to go through, make sure I understand the functional changes (e.g. plotting layouts, axis limits, and random number generation) and possibly revert a few more of the formatting changes, and then this should be ready for merger. I'll try to do that in the next few days.

I've been making the changes directly on your improve-examples branch; hopefully that's ok. If at any point you want to revert to what you had before I started making changes, the commit hash just prior to my changes is a76dd11.

PR #962, opened 2024-08-15
[BUGFIX] Fix bug in check_turbine example (not resetting operation)

misi9170:
@aclerc recently pointed out that [examples/examples_turbine/001_check_turbine.py](https://github.com/NREL/floris/pull/962/files#diff-898f12d839a184736bc46adc128b83f27019cd5b96edd284eed080b49a850f79) was producing unexpected results, apparently increasing the power of the IEA 10MW turbine with decreasing air density. After looking into this, it turned out to be that control setpoints (yaw angles, in particular) were not being reset, leading to erroneous results. 

Prior to the bug fix, 001_check_turbine.py produced the following (note the blue curve to the left of the black curve in the left plot):
iea10MW_prior

Following the bug fix, these change to the expected:
iea10MW_post


paulf81 (approved):
Looks good!  I downloaded, ran and checked the fix

Issue #885, opened 2024-04-20
`_reinitialize` method in `floris_model.py` is slower than it needs to be for minor changes


achenry:
I'm using Floris in a controller in which the model is updated at each time-step with new wind speeds and directions, and it looks like `_reinitialize` is a bottleneck, specifically the `from_dict` call at the end of the method. Would it be possible to set a flag to not update the parameters of floris unless necessary? I only need to update wind speed and direction.

Code profile:
![image](https://github.com/NREL/floris/assets/25332412/d8b1859b-8545-4eca-8404-499d10db1c14)

and the rest of the code profile...
<img width="1082" alt="image" src="https://github.com/NREL/floris/assets/25332412/dd4de0c2-01de-46ff-9262-c5d513ebb1b0">

paulf81:
hi @achenry , let's you and I and @misi9170 find some time to chat about this, maybe Tuesday or Thursday?

misi9170:
I believe this has now been substantially improved by #966, so I'm going to close this issue. Happy to reopen if needed.

Issue #864, opened 2024-04-04
Need to alter "turbine_type" field of turbine definition dictionary for changes to be reflected


misi9170:
Will build out description later. For now, see #856 and #851

PR #966, opened 2024-08-20
Speed up turbine loading operation

misi9170:
@Bartdoekemeijer points out in #965 that instantiating the `FlorisModel` can be slow, especially when there are many turbines. @achenry had also pointed out a similar problem in #885. 

After looking into this some, and using the profiling method suggested by @achenry, I found that a lot of time is being spent in the `construct_turbine_map()` method of `Farm`. Essentially, each `Turbine` is being instantiated separately, even if all turbines are the same type (which is often the case). 

This PR proposes a fix to this that only calls `Turbine.from_dict()` once for each _type_ of turbine, and then uses those (fewer) instantiations to construct the `Farm.turbine_map`. I'm seeing a substantial speed up, and now I think (after the change) that instantiation speed is dominated by `n_findex` rather than `n_turbines`.

There may be other opportunities for speed up during instantiation, but I think this may be the lowest hanging fruit.

### Notes
I took the opportunity to also partly address #864, which raises the issue that the `"turbine_type"` key of the turbine dictionary _must_ be changed in order to implement changes to other parts of the dictionary. While the changes I have made here do not necessarily "fix" the issue, now an informative error is raised when this situation arises.

### Affected areas of the code
- floris/core/farm.py

### Tests
- All existing tests pass
- New test for checking that an error is raised if two different turbine definitions are given with the same `"turbine_type"`
- New test that checks that modified turbine definitions make it into the `turbine_map` correctly

### Examples
The main relevant example is examples/examples_turbine/002_multiple_turbine_types.py. I see no change to the outputs, as expected.

### Performance
Using the example code in #965, I get the following performance on the develop branch:
dis965

while this PR provides:
dis965_fix

### DIFF
--- a/floris/core/farm.py
+++ b/floris/core/farm.py
@@ -120,6 +120,10 @@ class Farm(BaseClass):
 
     internal_turbine_library: Path = field(init=False, default=default_turbine_library_path)
 
+    # Private attributes
+    _turbine_types: List = field(init=False, validator=iter_validator(list, str))
+    _turbine_definition_cache: dict = field(init=False)
+
     def __attrs_post_init__(self) -> None:
         # Turbine definitions can be supplied in three ways:
         # - A string selecting a turbine in the floris turbine library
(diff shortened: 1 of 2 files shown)


paulf81:
Here is also the wrote from running it on python 3.11 on my machine:

image

misi9170:
Looking across the speed results here and on #965, it seems that the "remaining" time between the n_findex=300 and n_findex=3000 cases is approximately 5 seconds across various machines (or more like 3 seconds when using python 3.11) when setting 1000 turbines.

misi9170:
> Reviewing the code and checking the tests this all looks good. the only thought I had, would 1 or 2 additional tests where the number of unique turbine types is different than the number of turbines (and some assertions about what should be observed) be good to add while we're at it? Or is that all sufficiently covered by existing tests?

Yeah, I could add a quick test for that. The test I've added currently have 5 turbines, one of which is different, but I should check that now if `"turbine_type"` _is_ renamed, the setup goes correctly.

misi9170:
Slight update on the timings where we aren't rounding the time to the nearest integer:
dis965_fix_float

misi9170:
> Reviewing the code and checking the tests this all looks good. the only thought I had, would 1 or 2 additional tests where the number of unique turbine types is different than the number of turbines (and some assertions about what should be observed) be good to add while we're at it? Or is that all sufficiently covered by existing tests?

I have now added this test

paulf81:
Good to go from me then!

Bartdoekemeijer:
Awesome!

Here's my performance with 3.11.8 on WSL with Ubuntu 22 LTS:
image

Thank you so much for fixing this, and especially on such short notice!!

paulf81:
> Awesome!
> 
> Here's my performance with 3.11.8 on WSL with Ubuntu 22 LTS: image
> 
> Thank you so much for fixing this, and especially on such short notice!!

A nice improvement!

floris/core/farm.py
```diff
@@ -134,21 +138,29 @@ def __attrs_post_init__(self) -> None:
         # This allows to read the yaml input files once rather than every time they're given.
         # In other words, if the turbine type is already in the cache, skip that iteration of
         # the for-loop.
+
         turbine_definition_cache = {}
```
bayc, line 142: Could this just be modified to be `self._turbine_definition_cache`? Then you could remove line 219, I believe. Also, update `turbine_definition_cache` to `self._turbine_definition_cache` throughout.

bayc (approved):

bayc:
@misi9170 Yea, this looks nice! Just one suggestion in my comment above.

misi9170:
> @misi9170 Yea, this looks nice! Just one suggestion in my comment above.

This is now done

PR #968, opened 2024-08-21
Improved wind rose plot

cfrontin:
This is an improvement to the standard wind rose plot that replaces a legend with squares at each discrete wind speed with a colorbar that is matched to the discrete wind speeds that were previously there.


paulf81:
Thank you @cfrontin, I opened #969 to make the same changes on develop branch and debug above issues

paulf81:
Closing in favor of #969

PR #969, opened 2024-08-22
Improved Wind Rose Plot (v2)

paulf81:
# Improved Wind Rose Plot v2

This is a reimplementation of the code changes proposed by @cfrontin in #968 but built on the develop branch instead of main (something about running that way is triggering some errors, and anyway breaks our convention).  Running for example the wind_data_user notebook changes the look of wind rose plots from:

image

To:

image

Which is an improvement to my eye.  Thanks @cfrontin for the submission!

## Related issue
#968 

wind_data.py


paulf81:
If we make this change here should also consider WindTIRose

floris/wind_data.py
```diff
@@ -776,7 +779,10 @@ def plot(
                 )
 
         # Configure the plot
-        ax.legend(reversed(rects), ws_bins, **legend_kwargs)
+        # ax.legend(reversed(rects), ws_bins, **legend_kwargs)
+        cb_ws = ax.figure.colorbar(sm_ws, ax=ax, **legend_kwargs)
+        ax.figure.tight_layout()
+        cb_ws.outline.set_visible(False)
```
cfrontin, line 785: I actually didn't mean to leave the `cb_ws.outline.set_visible(False)`, which I added to conform to my personal style preferences for a paper. if you want to match standard numpy viz, you might want to remove that line.

cfrontin (approved):

misi9170:
Note that merging this PR could potentially introduce a breaking change for a very small number of users. The `ax` returned by the `plot()` method will now no longer contain a legend (and instead a colorbar). Any code that subsequently obtains and modifies the properties of the legend will no longer work.

While this is indeed a breaking change, and therefore against the [rules of semantic versioning](https://semver.org/) that we try to adhere to with FLORIS, we feel that this is a small enough change that it will affect very few (and quite possibly no) users, and are going to proceed to merge under the same major version of FLORIS (v4).

PR #976, opened 2024-09-05
Add gridded layout optimizer

misi9170:
This pull requests adds a new "optimization" method that simply searches for the gridded layout (either a square grid or a hexagonal grid) that fits the most wind turbines into a specified boundary area. The spacing in the grid is fixed by the user, but the grid's rotation and translation is optimized (via a simple parameter sweep) to fit the most turbines into the boundary.

`FlorisModel.run()` is not called, so there is no wake or AEP calculation---the optimizer simply chooses the grid that fits the most turbines. Where multiple rotations/translations result in the same number of turbines, the result returned is (arbitrarily) the first of those. @ejsimley has suggested we could consider a "tie-breaker" of running an AEP calculation for all layouts that have the maximum number of turbines and selecting that with the highest AEP; for now, this has been left as future work.

There is no requirement that the boundary be a single area---users may specify a boundary with separate areas without issue.

The core code is complete. However, it currently runs somewhat slower than I'd hoped, and I intend to profile the code to see if there are any easy ways to speed it up.

### To do:
- [x] Core code implemented
- [X] Existing tests pass
- [X] New tests added
- [x] Speed profiled
- [x] ~Low-hanging fruit for speed-ups exploited~
- [x] Example demonstrating capability
- [x] Documentation

### DIFF
--- a/floris/optimization/layout_optimization/layout_optimization_gridded.py
+++ b/floris/optimization/layout_optimization/layout_optimization_gridded.py
@@ -10,7 +10,7 @@ from shapely.geometry import (
 )
 
 from .layout_optimization_base import LayoutOptimization
-
+from .layout_optimization_random_search import test_point_in_bounds
 
 class LayoutOptimizationGridded(LayoutOptimization):
     def __init__(
@@ -18,9 +18,11 @@ class LayoutOptimizationGridded(LayoutOptimization):
         fmodel,
         boundaries,
         spacing: float | None=None,
-        spacing_D: float | None=None,
+        spacing_D: float | None=-1,
         rotation_step: float=5.0,
         rotation_range: tuple[float, float]=(0.0, 360.0),
+        translation_step: float | None=None,
+        translation_step_D: float | None=-1,
         hexagonal_packing: bool=False,
         enable_geometric_yaw: bool=False,
         use_value: bool=False,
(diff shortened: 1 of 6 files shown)


misi9170:
After profiling, not surprisingly it appears that the slowest part of the code is that which checks, point by point, whether candidate locations are inside the boundary. This code relies on [shapely's `contains` method](https://shapely.readthedocs.io/en/stable/reference/shapely.contains.html), which is called inside a nested loop: the outer loop is over the candidate rotations/translations, and the inner loop (represented as a list comprehension) is over the points within the candidate layout.

I tried a couple of quick things to speed this up, both of which failed:
1. Using numba's `@jit` decorator on the `test_point_in_bounds` method, which is simply a wrapper for `shapely.contains()` (numba could not recognize the shapely objects
2. Running `shapely.contains()` on a `MultiPoint` object, rather than looping over each `Point`, to try to remove one level of the nested loop (`shapely.contains()` appears to check whether _all_ of the points are contained in the shape and returns a scalar boolean value, rather than checking whether _each_ point is contained as we need.

Overall, I think we can treat the code as fast enough for now. When I run the new example 004_generate_gridded_layout.py with `min_dist_D=0.5`, the 1000m x 1000m square boundary can fit 256 turbines, and the code spends about 1s in the `test_point_in_bounds` function running on my laptop.

misi9170:
@paulf81 this is now ready for review

floris/optimization/layout_optimization/layout_optimization_gridded.py
```diff
@@ -0,0 +1,179 @@
...
+
+from .layout_optimization_base import LayoutOptimization
+from .layout_optimization_random_search import test_point_in_bounds
+
+
+class LayoutOptimizationGridded(LayoutOptimization):
+    def __init__(
+        self,
+        fmodel,
+        boundaries,
```
paulf81, line 13: Should we add type hints to complete the pattern?

floris/optimization/layout_optimization/layout_optimization_gridded.py
```diff
@@ -0,0 +1,179 @@
...
+        rotation_step: float = 5.0,
+        rotation_range: tuple[float, float] = (0.0, 360.0),
+        translation_step: float | None = None,
+        translation_step_D: float | None = -1,
+        translation_range: tuple[float, float] | None = None,
+        hexagonal_packing: bool = False,
+        enable_geometric_yaw: bool = False,
+        use_value: bool=False,
+
+    ):
```
paulf81, line 25: Need to add docstring

floris/optimization/layout_optimization/layout_optimization_gridded.py
```diff
@@ -0,0 +1,179 @@
...
+        # Save boundaries
```
paulf81, line 26: This is maybe in the wrong place

floris/optimization/layout_optimization/layout_optimization_gridded.py
```diff
@@ -0,0 +1,179 @@
...
+        # Handle spacing information
+        if min_dist is not None and min_dist_D is not None and min_dist_D >= 0:
+            raise ValueError("Only one of min_dist and min_dist_D can be defined.")
+        if min_dist is None and min_dist_D is None:
+            raise ValueError("Either min_dist or min_dist_D must be defined.")
+        if min_dist_D is not None and min_dist is None:
+            if min_dist_D < 0: # Default to 5D
+                min_dist_D = 5.0
```
paulf81, line 34: Would be good to explain this behavior in docstring

floris/optimization/layout_optimization/layout_optimization_gridded.py
paulf81: This could also underline this optimization differs in objective from some of the others

paulf81 (approved):
Overall this looks really great!  All tests pass and examples run and produce clear results.  Doc is also very clear.  Approving and leave remaining minor comments up to you @misi9170

floris/optimization/layout_optimization/layout_optimization_gridded.py
```diff
@@ -0,0 +1,179 @@
...
+
+        # Initial locations not used for optimization, but may be useful
+        # for comparison
+        self.x0 = fmodel.layout_x
+        self.y0 = fmodel.layout_y
+
+        # Create the default grid
+
+        # use min_dist, hexagonal packing, and boundaries to create a grid.
+        d = 1.1 * np.sqrt((self.xmax**2 - self.xmin**2) + (self.ymax**2 - self.ymin**2))
```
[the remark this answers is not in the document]
misi9170: Just as a little bit of over engineering to ensure that 'd' is  sufficient to cover the boundary area, even after translations/rotations. May not be necessary, but I thought just easy to add a safety factor.

PR #977, opened 2024-09-05
Initial work on a consistent parallel FLORIS model

misi9170:
Attempt to make a `ParallelFlorisModel` that is completely consistent with the `FlorisModel` introduced in v4. 

Work in progress. Tests currently passing, however, more tests are needed.


floris/parallel_floris_model_2.py
```diff
@@ -0,0 +1,309 @@
...
+                - **farm**: See `floris.simulation.farm.Farm` for more details.
+                - **turbine**: See `floris.simulation.turbine.Turbine` for more details.
+                - **wake**: See `floris.simulation.wake.WakeManager` for more details.
+                - **logging**: See `floris.simulation.core.Core` for more details.
+            interface: The parallelization interface to use. Options are "multiprocessing",
+               with possible future support for "mpi4py" and "concurrent"
+            max_workers: The maximum number of workers to use. Defaults to -1, which then
+               takes the number of CPUs available.
+            n_wind_condition_splits: The number of wind conditions to split the simulation over.
+               Defaults to the same as max_workers.
```
paulf81, line 47: The default is 1 above, is this a sentinal for same as max_workers?

floris/parallel_floris_model_2.py
```diff
@@ -0,0 +1,309 @@
...
+
+class ParallelFlorisModel(FlorisModel):
+    """
+    This class mimics the FlorisModel, but enables parallelization of the main
+    computational effort.
+    """
+
+    def __init__(
+        self,
+        configuration: dict | str | Path | FlorisModel,
```
paulf81, line 24: Is the FlorisModel type hint consistent with the fact that it will give a warning if passed in?

floris/parallel_floris_model_2.py
```diff
@@ -0,0 +1,309 @@
...
+            configuration = configuration.core.as_dict()
+        super().__init__(configuration)
+
+        # Save parallelization parameters
+        if interface == "multiprocessing":
+            import multiprocessing as mp
+            self._PoolExecutor = mp.Pool
+            if max_workers == -1:
+                max_workers = mp.cpu_count()
+            # TODO: test spinning up the worker pool at this point
```
paulf81, line 67: Still open?

floris/parallel_floris_model_2.py
```diff
@@ -0,0 +1,309 @@
...
+    This class mimics the FlorisModel, but enables parallelization of the main
+    computational effort.
+    """
+
+    def __init__(
+        self,
+        configuration: dict | str | Path | FlorisModel,
+        interface: str | None = "multiprocessing",
+        max_workers: int = -1,
+        n_wind_condition_splits: int = 1,
```
paulf81, line 27: Should this be -1?

floris/parallel_floris_model_2.py
```diff
@@ -0,0 +1,309 @@
...
+            # Extract and format all control setpoints as a dict that can be unpacked later
+            control_setpoints_subset = {
+                "yaw_angles": self.core.farm.yaw_angles[wc_id_split, :],
+                "power_setpoints": self.core.farm.power_setpoints[wc_id_split, :],
+                "awc_modes": self.core.farm.awc_modes[wc_id_split, :],
+                "awc_amplitudes": self.core.farm.awc_amplitudes[wc_id_split, :],
+                "awc_frequencies": self.core.farm.awc_frequencies[wc_id_split, :],
+            }
+
+            # Do I also need to hangle splitting the wind data object?
```
paulf81, line 173: typo

paulf81: Also, we can make some tests to ensure works with WindRose etc.,

floris/parallel_floris_model_2.py
```diff
@@ -0,0 +1,309 @@
...
+    Args:
+        fmodel: The FLORIS model to run.
+        set_kwargs: Additional keyword arguments to pass to fmodel.set().
+    """
+    fmodel = FlorisModel(fmodel_dict)
+    fmodel.set(**set_kwargs)
+    fmodel.run()
+    return fmodel.get_turbine_powers()
+
+def _optimize_yaw_angles_serial(
```
paulf81, line 282: Leaning toward not including this function

examples/0xx_parallel_computation.py
```diff
@@ -0,0 +1,117 @@
...
+from floris import (
+    FlorisModel,
+    TimeSeries,
+    WindRose,
+)
+from floris.parallel_floris_model import ParallelFlorisModel as ParallelFlorisModel_orig
+from floris.parallel_floris_model_2 import ParallelFlorisModel as ParallelFlorisModel_new
+
+
+if __name__ == "__main__":
```
paulf81, line 17: Something I've been wondering is there a line we could include within ParallelFlorisModel_new to detect whether or net it has been initialized following a if __name__ == "__main__":, and if not giving a helpful warning

floris/parallel_floris_model_2.py
misi9170: Agreed. I'm hopeful that it will, since the wind data objects are only actually used in preprocssing (in `set()`) and postprocessing (in `get_...`), but I'll add some tests to check.

paulf81:
Looks really nice @misi9170 , went through and left a few comments.  Also I tested tests and examples and ran well

tests/parallel_floris_model_2_unit_test.py
```diff
@@ -0,0 +1,140 @@
+
+import copy
+import logging
+
+import numpy as np
+import pytest
+
+from floris import (
+    FlorisModel,
+    UncertainFlorisModel,
```
paulf81, line 10: I think would be good to add tests with WindRose and TimeSeries just to be sure

floris/parallel_floris_model_2.py
```diff
@@ -0,0 +1,309 @@
...
+        for wc_id_split in wind_condition_id_splits:
+            # for ws_id_split in wind_speed_id_splits:
+            fmodel_dict_split = copy.deepcopy(fmodel_dict)
+            wind_directions = self.core.flow_field.wind_directions[wc_id_split]
+            wind_speeds = self.core.flow_field.wind_speeds[wc_id_split]
+            turbulence_intensities = self.core.flow_field.turbulence_intensities[wc_id_split]
+
+            # Extract and format all control setpoints as a dict that can be unpacked later
+            control_setpoints_subset = {
+                "yaw_angles": self.core.farm.yaw_angles[wc_id_split, :],
```
paulf81, line 166: This is bigger than this PR, but something I wonder, will we need a process to go throughout the code should we ever add a new control setpoint?

floris/parallel_floris_model_2.py
misi9170: Still something I'd like to look into a bit further

floris/parallel_floris_model_2.py
misi9170: Good catch! That was a typo and a bug---this default should be -1. I'm just rerunning tests etc with that change.

floris/parallel_floris_model_2.py
misi9170: Ah, sorry, my mistake---I didn't mean to commit that. Will remove.

floris/parallel_floris_model_2.py
misi9170: Good point---I've removed `FlorisModel` from the list, but checked that it does still run if an instantiated `FlorisModel` is passed in.

floris/parallel_floris_model_2.py
misi9170: I've been thinking a bit about this too. It would be nice if we could generalize all the control setpoints in some way, so that we don't have to add infrastructure each time we add a new one. They'll likely all be shape n_findex * n_turbines, and we could consider moving to a `control_setpoints` dictionary, which can be unpacked and checked programmatically some way.

tests/parallel_floris_model_2_unit_test.py
misi9170: Done

floris/parallel_floris_model_2.py
misi9170: I do need to add tests for at least some of the control setpoints though, to make sure those are currently handled correctly. Probably just testing with `yaw_angles` is enough for now.

floris/parallel_floris_model_2.py
misi9170: I've now added those tests and removed this comment.

floris/parallel_floris_model_2.py
misi9170: I've now added testing for `yaw_angles` (and did need to address a bug!).

misi9170:
Since there are a few separate pieces that will need to go into a new class to (eventually) replace the `ParallelFlorisModel`, I've created a [new branch](https://github.com/NREL/floris/tree/parallel-model-rework) on NREL/floris and have changed the base to point to that branch. There may be a handful of PRs to that branch as we develop it (which will be merged with "squash and merge" as they go in), and finally a merge commit to merge parallel-model-rework into develop when we are happy with the replacement class.

Issue #971, opened 2024-08-22
ParallelFlorisModel updated to match FlorisModel API


paulf81:
# ParallelFlorisModel updated to match FlorisModel API

ParallelFlorisModel currently has an API not exactly 1-1 with FlorisModel.  This may be necessary, for example in order to defer certain calculations until splitting onto cores, however ideally the variance will be minimized as much as possible.  This issue is a placeholder to hold the intention to do some improvements in this vane and more generally.

## Proposed solution

- [x] Provide default values to max_workers and n_wind_condition_splits based on detection of available cores
- [x] Remove unsupported parallel computing interfaces, perhaps standardize on one?
- [x] Make sure all control vectors are supported, not just yaw
- [x] Remove no_wake flags in favor of run, run_no_wake style of v4
- [x] Match as close as possible FlorisModel interface

## Additional context
 A small subset of improvements was added in #967

misi9170:
Closed by #982

Issue #944, opened 2024-07-12
Parallelize `YawOptimizationScipy` and `YawOptimizationSR` `optimize` methods across wind speeds/directions


achenry:
The optimization procedure, in particular the `YawOptimizationScipy`, would be much faster if parallelized across all available cores, given passed argument `parallel` to `optimize method. Would require instantiating copies of `fmodel' for each run.

Bartdoekemeijer:
Hi @achenry, take a look at the `parallel_floris_model` interface, https://github.com/NREL/floris/blob/01115bf89645aa0aefc1a2f5db35078bfe17896e/floris/parallel_floris_model.py#L456

That does serve your needs?

misi9170:
@achenry we could look into adding a method for scipy optimization to the `ParallelFlorisModel` too---either by creating a new function similar to [`_optimize_yaw_angles_serial()`](https://github.com/NREL/floris/blob/01115bf89645aa0aefc1a2f5db35078bfe17896e/floris/parallel_floris_model.py#L22C1-L22C32) for scipy optimization or by allowing the user to select between optimization routines in `_optimize_yaw_angles_serial()`.

achenry:
A few thoughts...
1. It might be more helpful to use the same FlorisModel, but pass a parallel flag to the optimize methods, to facilitate switching easily between sequential and parallel operation for debugging
2. For use with kestrel/HPC, need a working example including bash_script (I have these).

misi9170:
For Floris v4, the hope was to have both the `UncertainFlorisModel` and the `ParallelFlorisModel` be drop-in replacements for the `FlorisModel`; that is, the `FlorisModel` would not take optional `parallel` or `uncertain` arguments, but rather the line
`fmodel = FlorisModel(...)` 
could be simply replaced with 
`fmodel = ParallelFlorisModel(...)` / `fmodel = UncertainFlorisModel(...)`
to run in parallel or uncertainty, respectively.

Although we achieved this for the `UncertainFlorisModel`, we didn't end up getting to this for the `ParallelFlorisModel`---the user interface is still different for the `ParallelFlorisModel` than the `FlorisModel`. We do hope to address this (make a parallelized drop-in for the `FlorisModel`, which hopefully parallelizes the `run()` method) in a future release. I think this would be similar to what you're after?

[...]

PR #982, opened 2024-09-16
New parallel option that is fully consistent with FlorisModel

misi9170:
This pull requests provides a new interface for parallel computing with FLORIS. The existing `ParallelFlorisModel` was not fully consistent with the new `FlorisModel` and `UncertainFlorisModel` developed for FLORIS v4. This addresses the inconsistency in a new class, `ParFlorisModel`. `ParallelFlorisModel` will remain in FLORIS v4 to avoid making breaking changes, but now has a deprecation warning and will be removed in the next major version release of FLORIS. 

This addresses #971 and possibly partially address #944.

Development still to take place:
- [x] Create script for timing both the existing `ParallelFlorisModel` and the new class with increasing `n_findex` and increasing `n_turbines`, and can be run on both HPC and locally. Report results.
- [x] Test whether there is a significant speed increase in returning powers only, rather than full velocity fields. Even if not, we may leave that option. [Minimal effect when using `multiprocessing`]
- [x] Test other parallel computing paradigms and see whether they'd be easy to include (concurrent, mpi4py)
- [x] See whether it is possible to spin up the worker pool on instantiation of the class and then send it jobs, to reduce overhead if multiple `run()` calls are needed [Yes, with `pathos` package]

Checklist before converting this draft to full PR status and preparing for merger:
- [x] Remove temporary examples designed to be run on HPC
- [x] Add documentation
- [x] Select name for new class and rename
- [x] Add deprecation warning to existing ParallelFlorisModel
- [x] Add any new examples
- [x] Replace usage of existing ParallelFlorisModel in examples (if it’s there)


misi9170:
@paulf81 Something to decide on is the default parallel interface. My thinking is that the default should be `"multiprocessing"`, even though `"pathos"` is faster, because the `multiprocessing` package is pretty ubiquitous and `pathos` is still on version 0.x

paulf81:
> Something to decide on is the default parallel interface. My thinking is that the default should be `"multiprocessing"`, even though `"pathos"` is faster, because the `multiprocessing` package is pretty ubiquitous and `pathos` is still on version 0.x

I agree, using multiprocessing seems the better choice for default

misi9170:
We have decided to leave out mpi4py support for now. If requested, we may consider including this in future.

examples/examples_uncertain/003_uncertain_model_with_parallelization.py
misi9170, line 1: This example is perhaps a bit redundant now given examples_uncertain/001.. and 009_parallel_models.py? However, it is nice to include it here in case users are more focused on uncertainty than parallelization, but may still benefit from parallelization

docs/floris_models.ipynb
misi9170, line 1: Do we want to add a description of the `ApproxFlorisModel` here? Could also create an issue and leave this for a later development.

floris/parallel_floris_model.py
```diff
@@ -89,6 +88,11 @@ def __init__(
         print_timings (bool): Print the computation time to the console. Defaults to False.
         """
 
+        self.logger.warning((
```
paulf81, line 91: We could add a link here to docs, but downside is the link won't work until we merge to develop

floris/parallel_floris_model.py
```diff
@@ -112,11 +116,19 @@ def __init__(
...
                 f"Interface '{interface}' not recognized. "
                 "Please use 'concurrent', 'multiprocessing' or 'mpi4py'."
             )
 
+        # Raise error if uncertain model is passed in and refer to new parallel_floris_model_2
+        if isinstance(fmodel, UncertainFlorisModel):
+            raise ValueError(
+                "UncertainFlorisModel is not supported in this version of ParallelFlorisModel. "
+                "Please use the new version ParFlorisModel (par_floris_model) "
+                "for UncertainFlorisModel compatibility."
```
paulf81, line 129: Another spot we could theoretically link to docs, have an example of this usage there now

paulf81 (approved):
Confirmed all examples run, all docs built correctly and all tests pass.  Only a minor comment in one file we could hypothetically link to docs, but this seems a little brittle and so maybe shouldn't be done

floris/parallel_floris_model.py
misi9170: Yeah, as you point out in your comment further down, I think that will be too brittle---hopefully the suggestion will be enough and users will look at the examples/docs to find the `ParFlorisModel`

examples/examples_uncertain/003_uncertain_model_with_parallelization.py
paulf81: Yeah, but I thought it a helpful thing to repeat in that context

docs/floris_models.ipynb
paulf81: It's a good idea, I'll add it now

PR #984, opened 2024-09-19
[BUGFIX] Gridded layout opt incorrectly generating spanning grid

misi9170:
This small pull request fixes two bugs and adds one enhancement to the recently developed `LayoutOptimizationGridded` class.
- Fixes an incorrect line that could result in a "spanning" layout that did not actually span the boundary area
- Ensures that only a single `rotor_diameter` is returned, which was causing problem as the dimension of `fmodel.core.farm.rotor_diameters` is not always 1. 
- Makes minor enhancement to ensure that `n_turbines_max` is of type `int`, rather than a float that will always be a whole number, which is then useful for indexing arrays/lists.


paulf81 (approved):
These changes make sense to me

PR #985, opened 2024-09-26
Expand parallel processing interface options

misi9170:
Extends from existing `interface` options, currently just `"multiprocessing"`, to include:
- [x] `"pathos"`
- [x] `"concurrent"`
- [ ] ~`"mpi4py"`~

I've left mpi4py out for now, as this is more complex and quite possibly unnecessary at this level of the code.


misi9170:
To keep the ball rolling here, I'm going to merge this PR as is and we'll have a proper review process when we merge parallel-model-rework into develop.

PR #992, opened 2024-10-04
Add automatic benchmarking to FLORIS

paulf81:
# Add automatic benchmarking to FLORIS

This draft PR is meant to add automatic code benchmarking to FLORIS.  Proposed solution is to use pytest-benchmark to implement set timing tests:

https://pytest-benchmark.readthedocs.io/en/latest/

https://github.com/ionelmc/pytest-benchmark

And then try to schedule some semi-daily execution of these tests with logged performance checks so we can track changes over time.  Here focused on:

https://github.com/benchmark-action/github-action-benchmark

To this end I added a first test to the tests folder including benchmarking to the `tests/` folder and confirm the command line:

`pytest floris_benchmark_test.py `

Produces a benchmark result.  At this point might open up for discussion, or others research:

1. Do these benchmarks belong in tests?  Won't that cause them to be run with normal CI process?
2. What should we test?
3. We should set a test results page
4. Test parallel?
5. Want to track changes coming from our own work, and also those coming from improvement to python, should make sure CI checks most recent version of python?

To include:

- [ ] AEP for 100-turbine farm
- [ ] Analysis for different wake models
- [ ] Parallel FLORIS


Bartdoekemeijer:
This is great!! On my end, I would love to see performance on, let's say, AEP calculation for a 100-turbine farm. Would also be nice to see this for different wake model set-ups, in case modifications are made to specific submodels. Benchmarking parallel floris would be a nice plus.

.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,61 @@
+name: Floris Benchmark
+on:
+  schedule:
+    - cron: '0 3 * * *'  # Runs daily at 3am UTC
```
rafmudaf, line 4: I suggest tying this to a commit rather than a date since any given day might contain multiple commits and many will contain no commits.

.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,61 @@
...
+  workflow_dispatch:  # Allows manual triggering of the workflow
```
rafmudaf, line 5: That's neat

.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+permissions:
+  contents: write
+  deployments: write
```
rafmudaf, line 9: What's the need for access to deployment rights?

benchmarks/data/input_full.yaml
rafmudaf, line 1: I would avoid adding yet another input file (that will need to be maintained) just for this

setup.py
```diff
@@ -45,6 +45,7 @@
     },
     "develop": {
         "pytest",
+        "pytest-benchmark"
```
rafmudaf, line 48: Probably not needed yet, but if the list of dependencies only used in automated systems grows, you could add a new target to capture those and separate them from the ones we'll all install for development

rafmudaf (commented):
I know this pull request is still in progress, but just wanted to leave some comments for future reference. Good stuff @paulf81

rafmudaf:
It would be good to add a scaling test, as well. This should be a test of something like 10, 100, 1000 turbines across a meaningful number of conditions (~1000).

.github/workflows/benchmark.yaml
paulf81: Yeah, that makese sense, maybe commits to develop?

benchmarks/data/input_full.yaml
paulf81: that's a good point, this can be switched to one of the existing inputs

setup.py
paulf81: that makes sense

paulf81:
Thank you for your comments @rafmudaf !  I'll hope to take another pass at this soon and incorporate your suggestions

.github/workflows/benchmark.yaml
paulf81: Ok switched to push to develop

.github/workflows/benchmark.yaml
paulf81: I think we need it to display a live chart: https://github.com/benchmark-action/github-action-benchmark

benchmarks/data/input_full.yaml
paulf81: I think that should work, we should prioritize #1040 then

paulf81:
Closing this pull request in favor of #1060

Issue #989, opened 2024-10-01
Complete multi-dim code


paulf81:
# Complete multidim code

This issue picks up from issue #785 and proposes a set of steps to finish this work.

## Proposed solution

Recommend to take the following steps (can iterate as we go):

- [ ] Add example of applying multi-dim code to vary power by TI
- [ ] Add tests of this same functionality
- [ ] Include example of going to multidim via code versus from specially setup yaml
- [ ] Add tests of same functionality?
- [ ] Beef up a little bit multidim documentation

misi9170:
I'm working on this over on #996

paulf81:
This solution makes sense!

misi9170:
Addressed in #996

Issue #785, opened 2024-01-25
Allow TI to impact turbine power production


paulf81:
# Allow TI to impact turbine power production

In V4, TI can be set per time step in TimeSeries or wd/ws bin in WindRose.  It would be useful to reflect TI impact on power production of a turbine, separate of wake model impacts.  Talking this over with @ejsimley we think we have a way forward with 90% of the details in mind.  Here is the proposed solution to implement:

- As with tilt, assume a given power curve is defined at a certain reference turbulence intensity.  The power curves included with FLORIS are rather sharp and I believe our 0% TI, but it is on common to have a power curve defined for a certain TI.  So, include into turbine yaml the ref_ti value
- Therefore the amount to correct a given power should be:
  - 0 if TI <= ref_ti
  - delta_ti = ti - ref_ti if ti > ref_ti
- Correct the power curve (and ct) using a version of the method previously implemented in v2:

```
  # define wind speed, ti, and power curve components
  ws = np.array(self.power_thrust_table["wind_speed"])
  cp = np.array(self.power_thrust_table["power"])
  ws = ws[np.where(cp != 0)]
  ciws = ws[0]  # cut in wind speed
  cows = ws[len(ws) - 1]  # cut out wind speed
  speed = self.average_velocity
  ti = self.current_turbulence_intensity

  if ciws >= speed or cows <= speed or ti == 0.0 or math.isnan(speed):
      return 1.0
  else:
      # define mean and standard deviation to create normalized pdf with sum = 1
      mu = speed
      sigma = ti * mu
      if mu + sigma >= cows:
          xp = np.linspace((mu - sigma), cows, 100)
      else:
          xp = np.linspace((mu - sigma), (mu + sigma), 100)
      pdf = norm.pdf(xp, mu, sigma)
      npdf = np.array(pdf) * (1 / np.sum(pdf))

      # calculate turbulence parameter (ratio of corrected power to original power)
      return np.sum([npdf[k] * self.powInterp(xp[k]) for k in range(100)]) / (
          self.powInterp(mu)
      )
```
- Note in essence the code returns a gaussian blur of points on the power (or thrust) curve.  I think in v4 we should implement the logic of this, but find a way to precompute this effect
- Then just need to route this logic around
- Finally, in v2 there was a boolean value whether the user wanted this turbulence correction applied (default to False) do we want the same?

[...]

Issue #901, opened 2024-05-06
Bug report: TurbineInterface not covered in tests and examples


misi9170:
In merging #888, the deploy-book action failed because I had forgotten to update some function calls in the `TurbineInterface` object on on floris/turbine_library/turbine_previewer.py

Looking into this more, the reason that I missed these is that `TurbineInterface` is not tested, and also not used in any of the examples. 

I would suggest that we consider either:
1. Adding a test for the `TurbineInterface` or
2. Deprecating `TurbineInterface` and just going through the `FlorisModel` in the documentation (full deprecation would need to wait until a later major release)

paulf81:
Based on this issue patched by #900 and the fact that it is only in documentation, it does seem reasonable to deprecate the functionality into `FlorisModel`, which will be more easily tracked

misi9170:
An option is to have a convenience class that is a subclass of `FlorisModel` for the specific purpose of turbine visualization. I'll need to think a bit more about whether that's worthwhile.

misi9170:
After some discussion, I have added deprecation warnings to the `TurbineInterface` and `TurbineLibrary` classes and have removed mention of them from the documentation in #996. I'm closing this as complete, and these classes will be removed or revamped in a future major version release.

PR #996, opened 2024-10-09
[BUGFIX] Improve handling of multidimensional turbine conditions

misi9170:
#989, #785


misi9170:
I'm now done with code changes (I think), but starting to work on the docs I think this is a good time to go ahead and improve turbine documentation across the board. I plan to add a Turbine "subfolder" to the documentation, and in it place (while reworking):
- [x] Turbine Operation Models
- [x] Floating turbine modeling #1125
- [x] Turbine library interface (add deprecation note, possibly warning in code) #901 
- [x] (new) multidimensional Cp/Ct definition information #1118 
- Turbine input file reference #1053, #1125

misi9170:
After discussing with @dzalkind , I've decided that in this PR, I will include the bug fix for specifying multidimensional input conditions but not yet update reference values for the floating IEA 15MW turbine. In the mean time, I have updated various pieces of the documentation for floating turbines and indicated that the floating tilt table, along with the dependency of the thrust coefficient and power curves on wave height and wave period, should be treated as fictional. I will indicate this decision in related issues and discussions.

examples/examples_floating/001_floating_turbine_models.py
```diff
@@ -89,7 +89,7 @@
     tilt_angles_floating_defined_floating,
     color="m",
     ls="--",
-    label="Floating (cp/ct not scaled by tilt)",
+    label="Floating (power/thrust coefficient not scaled by tilt)",
```
paulf81, line 92: This could be a prior issue, but shouldn't the magenta dashed line be on top of the blue line in the 0th axis?

examples/examples_multidim/003_multi_dimensional_cp_ct_TI.py
```diff
@@ -0,0 +1,45 @@
...
+# Use a sweep of wind speeds
+wind_speeds = np.arange(5, 20, 1.0)
+time_series = TimeSeries(
+    wind_directions=270.0, wind_speeds=wind_speeds, turbulence_intensities=0.06
+)
+fmodel.set(wind_data=time_series)
+
+# Loop over different turbulence intensities using set()
+fig, axarr = plt.subplots(1, 3, sharex=True, figsize=(12, 4))
+for ti, col in zip([0.06, 0.10], ["k", "r"]):
```
paulf81, line 34: The example works so I think is fine, but noting that TI is outside of the define bounds of the table, maybe comment if this is then fixed at the nearest or extrapolated?  Or this is covered in docs, just flagging

examples/examples_multidim/003_multi_dimensional_cp_ct_TI.py
```diff
@@ -0,0 +1,45 @@
+"""Example: Multi-dimensional power/thrust coefficients with turbulence intensity
+This example follows the previous example, but demonstrating how a multidimensional turbine can be
+used to model the effect of turbulence intensity on power and thrust coefficient.
+
+NOTE: The multi-dimensional power/thrust coefficient data used in this example is fictional for the
+purposes of facilitating this example.
```
paulf81: Turbine 0's power is increased by turbulence, even near the knee of the power curve.  You point out a few times this is fake data but could also say "... and that is why the changes in power curve due to TI don't follow usual behaviors..."

docs/input_reference_main.md
```diff
@@ -1,6 +1,6 @@
 # Main Input File Reference
 
-In additional to reinitializing {py:class}`FlorisInterface`, users can configure FLORIS
+In additional to calling the `set()` method on {py:class}`FlorisModel`, users can configure FLORIS
```
paulf81: addition

paulf81 (approved):
I reviewed and ran the tests and went through and ran all examples.  I reviewed all the code and docs and all looks good to me.  Left a few minor comments

examples/examples_multidim/003_multi_dimensional_cp_ct_TI.py
misi9170: Good point---the multidimensional turbine code always uses a nearest-neighbor approach. I've put a comment into the examples to this effect, and also added this information to the multidimensional turbine documentation.

examples/examples_floating/001_floating_turbine_models.py
misi9170: I think this is currently correct. The idea here is that the black line represents the fixed-bottom NREL 5MW; the blue line represents what happens if you took the fixed NREL 5MW and put it on a floater (with a fictional level of added tilt for different wind speeds); and the magenta dashed line is the (fictional) case where the NREL 5MW power and thrust coefficient is defined as being on a floating platform, so no tilt correction to the power and thrust curve should be applied (as the effect of tilt is already inherently included).

misi9170:
@paulf81 Thanks for the review! I've now addressed your comments.

I also realized that we didn't have a way to store the multidimensional csv file in any user location---only in floris/turbine_library/, which was a bit unsatisfying. I've now made this possible by passing the `turbine_library_path` down to the `Turbine()` class, which instantiates the turbines and loads the multidimensional csv file, if it exists. This new approach is now used in examples/examples_multidim/003_multidimensional_cp_ct_TI.csv and also in the new test `test_set_multidim()` on floris_model_integration_test.py.

Assuming all examples run to completion, this PR will be ready for merge into the develop branch.

paulf81:
agreed

PR #1006, opened 2024-10-21
Add flow visualization to wake steering example

misi9170:
It seems that none of the wake steering examples currently include any flow visualizations to demonstrate the deflection of the wake. This small PR makes a change to examples_control_optimization/001_opt_yaw_single_ws.py to add such visualization.

There _is_ flow visualization under yaw misalignment in examples_visualizations/001_layout_visualizations.py, but this is not a very intuitive place to go look for that example and is also included in a figure with a lot of other things going on.

A new figure is now generated by 001_opt_yaw_single_ws.py:

yaw_optimization


paulf81 (approved):
Pulled it down and example worked on my machine

Issue #1004, opened 2024-10-21
Bug report: FLORIS 2.5 is not able to handle ngrid=1 correctly


gogannes:
# FLORIS 2.5 is not able to handle ngrid=1 correctly

The `ngrid` option does not work correctly for `ngrid=1`.

I know that most people may not use 2.4 anymore, but for those who may have to use it still in some cases, this issue may help if ngrid=1 is chosen.

I believe this isn't an issue with FLORIS 4.x anymore, as the similar issue is mentioned within the comments of `grid.py`: "If a grid resolution of 1 is selected, create a disc_grid of zeros, as np.linspace would just return the starting value of -1 * disc_area_radius  which would place the point below the center of the rotor. (`turbine_grid_points`)."

## How to reproduce

```
import numpy as np
import floris.tools as wfct

for ngrid in [5, 4, 3, 2, 1]:
    fi = wfct.floris_interface.FlorisInterface("examples/example_input.json")
    fi.reinitialize_flow_field(layout_array=[[0], [0]], wind_speed=6, wind_direction=0)
    fi.change_turbine([0], {"ngrid": ngrid})
    fi.calculate_wake(no_wake=True)
    print(f"turbine power for ngrid={ngrid}: {np.array(fi.get_turbine_power())}")
```

## Relevant output

```
turbine power for ngrid=5: [702300.90048392]
turbine power for ngrid=4: [701715.73960506]
turbine power for ngrid=3: [700546.00421529]
turbine power for ngrid=2: [697099.02181379]
turbine power for ngrid=1: [602547.23424941] << changing from ngrid=2 to ngrid=1 affects turbine power significantly!
```
## How to fix
in `flow_field.py`, replace:
```
yt = np.linspace(
    x2 - pt,
    x2 + pt,
    ngrid,
)
zt = np.linspace(
    x3 - pt,
    x3 + pt,
    ngrid,
)
```
by
```
if ngrid == 1:
    yt = [x2]
    zt = [x3]
else:
    yt = np.linspace(
        x2 - pt,
        x2 + pt,
        ngrid,
    )
    zt = np.linspace(
        x3 - pt,
        x3 + pt,
        ngrid,
    )
```
and the output changes to

```
turbine power for ngrid=5: [702300.90048392]
turbine power for ngrid=4: [701715.73960506]
turbine power for ngrid=3: [700546.00421529]
turbine power for ngrid=2: [697099.02181379]
[...]

PR #1008, opened 2024-10-21
[BUGFIX] Correct grid coordinates for ngrid=1

gogannes:
# [BUGFIX] Correct grid coordinates for ngrid=1

When `ngrid=1` was chosen, np.linspace would just return the starting value of -1 * pt which would not place the point at the center of the rotor. With this fix the center of the rotor is chosen.

## Related issue

#1004 

The generation of the y and z coordinates for each of the turbines is affected in cases where `ngrid=1` is chosen.

See #1004 

Locally all existing tests run without issues. The example case of #1004 now works as expected.

### DIFF
--- a/floris/simulation/flow_field.py
+++ b/floris/simulation/flow_field.py
@@ -100,16 +100,20 @@ class FlowField:
             pt = turbine.rloc * turbine.rotor_radius
 
             xt = [coord.x1 for coord in self.turbine_map.coords]
-            yt = np.linspace(
-                x2 - pt,
-                x2 + pt,
-                ngrid,
-            )
-            zt = np.linspace(
-                x3 - pt,
-                x3 + pt,
-                ngrid,
-            )
+            if ngrid == 1:
+                yt = [x2]
+                zt = [x3]
+            else:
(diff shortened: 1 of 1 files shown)


paulf81 (approved):
The change seems very reasonable to me, and I was able to pull down the PR and run the tests successfully on my mac.  I'm comfortable approving.  Thank you @gogannes !

rafmudaf:
The CI infrastructure has gotten quite stale since v2. To address the failing tests, I suggest to remove macOS from the test matrix and stick with only Linux. In any case, this is consistent with what we've done in v3 and v4.

misi9170:
> The CI infrastructure has gotten quite stale since v2. To address the failing tests, I suggest to remove macOS from the test matrix and stick with only Linux. In any case, this is consistent with what we've done in v3 and v4. You can make that change here: https://github.com/gogannes/floris/blob/v2/.github/workflows/continuous-integration-workflow.yaml#L28.

Thanks Rafael---that sounds good. I'll create a separate PR that does that; and we can merge that first, then this one.

misi9170 (approved):

PR #1017, opened 2024-11-06
[BUGFIX] Address reference wind height warnings raised unnecessarily

misi9170:
#1000 introduced a warning that is raised when `turbine_type` is set without setting the `reference_wind_height`, to let users know that the reference wind height had not been updated. This was pointed out by @cfrontin and also something I ran into recently while working on other aspects of the FLORIS code.

While this works as intended, the warning is now raised in some places (particularly `set_operation_model`) where the `turbine_type` is set implicitly (without direct input from the user). This pull request addresses the unnecessary warning raised by explicitly setting the `reference_wind_height` (to be the same as it was) when `set_operation_model` is called.

I've also updated some examples to add the reference wind height when the turbine model is set, as well as the `calculate_horizontal_plane_with_turbines()` function (which also implicitly sets the `turbine_type`). The following examples are updated:
- examples_turbopark/001_...py
- examples_turbine/003_...py
Note that examples_turbine/001_check_turbine.py also raises the warning; however, in that case, the warning is raised appropriately, and the `assign_hub_height_to_ref_height()` method is then called to "correct" the problem. I therefore left this example as is.

In making the fixes, I also have created a `@property` for the `reference_wind_height` at the `FlorisModel` level to make it easy for users to access.

### DIFF
--- a/examples/examples_turbine/003_specify_turbine_power_curve.py
+++ b/examples/examples_turbine/003_specify_turbine_power_curve.py
@@ -51,6 +51,7 @@ fmodel.set(
     wind_speeds=wind_speeds,
     turbulence_intensities=turbulence_intensities,
     turbine_type=[turbine_dict],
+    reference_wind_height=fmodel.reference_wind_height
 )
 fmodel.run()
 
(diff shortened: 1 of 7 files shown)


paulf81 (approved):
This looks good @misi9170 , I reviewed the code and all makes sense to me.  Thanks!

PR #734, opened 2023-11-09
Bug fix: GCH effective yaw angle in FLORIS v2

fg320:
# Fix incorrect GCH effective yaw angle in FLORIS v2

This pull request addresses the issue in #684, where a bug in FLORIS v2 is pointed out regarding the GCH calculations for the effective yaw angle. The bug is in the `calculate_effective_yaw_angle` method in the file `floris/simulation/wake_deflection/gauss.py`. 

After defining the minimum and maximum limits for the effective yaw angles as
L300 `min_yaw = -45.0`
L301 `max_yaw = 45.0`,
the actual implementation introduces a second negative sign for the negative yaw angle
L347 `if idx_1 == 90 or idx_2 == 90:`
L348   `yaw_effective = max_yaw`
L350 `elif idx_1 == 0 or idx_2 == 0:`
L351   `yaw_effective = -min_yaw`.

This causes the application of a +45 deg effective yaw angle instead of a -45 deg one (refer to the example code and plots below). L351 should be replaced by `yaw_effective = +min_yaw`.

## Related issue
 Issue #684

## Impacted areas of the software
 `floris/simulation/wake_deflection/gauss.py` 

## Example test
 Please refer to the simple example code below, where `gauss_legacy_ste.json` is based on the available examples and included in the last section of this report.

```
# -*- coding: utf-8 --

import os
import numpy as np
import matplotlib.pyplot as plt
import floris
import floris.tools as wfct
from floris.tools.floris_interface import FlorisInterface as FI

# FLORIS BUG yaw effective V2.4 #

# Change working directory to current file directory
cwd = os.path.dirname(os.path.realpath(__file__))
os.chdir(cwd)

# Initialise floris
fi = FI("inputs/gauss_legacy_ste.json")

# Change layout
rows = 5
cols = 1
spacing_x = 7
spacing_y = 5
D = fi.floris.farm.turbines[0].rotor_diameter
x = np.array([[x*spacing_x*D]*cols for x in range(rows)])
x_coordinates = x.flatten()
y_coordinates = np.array([y*spacing_y*D for y in range(cols)]*rows)
fi.reinitialize_flow_field(layout_array=(x_coordinates, y_coordinates))

# Set atmospheric conditions
fi.reinitialize_flow_field(wind_direction=[270.0],
                           wind_speed=[8.0],
                           turbulence_intensity=0.05,
                           wind_shear=0.0,
                           wind_veer=0.0)

#%% Power prediction
yaw_angles = [-25,-25,0,0,0]
[...]

PR #1020, opened 2024-11-08
Remove testing for Mac OS for FLORIS v2

misi9170:
As suggested by @rafmudaf in #1008 , this PR removes Mac OS tests from the testing matrix for FLORIS v2. This will address failing tests in #1008, as well as previous PRs to v2 (e.g. #734 ).

Assuming this is fine, we can merge this PR ahead of #1008 so that tests pass on that PR prior to merger. I will then increment the version number on v2 and create a v2.5.2 release.


rafmudaf (approved):
Looks good to me, thanks @misi9170

PR #1025, opened 2024-11-13
Add Nearest-Neighbor interpolation to heterogeneous wind map

Bartdoekemeijer:
# Add nearest-neighbor interpolation functionality to the HeterogeneousMap object
As the title reads, allows interpolation using NN rather than linear interpolation for heterogeneous background flows.

## Related issue

N/A.

The `heterogeneous_map` and `flow_field` libraries.

The main reason for this is computation time. I don't see it being used very often, but can be useful for .WRGs where you have many sample points and you don't want to spend minutes preprocessing and linearly interpolating to the right points in the grid.

FYI, this PR already includes the minor code changes from #1024 for computational speed.

Here's a simple script to benchmark these changes:
```
import numpy as np
import pandas as pd

from floris import (
    FlorisModel,
    TimeSeries,
    HeterogeneousMap
)


if __name__ == "__main__":
    # Create big grid of wind conditions and wind speeds for which we assume to have evaluated het_map
    wd_grid, ws_grid = np.meshgrid(
        np.arange(0.0, 360.0, 3.0),
        np.arange(6.0, 20.01, 1.0)
    )
    df = pd.DataFrame({"wd": wd_grid.flatten(), "ws": ws_grid.flatten()})
    print(f"We have {df.shape[0]} findices.")

    # Create a grid of sensors throughout the farm in x, y, and z
    xg, yg, zg = np.meshgrid(
        np.linspace(-3000.0, 3000.0, 11),
        np.linspace(-3000.0, 3000.0, 11),
        np.arange(0.0, 350.01, 25.0),
    )
    xg = xg.flatten()
    yg = yg.flatten()
    zg = zg.flatten()
    speedups = np.ones((df.shape[0], len(xg)))
    print(f"We have {len(xg)} number of coordinates with het_map information.")

    # Now create FLORIS and a timeseries object with het_map information
    fmodel.set(
        layout_x = 7.0 * 126.0 * np.arange(30),  # Row of 30 turbines
        layout_y = np.zeros(30),  # Row of 30 turbines
        wind_shear=0.0  # Required when working with 3D het_map objects
    )

    print("")
    for interp_method in ["linear", "nn"]:
        het_map = HeterogeneousMap(
            x=xg,
            y=yg,
            z=zg,
            speed_multipliers=speedups,
            wind_directions=wd_grid.flatten(),
            wind_speeds=ws_grid.flatten(),
            interp_method=interp_method,
        )
        ts = TimeSeries(
            wind_directions=wd_grid.flatten(),
            wind_speeds=ws_grid.flatten(),
            turbulence_intensities=0.06 * np.ones(len(wd_grid.flatten())),
            heterogeneous_map=het_map,
        )

        t0 = timerpc()
        fmodel.set(wind_data=ts)
        print(f"Time spent for fmodel.set() ({interp_method}): {timerpc() - t0:.2f} s")
        fmodel.core.initialize_domain()
        print(f"Time spent for fmodel.set() and .initialize_domain() ({interp_method}): {timerpc() - t0:.2f} s")

```

Which results in
```
We have 1800 findices.
We have 1815 number of coordinates with het_map information.

Time spent for fmodel.set() (linear): 1.24 s
Time spent for fmodel.set() and .initialize_domain() (linear): 81.80 s

Time spent for fmodel.set() (nn): 0.67 s
Time spent for fmodel.set() and .initialize_domain() (nn): 1.26 s
```

Basically bringing time for initializing the domain from 80 seconds to 1 second.


paulf81:
This is looking really good @Bartdoekemeijer , thank you!  My thinking is we should take the following steps:

1) Merge #1024 into develop
2) Merge that code back into here as it needs to blend with the changes to flow_field
3) Add some tests of using the different interpolators

Bartdoekemeijer:
> This is looking really good @Bartdoekemeijer , thank you! My thinking is we should take the following steps:
> 
> 1. Merge [Reduce computation time massively in large het_map objects #1024](https://github.com/NREL/floris/pull/1024) into develop
> 2. Merge that code back into here as it needs to blend with the changes to flow_field
> 3. Add some tests of using the different interpolators

Yes exactly, that sounds good to me! Maybe one test could be to compare accuracy and speed:
1. Computation time and AEP without speedup map.
2. Computation time and AEP with speedup map (linear).
3. Computation time and AEP with speedup map (nearest-neighbor).

floris/core/flow_field.py
```diff
@@ -301,11 +307,19 @@ def generate_heterogeneous_wind_map(self):
...
+                speed_multipliers[0],
+                fill_value=1.0,
+                interp_method=interp_method,
             )
             # Copy the interpolant for each findex and overwrite the values
             for findex in range(self.n_findex):
-                interp_3d.values = speed_multipliers[findex, :].reshape(-1, 1)
+                if interp_method == "linear":
+                    interp_3d.values = speed_multipliers[findex, :].reshape(-1, 1)
+                else:
```
paulf81: Maybe we should make this more explicit to `elif`?

paulf81:
ok @Bartdoekemeijer  and @misi9170 , I think this one caught up with the new develop, but I had to make a number of changes so wouldn't mind a check over that this matches your original intention @Bartdoekemeijer .

One hanging questions (in addition to comment above), is whether we have a behavior issue with the lack of fill value in `NearestNDInterpolator`.  Meaning, the current HetMap assumes no speed-ups outside of region defined by the points.  I think then in `NearestNDInterpolator` this behavior will be different.  Maybe we just need to call that out in a few places?  I think remaining steps in my mind:

- [ ] @Bartdoekemeijer review my updates
- [ ] Decide on above question
- [ ] Maybe add to het docs

Bartdoekemeijer:
Just pulled this down and tested it with my in-house scripts. Works great, thank you Paul! I have no strong opinion on a `fill_value` with `nearest`. Maybe to stay in line with how SciPy defines linear and nearest interpolation, better to ignore the `fill_value` for `nearest` and just raise a warning?

paulf81:
that sounds reasonable @Bartdoekemeijer , @misi9170 maybe could discuss with you wrapping up this one on a preferred behavior.  I think a warning that using this interpolator option means the points outside the convex hull will reflect the speed-up of the nearest point, rather than forced to 1.0 as in the default `LinearNDInterpolator` method.  Since it's not a default it feels more fair it can have behavior changes since you actively select it.

paulf81:
ok, instead of a warning, I just state this in the docstring and docs page. @misi9170 I think ready for your review

misi9170:
I'm sorry that I sat on this for so long. I was originally a bit skeptical as @paulf81 noted about the "outside of boundary" interpolation that the NN approach could generate, but I'm coming around. I'm going to try this out a bit in the coming days and hope to be able to provide a full review soon.

misi9170:
Noting here that `Heterogeneous_Map.plot_single_speed_multiplier()` doesn't necessarily provide the expected output, since it doesn't actually use the interpolation method specified.

Also, we may want to skip producing [this warning](https://github.com/NREL/floris/blob/a6fc5d35aa32614edc450dc399c42af60a816887/floris/core/flow_field.py#L184) for the nearest neighbor case, as it makes less sense there.

misi9170:
The behavior of the `NearestNDInterpolator` can be a little strange for query points that are directly between specified points and seems to depend on the order that the points are specified. For example:
```python
from scipy.interpolate import NearestNDInterpolator

interp = NearestNDInterpolator([(0.0, 0.0),  (1.0, 0.0)], [2.0, 3.0])
print(interp((0.49, 0.0))) # Produces 2.0
print(interp((0.50, 0.0))) # Produces 2.0
print(interp((0.51, 0.0))) # Produces 3.0

# Switch order in which points are specified and requery
print("\n")
interp = NearestNDInterpolator([(1.0, 0.0),  (0.0, 0.0)], [3.0, 2.0])
print(interp((0.49, 0.0))) # Produces 2.0
print(interp((0.50, 0.0))) # Produces 3.0
print(interp((0.51, 0.0))) # Produces 3.0
```
This isn't a problem necessarily, but something to watch out for.

misi9170 (approved):
@Bartdoekemeijer @paulf81 

I've gone through this and made a few changes throughout as I saw fit.
- No warning raised for "out of bounds" points in NN interpolation
- Shuffle interpolation method higher into `HeterogeneousMap` documentation and explicit about "out of bounds" approach. Also add the interpolation method to `HeterogeneousMap.__str__`.
- Test for warnings
- Updates to pass `interp_method` along on `heterogeneous_inflow_config` when creating visualizations
- Updates to `HeterogeneousMap.plot_single_speed_multiplier()`

@paulf81 , in `HeterogeneousMap.plot_single_speed_multiplier()`, I removed an edge case where there was a `try`/`except` statement that used NN for visualization even if the `interp_method` is `linear`. I didn't feel very comfortable about that, and I felt we didn't need it so I removed it, but happy to discuss that further.

A nice way to check visualizations is to specify `interp_method = "nearest"` in the instantiation of `HeterogeneousMap` in examples/examples_heterogeneous/003_heterogeneous_speedup_by_wd_and_ws.py, or by adding `interp_method = "nearest"` to the `heterogeneous_inflow_config` dict in examples/examples_heterogeneous/002_heterogeneous_using_wind_data.py.

@paulf81 I'm going to request a re-review from you now that I've made changes. @Bartdoekemeijer , feel free also to comment on any of the above.

floris/core/flow_field.py
misi9170: Thanks for this comment, @paulf81 . This portion of the code is just meant to be reshaping the `.values` attribute to match what it is "supposed to be" for the different interpolator types; I've now updated this to simply match the shape of the originally generated `interp3d`, which is clearer and avoids the need to check what the `interp_method` is.

misi9170:
Thanks for the review @paulf81 . Assuming all tests pass, I plan to merge this PR into develop so that it included in the next minor release of FLORIS.

PR #1026, opened 2024-11-13
Move all metadata to pyproject.toml

paulf81:
# Move all metadata to pyproject.toml

Following examples such as:

https://packaging.python.org/en/latest/guides/modernize-setup-py-project/

This PR moves all metadata into pyproject.toml.  Specifically:

- All metadata now in pyproject.toml
- setup.py and version.py deleted
- References to setup.py and version.py updated according to guide above


floris/__init__.py
```diff
@@ -1,9 +1,9 @@
 
+from importlib.metadata import version
 from pathlib import Path
 
 
-with open(Path(__file__).parent / "version.py") as _version_file:
-    __version__ = _version_file.read().strip()
+__version__ = version("flasc")
```
rafmudaf: :open_mouth:

pyproject.toml
```diff
@@ -2,6 +2,64 @@
...
+name = "floris"
+version = "4.2"
+description = "A controls-oriented engineering wake model."
+readme = "README.md"
+requires-python = ">=3.9"
+authors = [
+    { name = "Rafael Mudafort", email = "rafael.mudafort@nrel.gov" },
+    { name = "Paul Fleming", email = "paul.fleming@nrel.gov" },
+    { name = "Michael (Misha) Sinner", email = "Michael.Sinner@nrel.gov" },
+    { name = "Eric Simley", email = "Eric.Simley@nrel.gov" },
```
rafmudaf, line 15: Probably add Chris, too

setup.py
```diff
@@ -1,89 +0,0 @@
...
-    version=VERSION,
-    description=DESCRIPTION,
-    long_description=DESCRIPTION,
-    long_description_content_type="text/markdown",
-    author=AUTHOR,
-    author_email=EMAIL,
-    python_requires=REQUIRES_PYTHON,
-    url=URL,
-    packages=find_packages(exclude=["tests", "*.tests", "*.tests.*", "tests.*"]),
-    package_data={
```
rafmudaf, line 70: Don't forget about this part

rafmudaf (changes requested):

floris/__init__.py
paulf81: Fixed sorry for that one!

setup.py
paulf81: Good catch, thank you, think I've got it back in

pyproject.toml
paulf81: Added!

paulf81:
@rafmudaf , 3.8 is the only failing test, but 3.8 is also end of life:
https://devguide.python.org/versions/
Should I just remove it?

rafmudaf:
Changing the minimum Python version is kind of a big decision, so I wouldn't make that change just because a test is failing. However, whether to support Python 3.8 is a fair question. It may be worth asking for some community input.

rafmudaf:
In this case, the error is because you specified Python >= 3.9, but the GitHub Actions job is testing Python 3.8. So the fix as it relates to the scope of this pull request is to make use `requires-python = ">=3.8"`.

paulf81:
> In this case, the error is because you specified Python >= 3.9, but the GitHub Actions job is testing Python 3.8. So the fix as it relates to the scope of this pull request is to use `requires-python = ">=3.8"`.

You're right, think I've fixed this now

misi9170:
Regarding python versions, just wanted to remind of this #953

It appears that FLORIS doesn't fully work with python 3.8 and 3.9 on Windows

misi9170:
> Nice catch @misi9170, I think I understand @rafmudaf 's point to be that raising the python requirement for FLORIS is a valid discussion but should not be done within the context of this PR as it's out of scope

Sounds good, agreed.

.github/workflows/python-publish.yml
```diff
@@ -20,11 +20,11 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install setuptools wheel twine
+        pip install build twine
```
misi9170, line 23: I think this is the change that hung us up over on FLASC---presumably as long as thee `python -m build` line runs below, it should upload to PyPI correctly?

misi9170:
@paulf81 I was able to pip install this into a new conda environment and ran an example without issue.

.github/workflows/python-publish.yml
paulf81: That's right, I think I've gotten to learn from that and now this should go off fine, easy to test locally too but you don't want to run the twine call

misi9170: Yeah, ok, makes sense

rafmudaf: If you want to fully test it, you could [use the TestPyPI instance](https://packaging.python.org/en/latest/guides/using-testpypi/). If you did that, you could create a release on your fork and see that it uploads correctly without affecting anything on the primary repository.

Also heads up that this pull request is pointing to `main` rather than `develop`. Maybe that was intentional, but just pointing it out in case it wasn't.

paulf81: Should've been develop, moved over now

misi9170 (approved):

Issue #1022, opened 2024-11-08
[BUG] Yaw optimization ignores provided `power_setpoints`


misi9170:
# Yaw optimization ignores provided power setpoints
As raised by @MiguelMarante in #1016, `YawOptimizerSR` (as well as other possibly other yaw optimization routines) do not correctly handle other controls being set, in particular, `disable_turbines` (which is a wrapper for near-zero `power_setpoints`).

The issue is that (several) local copies of the `FlorisModel` is made within the yaw optimizer, and the `power_setpoints` aren't copied over correctly. 

## How to reproduce
The following code demonstrates the problem (thanks to @MiguelMarante !)
```
import numpy as np

from floris import FlorisModel
from floris.optimization.yaw_optimization.yaw_optimizer_sr import YawOptimizationSR

# Create floris object
yaml_file = 'inputs/emgauss.yaml'
fmodel = FlorisModel(yaml_file)

# Set wind conditions
fmodel.set(wind_directions=[270.]*2, wind_speeds=[8.]*2, turbulence_intensities=[.06]*2)

# Disable turbines
fmodel.set_operation_model('mixed')
fmodel.set(disable_turbines=np.array([[False, True, False], [True, False, False]]))

# Run optimization and print results
yaw_opt = YawOptimizationSR(fmodel)
df_opt = yaw_opt.optimize()
print(df_opt.yaw_angles_opt)
```

## Relevant output
Output:
```
0    [25.0, 22.65625, 0.0]
1    [25.0, 22.65625, 0.0]
Name: yaw_angles_opt, dtype: object
```
The expected output is not totally clear; the disabled turbine's angle is not well defined (since it produces no thrust and no power, any yaw angle should be optimal). However, we might expect the middle turbine to have a yaw angle of 25 degrees for the second case.

## Floris version
v4.1, v4.2

misi9170:
Addressed in #1031 (previously #1027)

PR #1027, opened 2024-11-15
Allow yaw optimization with disabled turbines

misi9170:
In #1016, @MiguelMarante raised the issue that the Serial Refine yaw optimizer ignores that some turbines are disabled. I converted this to an issue in #1022 . 

This pull request enables that behavior, both for the Serial Refine routine as well as the Geometric yaw optimizer. Note that yaw optimization using Scipy is still not supported for the `"mixed-operation"` model (I looked into enabling this as well, but this will take more time and is not the priority).

### Affected areas of the code
In making the necessary changes to the yaw optimization routines, I also had to make the following changes:
- floris/floris_model.py: `FlorisModel.set()` was incorrectly treating lower power setpoints as though they were the default, which caused them to be ignored and reset when `set()` is called again. This was introduced in #823 by an erroneous commit of mine, a00bff8. This bug meant that calling `set(disabled_turbines=[...])` followed by any other `set()` call would _drop_ the information about disabling turbines.
- floris/core/turbine/operation_models.py: Clean up of the `MixedOperationModel` to work better with yaw optimization.

I've also added a new example, examples/examples_control_optimization/008_....py, which is based heavily on @MiguelMarante 's example script in #1016.

### Notes
In the yaw optimization routines, the solution involves passing `power_setpoints` through from the users `fmodel` to the `fmodel_subset` attributes on the yaw optimization routines. While this works fine, it is yet another example of where control setpoints have to be piped through various pieces of the FLORIS code. To enable other setpoints to be used while optimizing yaw, this process will have to be repeated. We could consider moving to some sort of high-level data structure (possible just a dictionary) for the control setpoints, so that they can be lumped together and passed through the FLORIS code more readily.

### Results
Running the new example produces (removing the progress output from Serial Refine):
```
Serial Refine optimized yaw angles (all turbines active) [deg]:
 0    [25.0, 22.65625, 0.0]
1    [25.0, 22.65625, 0.0]
Name: yaw_angles_opt, dtype: object

Geometric optimized yaw angles (all turbines active) [deg]:
 0    [19.9952335557674, 19.9952335557674, 0.0]
1    [19.9952335557674, 19.9952335557674, 0.0]
Name: yaw_angles_opt, dtype: object

Serial Refine optimized yaw angles (some turbines disabled) [deg]:
 0    [20.3125, 0.0, 0.0]
1      [0.0, 25.0, 0.0]
Name: yaw_angles_opt, dtype: object

Geometric optimized yaw angles (some turbines disabled) [deg]:
 0    [14.990467111534794, 0.0, 0.0]
1      [0.0, 19.9952335557674, 0.0]
```
which is in line with expectations, see #1022.

### DIFF
--- a/examples/examples_control_optimization/008_optimize_yaw_with_disabled_turbines.py
+++ b/examples/examples_control_optimization/008_optimize_yaw_with_disabled_turbines.py
@@ -11,11 +11,21 @@ from floris import FlorisModel
 from floris.optimization.yaw_optimization.yaw_optimizer_geometric import YawOptimizationGeometric
 from floris.optimization.yaw_optimization.yaw_optimizer_sr import YawOptimizationSR
 
+
 # Load a 3-turbine model
 fmodel = FlorisModel("../inputs/gch.yaml")
 
 # Set wind conditions to be the same for two cases
 fmodel.set(wind_directions=[270.]*2, wind_speeds=[8.]*2, turbulence_intensities=[.06]*2)
+
+# First run the case where all turbines are active and print results
+yaw_opt = YawOptimizationSR(fmodel)
+df_opt = yaw_opt.optimize()
+print("Serial Refine optimized yaw angles (all turbines active) [deg]:\n", df_opt.yaw_angles_opt)
+
+yaw_opt = YawOptimizationGeometric(fmodel)
+df_opt = yaw_opt.optimize()
+print("\nGeometric optimized yaw angles (all turbines active) [deg]:\n", df_opt.yaw_angles_opt)
 
 # Disable turbines (different pattern for each of the two cases)
 # First case: disable the middle turbine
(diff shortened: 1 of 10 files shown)


paulf81 (approved):
This all looks good @misi9170 , I left a few small comments.  Going to approve but one additional thought I had is we could perhaps add a small test or two to ensure we keep this desired behavior, but isn't clearly necessary

misi9170:
> This all looks good @misi9170 , I left a few small comments. Going to approve but one additional thought I had is we could perhaps add a small test or two to ensure we keep this desired behavior, but isn't clearly necessary

Thanks @paulf81 --- testing is a good idea, I'll write a small test to check it. I don't see any comments though---perhaps those got lost somehow?

paulf81:
Thanks @misi9170 good to go on my end!

misi9170:
@paulf81 I have now added a small piece of code in the constructor of the base `YawOptimization` class that automatically assigns minimum and maximum yaw angles of zero to disabled turbines. This does help make the output yaw angles more intuitive---thanks for the suggestion.

misi9170:
I incorrectly merged this PR directly into the main branch. I have forced pushed the reversion back to the previous commit on main, so main no longer contains this PR's changes. I've opened a new PR #1031 to merge this code into develop, as intended.

PR #1019, opened 2024-11-07
Test on all supported Python versions

rafmudaf:
In the [Python project configuration](https://github.com/NREL/floris/blob/main/setup.py#L13), FLORIS supports Python versions `>=3.8.0`. However, the automated checks through GitHub Actions are testing inconsistent and incomplete versions, as shown below.

| workflow | 3.8 | 3.9 | 3.10 | 3.11 | 3.12 | 3.13 |
| --------- | --- | --- | --- | --- | --- | --- |
| [check-working-examples](https://github.com/NREL/floris/blob/main/.github/workflows/check-working-examples.yaml) |  | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |  |  |
| [continuous-integration-workflow](https://github.com/NREL/floris/blob/main/.github/workflows/continuous-integration-workflow.yaml) | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |  |  |
| [deploy-pages](https://github.com/NREL/floris/blob/main/.github/workflows/deploy-pages.yaml)|  |  | :heavy_check_mark: |  |  |  |
| [quality-metrics-workflow](https://github.com/NREL/floris/blob/main/.github/workflows/quality-metrics-workflow.yaml) |  |  | :heavy_check_mark: |  |  |  |

This pull request updates these workflows to support Python v3.13 and moves the pages build and speed checks to 3.13. The version changes are captured below.

| workflow | 3.8 | 3.9 | 3.10 | 3.11 | 3.12 | 3.13 |
| --------- | --- | --- | --- | --- | --- | --- |
| [check-working-examples](https://github.com/NREL/floris/blob/main/.github/workflows/check-working-examples.yaml) | :heavy_plus_sign: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_plus_sign: | :heavy_plus_sign: |
| [continuous-integration-workflow](https://github.com/NREL/floris/blob/main/.github/workflows/continuous-integration-workflow.yaml) | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_plus_sign: | :heavy_plus_sign: |
| [deploy-pages](https://github.com/NREL/floris/blob/main/.github/workflows/deploy-pages.yaml)|  |  | :heavy_minus_sign: |  |  | :heavy_plus_sign:  |
[...]

PR #1034, opened 2024-11-20
Fix docs

paulf81:
# Fix docs

Recently the documentation failed to build:

https://github.com/NREL/floris/actions/runs/11938591041

I believe I've traced the issue to the recent version bump of autoyaml to 1.1.2.  Locally at least, pinning to 1.1.1 resolves the issue.  I've also posted a question as an issue here:

https://github.com/Jakski/sphinxcontrib-autoyaml/issues/26

I think the PR should fix it in github actions as well.  Then also:

- ~I set back python to 3.10 for building docs as this has worked in the past~
- Fix a minor test issue that came up in the course of this.

### DIFF
--- a/.github/workflows/deploy-pages.yaml
+++ b/.github/workflows/deploy-pages.yaml
@@ -18,7 +18,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
-        python-version: "3.10"
+        python-version: "3.13"
 
     - name: Install dependencies
       run: |
--- a/.github/workflows/deploy-pages.yaml
+++ b/.github/workflows/deploy-pages.yaml
@@ -18,7 +18,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
-        python-version: "3.13"
+        python-version: "3.10"
 
     - name: Install dependencies
       run: |
--- a/.github/workflows/deploy-pages.yaml
+++ b/.github/workflows/deploy-pages.yaml
@@ -2,10 +2,10 @@ name: deploy-pages
 
 on:
   push:
-    branches:
-    - develop
-    paths:
-    - docs/**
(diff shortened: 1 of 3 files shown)


misi9170:
@rafmudaf in #1019, we moved to building the docs with python 3.13 rather than 3.10---was there a specific need for that, or was it just to keep things up to date?

rafmudaf:
The intent was to use a consistent Python version across all CI jobs aside from the automated tests which has a matrix of Python versions. I chose 3.13 only because it's the latest.

As for the failing build, @paulf81 noted it was due to a version incompatibility with sphinxcontrib-autoyaml. Is it also related to the Python version?

paulf81:
> The intent was to use a consistent Python version across all CI jobs aside from the automated tests which has a matrix of Python versions. I chose 3.13 only because it's the latest.
> 
> As for the failing build, @paulf81 noted it was due to a version incompatibility with sphinxcontrib-autoyaml. Is it also related to the Python version?

You're right, the python version ended up being a red herring.  I've restored 3.13 and checked locally and docs build.  I also have a response on the autoyaml issue that we should stay pinned and they'll let us know when ok to update.

misi9170:
Thanks @rafmudaf , @paulf81 ! I'll try building locally this morning and hopefully we can get this merged.

rafmudaf:
If you want to test it on GitHub, just change the Actions job to build on any event:
```yaml
on:
  push:
#    branches:
#    - develop
#    paths:
#    - docs/**
```

 and comment out the part where it uploads to GitHub pages:
```
    # Push the book's HTML to github-pages
#    - name: GitHub Pages action
#      uses: peaceiris/actions-gh-pages@v3.6.1
#      with:
#        github_token: ${{ secrets.GITHUB_TOKEN }}
#        publish_dir: ./docs/_build/html
```

PR #1035, opened 2024-11-21
Add versions to optional dependencies

paulf81:
This small pull request does two things:

1) Removes the pin on autoyaml put in place in #1034 now that the issue is resolved (see https://github.com/Jakski/sphinxcontrib-autoyaml/issues/26)

  -- Confirm docs build locally

2) Adds versions to optional dependencies as a partial guard against future major version jumps (wouldn't have helped here but good practice!)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,17 +42,17 @@ dependencies = [
 
 [project.optional-dependencies]
 docs = [
-    "jupyter-book",
-    "sphinx-book-theme",
-    "sphinx-autodoc-typehints",
-    "sphinxcontrib-autoyaml==1.1.1",
-    "sphinxcontrib.mermaid",
+    "jupyter-book~=1.0",
+    "sphinx-book-theme~=1.0",
+    "sphinx-autodoc-typehints~=2.0",
+    "sphinxcontrib-autoyaml~=1.0",
+    "sphinxcontrib.mermaid~=1.0",
 ]
 develop = [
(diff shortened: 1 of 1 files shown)


paulf81:
python 3.8 doesn't support the recent major version of pre-commit, I think my preference is to drop 3.8 since it's end of life: https://devguide.python.org/versions/, but we could also give the pre-commit dependency more options,

misi9170:
My feeling is that dropping 3.8 is appropriate since it is end of life, but @rafmudaf also suggested we could have a more dedicated discussion. Is there also an option to remove tests for 3.8, but not explicitly _require_ that users are at python >= 3.9? What is the usual protocol on that?

paulf81:
I'm ok with that approach, but it seems it could only be transitory status, where we allow 3.8 but can't guarantee tests pass

misi9170:
Right. Perhaps that's not good form---perhaps we should strictly require 3.9, I'm not sure.

paulf81:
that's my vote, since it's end of life it seems that is the right move,

misi9170:
I've now opened a PR that removes support for python version 3.8 #1042 . Once we are happy with that and it is merged, we can merge develop back into this branch and tests should pass.

misi9170 (approved):
I also tried running the example with 002 having a 5 degree jump (rather than 45 degrees) so that the change is not so visibly obvious, and that also worked well.

PR #1040, opened 2024-11-27
Support initializing FLORIS with default values

rafmudaf:
# Add an option to initialize FLORIS with default options for use as a library

When used as a library in third-party software, the calling-code typically must carry along a version of the FLORIS input files or require them as user input in order to initialize FLORIS. Often, the inputs are immediately changed by the calling-code such as in an optimization routine. This pattern leads to unnecessary complexity when integrating FLORIS in third-party software, and it could be avoided by providing an option to initialize FLORIS with default data with the understanding that much of it will be overwritten. See https://github.com/NREL/floris/discussions/1037.

This pull request modifies the initialization method of `FlorisModel` to load some set of default values when `configuration="defaults"`. Additionally, a method to retrieve the default inputs as a Python dictionary is provided so that it can be retrieved and modified prior to initialization from the calling code. This is required, for example, if the calling code will change a value that isn't supported in `FlorisModel.set(...)` such as the wake model parameters. I've also added a documentation block in the Advanced Concepts section.

### Another input file
I considered setting defaults through the `attrs.field` function, but decided against this since it would spread the set of default parameters over many source code files. Instead, they're all contained in one file, `floris/default_inputs.py`. It hurts my heart to add yet another input file to the FLORIS repository, and I apologize in advance to any future developer who has to update these 💔.

### What are good defaults?
I chose the GCH example input file as the defaults. Is that reasonable?

### Guessable
Is this confusing to users who use FLORIS directly rather than through another tool? If so, how can it be made more explicit?

### DIFF
--- a/docs/advanced_concepts.ipynb
+++ b/docs/advanced_concepts.ipynb
@@ -96,15 +96,92 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## FLORIS as a library\n",
+    "\n",
+    "FLORIS is commonly used as a library in other software packages.\n",
+    "In cases where the calling-code will create inputs for FLORIS rather than require them from the\n",
+    "user, it can be helpful to initialize the FLORIS model with default inputs and then\n",
+    "change them in code.\n",
+    "In this case, the following workflow is recommended."
+   ]
+  },
+  {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 3,
    "metadata": {},
    "outputs": [],
-   "source": []
+   "source": [
+    "import floris\n",
+    "\n",
+    "# Initialize FLORIS with defaults\n",
+    "fmodel = floris.FlorisModel(\"defaults\")\n",
+    "\n",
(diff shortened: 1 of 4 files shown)


paulf81 (approved):
hi @rafmudaf this looks good to me.  I checked the code changes and ran the tests.  Thanks for adding additional documentation.  Finally, just wanted to note I agree with GCH as default since that is also default in examples.

rafmudaf:
TODO Add an option to print some or all default values

rafmudaf:
@misi9170 I've added a function here to display the current configuration in a short and long form. See https://github.com/rafmudaf/floris/blob/defaults_init/docs/advanced_concepts.ipynb for an example.

misi9170 (approved):
This looks great, thanks @rafmudaf . I made a small change to re-add the `print_dict` method (with a deprecation warning) as a wrapper to `show_config` in case anyone was using it, I don't want to introduce a breaking change. 

Assuming you are happy with that, I can get this merged in.

Bartdoekemeijer:
FYI, this is super helpful for me! I've been copying the default input scripts between different repositories.

PR #1044, opened 2024-12-05
Improve error check for FlorisModel merge function

rafmudaf:
# Expand error checking
The `FlorisModel.merge_floris_models` function currently checks the input argument's first element for the correct type, but it requires that all elements be of the same type. See https://github.com/NREL/floris/pull/866#discussion_r1871854125.

This pull request adds a unit test to show that this error is not currently caught. However, another error is raised since the function will proceed assuming that all elements in the list are `FlorisModel`. Then, the pull request modifies the error checking line to ensure all elements are indeed `FlorisModel`.


rafmudaf:
I've opened this pull request prior to committing the fixed lines of code so that reviewers can see the red X followed by the green tick in the checks. The first commit exposes the issue in the tests, and the following commit fixes the issue.

misi9170 (approved):
Thanks @rafmudaf !

PR #1031, opened 2024-11-18
Allow yaw optimization with disabled turbines (into develop)

misi9170:
I erroneously merged #1027 into the main branch, rather than develop. I have undone the bad merge into main, and this PR is the equivalent for merging into develop. See discussion on #1027 for PR details.

Issue #974, opened 2024-08-30
Allow "unsetting" non-critical keyword arguments to `FlorisModel.set()`


misi9170:
# Provide ability to "unset" certain elements of the `FlorisModel`
When keyword arguments are set on the instantiated `FlorisModel` using the `set()` method, they are remembered. To override a previously set variable, one most override that variable with a new call to `set()`. However, for some arguments, in particular, `heterogeneous_inflow_config`, it would be nice to be able to "unset" or forget a previous setting. `FlorisModel` currently has no way to do this, since passing `None` (e.g. `fmodel.set(heterogeneous_inflow_config=None)`) simply means using the previously set case.

## Proposed solution
A possible solution would be to allow a special (unambiguous) argument to be passed to keywords that could be "unset" in order to trigger the unsetting action. For example, `fmodel.set(heterogeneous_inflow_config=-1)` or `fmodel.set(heterogeneous_inflow_config=False)`.

## Alternatives considered
A second alternative is to define new methods on `FlorisModel` for the unsetting procedure. This is actually something that we already have for unsetting previously set control actions: [`FlorisModel.reset_operation()`](https://github.com/NREL/floris/blob/main/floris/floris_model.py#L456). However, if we go down this route, we'll need a new method for each non-critical variable, which seems a little annoying (especially if these changes are to be reflected in `UncertainFlorisModel` and `ParallelFlorisModel`).

## Current workaround
For now, users can "forget" things by simply reinstantiating their `FlorisModel` from scratch.

## Additional context
I'd like a general solution here, which requires identifying which of the arguments to `FlorisModel.set()` are valid candidates for "unsetting". I would think at least the following:
- `heterogeneous_inflow_config`
- Control setpoints (`yaw_angles`, `power_setpoints`, `awc_modes`, `awc_amplitudes`, `awc_frequencies`, `disable_turbines`)
- _possibly_ `wind_data`, `turbine_type` and `turbine_library_path`, `solver_settings`

misi9170:
#959 may be related

PR #1045, opened 2024-12-06
[BUGFIX] Prevent unnecessary warning being raised in yaw optimization procedures

misi9170:
FLORIS [v4.2.1](https://github.com/nreL/floris/releases/v4.2.1) included a fix/enhancement #1031 for the yaw optimization routine. As part of that fix, I changed `self.fmodel_subset = self.fmodel.copy()` to a `self.fmodel_subset = copy.deepcopy(self.fmodel)` at [this line of yaw_optimization_base.py](https://github.com/NREL/floris/blob/v4.2.1/floris/optimization/yaw_optimization/yaw_optimization_base.py#L233). I did this to ensure that control setpoints were carried over to the `fmodel_subset` attribute. While this worked, it also copied over the `wind_data` stored at `self.fmodel._wind_data`.

In itself, that is not really a problem. However, in the yaw optimization routine, the `wind_directions`, `wind_speeds`, and `turbulence_intensities` are `set()` on `self.fmodel_subset`, see [here](https://github.com/NREL/floris/blob/v4.2.1/floris/optimization/yaw_optimization/yaw_optimization_base.py#L358-L364). This overrides the stored `_wind_data`, which again is not an issue, except that it raises a warning because of [these lines on the `FlorisModel` class](https://github.com/NREL/floris/blob/v4.2.1/floris/floris_model.py#L210-L212). The repeated printout of this warning each time `YawOptimization._calculate_farm_power()` is called is tedious and disconcerting. This is pointed out by @ @MiguelMarante in [#1016](https://github.com/NREL/floris/discussions/1016#discussioncomment-11482947)

To see this, one may run example 001_opt_yaw_single_ws.py in examples/examples_control_optimization/, and see the following printout
image
where previously, on v4.2 before #1031 was enacted, the printout was
image
(Note that the warning raised is once after the yaw optimization procedure is complete, but this is due to [this line in the example](https://github.com/NREL/floris/blob/v4.2.1/examples/examples_control_optimization/001_opt_yaw_single_ws.py#L72) and is the kind of situation the warning was intended for.)

To resolve this, I see a range of possible fixes:
1. (currently implemented on this branch): Simply set `self.fmodel_subset._wind_data = None` after the `self.fmodel_subset = copy.deepcopy(self.fmodel)` statement on `YawOptimization`. This works well, because it sets the unused `._wind_data` to `None` and the warning is then not raised. However, it accesses a private attribute, and is therefore pretty clearly a "hacky" solution.

2. Create a method on `FlorisModel`, perhaps called `reset_wind_data()`, that sets the private attribute to `None`, and then call this method after the `self.fmodel_subset = copy.deepcopy(self.fmodel)` statement on `YawOptimization`. However, it should be noted that although this seems similar to [`reset_operation()`](https://github.com/NREL/floris/blob/v4.2.1/floris/floris_model.py#L475), all it would do is override the `_wind_data` attribute with `None`. This is arguably a bit dangerous, because the `wind_speeds`, `wind_directions`, and `turbulence_intensities` that were set by the originally-passed `WindData` object would _not_ be dropped (as intended), so the name and purpose of the method isn't very clear.

3. Simply remove the warning raised in [`FlorisModel`](https://github.com/NREL/floris/blob/v4.2.1/floris/floris_model.py#L210-L212). I'm not sure how useful this warning is in any circumstance; however, there may be some unintended consequences in situations for which the warning was originally intended.

4. A more thorough fix would be to allow `wind_data`, along with other arguments to `FlorisModel.set()`, to be explicitly removed. This is discussed in #974.

@rafmudaf , @paulf81 , I'd like to hear your input on this. We actually originally did not store the `wind_data` on the `FlorisModel` at all, which would also have avoided this issue, but it was added in #849 (along with the warning) to make post-processing methods more straightforward. Looking back at #849, perhaps option (3.) above is not so bad; perhaps I was a little too liberal with my raising of warnings.

### DIFF
--- a/floris/optimization/yaw_optimization/yaw_optimization_base.py
+++ b/floris/optimization/yaw_optimization/yaw_optimization_base.py
@@ -231,6 +231,7 @@ class YawOptimization(LoggingManager):
 
         # Initialize subset variables as full set
         self.fmodel_subset = copy.deepcopy(self.fmodel)
+        self.fmodel_subset._wind_data = None # Accessing private attribute!
         n_findex_subset = copy.deepcopy(self.fmodel.core.flow_field.n_findex)
         minimum_yaw_angle_subset = copy.deepcopy(self.minimum_yaw_angle)
         maximum_yaw_angle_subset = copy.deepcopy(self.maximum_yaw_angle)


misi9170:
Rebasing now to be from main for this bugfix.

misi9170:
After discussing with @rafmudaf , we've decided that the currently-implemented bugfix (1. in the list above) is a reasonable solution. If we find that we are accessing the private `_wind_data` attribute often, we will look into a more rigorous fix.

Issue #1030, opened 2024-11-15
Potential need to update numpy version to 2.0


paulf81:
# Potential need to update numpy version

I've been re-installing FLORIS/python following an OS update and received this error on install:

```
error: Failed to prepare distributions
  Caused by: Failed to download and build `numpy==1.24.4`
  Caused by: Build backend failed to determine requirements with `build_wheel()` (exit status: 1)

[stderr]
Traceback (most recent call last):
  File "<string>", line 8, in <module>
    from setuptools.build_meta import __legacy__ as backend
  File "/Users/pfleming/.cache/uv/builds-v0/.tmpbOAuSo/lib/python3.13/site-packages/setuptools/__init__.py", line 10, in <module>
    import distutils.core
ModuleNotFoundError: No module named 'distutils'
  Caused by: distutils was removed from the standard library in Python 3.12. Consider adding a constraint (like `numpy >1.24.4`) to avoid building a version of numpy that depends on distutils.
```

We use currently `numpy~=1.20` which means we can jump up to v2.  I think for now I can back myself to python 3.11 to get around this issue but wanted to flag the potential.

## Proposed solution
Update numpy requirement to 2.0 and make any required changes

thebaptiste:
Yes, a new release compatible with numpy 2 would be appreciated.
May be you don't need to have a requirement such as `numpy~=2.0` but rather such as `numpy>=1.20`
I'm not found of this kind of dependency requirements (`~=`) preventing any testing with new releases of dependencies (which might possibly succeed)

misi9170:
Addressed in #1051

PR #1042, opened 2024-11-27
Remove support for python 3.8

misi9170:
Python 3.8 is now [at end-of-life](https://test-devguide.readthedocs.io/en/latest/versions/). We intend that FLORIS support only active versions of python. [FLORIS v4.2](https://github.com/NREL/floris/releases/tag/v4.2.1) is the last minor version of FLORIS that supports version 3.8.

This pull request sets the required version of python at 3.9, and updates the testing matrices (#1019) to reflect that 3.8 is no longer supported. The testing matrix is now

| workflow | 3.8 | 3.9 | 3.10 | 3.11 | 3.12 | 3.13 |
| --------- | --- | --- | --- | --- | --- | --- |
| [check-working-examples](https://github.com/NREL/floris/blob/main/.github/workflows/check-working-examples.yaml) | :heavy_minus_sign: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |  :heavy_check_mark: |
| [continuous-integration-workflow](https://github.com/NREL/floris/blob/main/.github/workflows/continuous-integration-workflow.yaml) | :heavy_minus_sign:  | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| [deploy-pages](https://github.com/NREL/floris/blob/main/.github/workflows/deploy-pages.yaml)|  |  |  |  |  | :heavy_check_mark:   |
| [quality-metrics-workflow](https://github.com/NREL/floris/blob/main/.github/workflows/quality-metrics-workflow.yaml) |  |  | |  |  | :heavy_check_mark:  |

PR #1051, opened 2024-12-12
Update numpy req to 2

paulf81:
# Update numpy req to 2

Testing an update based on #1030 where the numpy requirement is pushed up to compatiable version of 2.0.  

An alternative approach would be to span version 1 and 2 using > and < operators, but I understand this to be non-preferred in semantic versioning, since testing would only apply to newer major version, and so the previous major version is not truly supported.

Using this PR also to check if having the requirement at ~=2.0 raises any incompatibilities.

## Related issue
#1030

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -10,6 +10,8 @@ the conversation in [GitHub Discussions](https://github.com/NREL/floris/discussi
 
 ```{note}
 Support for python version 3.8 will be dropped in FLORIS v4.3. See {ref}`installation` for details.
+
+FLORIS v4.3 will also move to requiring `numpy` version 2. See the [numpy documentation for details](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html).
 ```
 
 ## Quick Start
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -14,7 +14,9 @@ and sandboxed environment. The simplest way to get started with virtual environm
 [conda](https://docs.conda.io/en/latest/miniconda.html).
 
 ```{warning}
-Support for python version 3.8 will be dropped in FLORIS v4.3.
+Support for python version 3.8 will be dropped in FLORIS v4.3. 
+
+FLORIS v4.3 will also require `numpy` version 2. See the [numpy documentation for details](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html).
 ```
 
 Installing into a Python environment that contains a previous version of FLORIS may cause conflicts.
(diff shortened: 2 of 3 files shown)


misi9170:
I'm OK with going up to `~=2.0`. We'll also be dropping support for python 3.8 in the next minor FLORIS version #1042 , and this feels like a similar thing. However, there are likely to be users that will still be on numpy 1.x, who may be prevented from upgrading to FLORIS 4.3 if we introduce this requirement.

thebaptiste:
May be you could check it works with numpy 2 (and fix incompatibilities if necessary) but set numpy requirement to `numpy>=1.20` or something like `numpy>=1.20,<2.1` (if its possible to have a code working with both numpy 1 and numpy 2) in order to let users use numpy 1 or numpy 2 ?

thebaptiste:
Here is another problem with requirement `numpy~=1.20` : if you want to use together both floris and numpy 2.0, it's possible ! you have to use... floris 3.5 ! because in floris 3.5 the requirement was `numpy>=1.20`. Awful isn't it ? (it probably doesn't work but it's possible to install !).

I think there will be more difficulties for users if you do not upgrade to numpy 2.x than if you upgrade to numpy 2.x
There are less and less packages not compatible with numpy 2.0, Floris is one of them...
I think it's better to encourage users to upgrade to numpy 2 (for their own good) if they want to use new floris  functionalities than to keep them in their "dead end" with numpy 1.20...

rafmudaf:
Some things to consider:
- I see 2.1 drops support for Python v3.9 (https://numpy.org/devdocs/release/2.1.0-notes.html), how does that impact the Numpy version that FLORIS should support?
- What's the impact to performance? Reading through the Numpy release notes, I think moving to Numpy v2 will significantly improve performance, but it would be good to quantify this
- It's worth double checking that adjacent tools in the WETO Stack also support Numpy v2 (FLASC, OpenOA, anything else that integrates FLORIS -> FLOWERS?)

paulf81:
Ok, I think we're in agreement that we should make the move up to numpy 2, but after some discussions we'd like to do things in particular order.  I will open a separate discussion to track the sequence

paulf81:
@misi9170 , based on #1056 we're now square to merge this PR

misi9170:
I've added a couple of notes to the documentation page to highlight this change. Once this PR is merged, those will stay on the docs pages (at least until we release v4.3, and possible after).

Regarding dropping support for python 3.9---that does seem like a big deal, although for the moment at least, our tests and examples workflows seems to run successfully, so that's a small comfort for users that are still on python 3.9?

misi9170 (approved):

PR #1054, opened 2025-01-09
Feature/load heuristic

paulf81:
# Feature/load heuristic

Very rough draft @misi9170  @ejsimley

### DIFF
--- a/examples/examples_control_types/001b_derating_for_load.py
+++ b/examples/examples_control_types/001b_derating_for_load.py
@@ -1,0 +1,90 @@
+"""Example of de-rating for load
+
+TBD.
+"""
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+from floris import FlorisModel
+
+
+fmodel = FlorisModel("../inputs/gch.yaml")
+
+# Change to the simple-derating model turbine
+# (Note this could also be done with the mixed model)
+fmodel.set_operation_model("simple-derating")
+
+# Convert to a simple two turbine layout with derating turbines
+fmodel.set(layout_x=[0, 126.0 * 5], layout_y=[0.0, 0.0])
+
+
+# Set the wind directions and speeds to be constant over n_findex = N time steps
+N = 50
+fmodel.set(
+    wind_directions=270 * np.ones(N),
+    wind_speeds=7.0 * np.ones(N),
+    turbulence_intensities=0.06 * np.ones(N),
+)
+fmodel.run()
+turbine_powers_orig = fmodel.get_turbine_powers()
+
+# Get the nominal operating power of the upstream turbine
+nom_pow = turbine_powers_orig[0,0]
+
+# # Derate the front turbine from full power to 75 %
+power_setpoints_front = np.linspace(nom_pow,nom_pow *.75,N)
+full_rating = np.ones_like(power_setpoints_front) * 5E6
+
+# Only derate the front turbine
+power_setpoints = np.column_stack([power_setpoints_front, full_rating])
+fmodel.set(power_setpoints=power_setpoints)
+fmodel.run()
+turbine_powers_derated = fmodel.get_turbine_powers()
+
+# Compute the mean power
+power_mean = np.mean(turbine_powers_derated,axis=1)
+
+# Define de-rating as percent reduction
+de_rating = 100 * (nom_pow - power_setpoints_front) / nom_pow
+
+# Grab the load heuristic
+load_h = fmodel._get_turbine_load_h()
+
+# Compute mean load
+load_h_mean = np.mean(load_h,axis=1)
+
+# Plot the results
+fig, axarr = plt.subplots(2,3,sharex=True, sharey='row',figsize=(12,8))
+
+ax = axarr[0,0]
+ax.plot(de_rating, turbine_powers_derated[:,0]/turbine_powers_derated[0,0],'k')
+ax.set_title('Turbine 0')
+ax.set_ylabel('Power (ratio to nomimal)')
+
+ax = axarr[0,1]
+ax.plot(de_rating, turbine_powers_derated[:,1]/turbine_powers_derated[0,1],'k')
+ax.set_title('Turbine 1')
+
+ax = axarr[0,2]
+ax.plot(de_rating, power_mean/power_mean[0],'k')
+ax.set_title('Mean')
+
+ax = axarr[1,0]
+ax.plot(de_rating, load_h[:,0]/load_h[0,0],'k')
+ax.set_ylabel('Load H (ratio to nomimal)')
+ax.set_xlabel('Front De-Rating %')
+
+ax = axarr[1,1]
+ax.plot(de_rating, load_h[:,1]/load_h[0,1],'k')
+ax.set_xlabel('Front De-Rating %')
+
+ax = axarr[1,2]
+ax.plot(de_rating, load_h_mean/load_h_mean[0],'k')
+ax.set_xlabel('Front De-Rating %')
+
+for ax in axarr.flatten():
+    ax.grid(True)
+
+
+plt.show()
--- a/floris/floris_model.py
+++ b/floris/floris_model.py
@@ -36,6 +36,7 @@ from floris.utilities import (
     nested_get,
     nested_set,
     print_nested_dict,
+    rotate_coordinates_rel_west,
 )
 from floris.wind_data import (
     TimeSeries,
@@ -503,8 +504,112 @@ class FlorisModel(LoggingManager):
         self.core.finalize()
 
 
-    ### Methods for extracting turbine performance after running
+    #TODO: This could probably be moved elsewhere
+    #TODO: load_h is a stand-in term for "load heuristic" until something better comes to mind
+    def _get_turbine_load_h(self) -> NDArrayFloat:
 
+        #TODO: Probably these should be function input parameters
+        wake_slope = 0.3
+        max_dist_D = 10.0
+        weight_ws = 1.0
+        weight_ct = 1.0
+        weight_ti = 1.0
+
+        # This one I'm not sure, maybe it will be ambient ti, maybe a 1D TD load_ambient_ti
+        load_ambient_ti = 0.10
+
+        # Store for convenience
+        n_turbines = self.n_turbines
+        n_findex = self.n_findex
+
+        # TODO for now assume the first one
+        D = self.core.farm.rotor_diameters[0,0]
+
+        # Get the wind direction rotated farm layouts with wind flowing left to right
+        coord_rotate = rotate_coordinates_rel_west(self.wind_directions,
+                                                    self.core.grid.turbine_coordinates)
+        x_rot = coord_rotate[0]
+        y_rot = coord_rotate[1]
+
+        # Get all turbine thrust coefficients
+        cts = self.get_turbine_thrust_coefficients()
+
+        # Get ws, was originally thinking to use the actual rotor wind speeds (commented first line)
+        # but think more consistent with load methods to use ambient here since those ambient flows
+        # will surround the wake somehow for load inducing purposes
+        # ws = np.mean(self.core.flow_field.u, axis=(2,3))
+        ws = self.wind_speeds
+        # Make the same for every turbine
+        ws = np.tile(ws[:, np.newaxis], (1, n_turbines))
+
+        # Initialize the load_ti to the load_ambient_ti
+        load_ti = np.ones((n_findex, n_turbines)) * load_ambient_ti
+
+        # For each findex and turbine, compute the "load turbulence"
+        for findex in range(n_findex):
+
+            # t_i is the index of the "current" turbine whose "load_ti" we want to compute
+            for t_i in range(n_turbines):
+                nearest_t_dist = 1 + np.max(x_rot) - np.min(x_rot) # Initialize to the max + 1
+
+                # t_j is the index of the turbine who is proposed to be waking t_i
+                for t_j in range(n_turbines):
+
+                    # If the same turbine skip
+                    if t_i == t_j:
+                        continue
+
+                    # Compute the distance from proposed waking to turbine to the current turbine
+                    # with the distance positive if the current turbine is downstream
+                    x_diff = x_rot[findex, t_i] - x_rot[findex, t_j]
+
+                    # If the proposed waking turbine is downstream skip
+                    if x_diff<=0:
+                        continue
+
+                    # If the distance between the turbines is more than the max skip
+                    if x_diff >= max_dist_D * D:
+                        continue
+
+                    # If this is not closer than previouly found wakes skip
+                    if x_diff >= nearest_t_dist:
+                        continue
+
+                    # Compute the cross-stream distance
+                    y_diff_abs = np.abs(y_rot[findex, t_i] - y_rot[findex, t_j])
+
+                    # If the t_i turbine is above the wake cone of the
+                    # proposed waking turbine skip
+                    if y_diff_abs > D + x_diff * wake_slope:
+                        continue
+
+                    # Set nearest_t_dist to the current value
+                    nearest_t_dist = x_diff
+
+                    # Get the c_t of the proposed waking turbine
+                    c_t_j = cts[findex, t_j]
+
+                    # Get the ambient ws (note all turbines same value
+                    #  but might as well use the current turbine)
+                    ambient_ws = ws[findex,t_i]
+
+                    # Compute the added turbulence
+                    ti_add = ambient_ws / (1.5 + 0.8 * (x_diff/D) / np.sqrt(c_t_j) )
+
+                    # Update the ti_term
+                    load_ti[findex, t_i] = (
+
+                        np.sqrt(ti_add**2 + (load_ambient_ti * ambient_ws)**2) / ambient_ws
+
+                    )
+
+
+
+        # Compute load_h
+        return weight_ws * ws + weight_ct * cts + weight_ti * load_ti
+
+
+    ### Methods for extracting turbine performance after running
     def _get_turbine_powers(self) -> NDArrayFloat:
         """Calculates the power at each turbine in the wind farm.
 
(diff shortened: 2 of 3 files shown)


misi9170:
@paulf81 I've converted this to a draft for now, but feel free to switch it back if/when you're ready for a full review

floris/optimization/load_optimization/load_optimization.py
```diff
@@ -0,0 +1,459 @@
...
+def find_A_to_satisfy_rev_voc_ratio(
+    fmodel: FlorisModel,
+    target_rev_voc_ratio: float,
+    load_ambient_tis: np.array,
+    wake_slope: float = 0.3,
+    max_dist_D: float = 10.0,
+    exp_ws_std: float = 1.0,
+    exp_thrust: float = 1.0,
+):
+    """Find the value of A that satisfies the target revenue to VOC ratio.
```
misi9170, line 329: I think this function needs a bit more documentation to explain what the `target_rev_voc_ratio` is and why one might want to find a value of A to meet that target. I think you mentioned this was useful for initializing the optimizer?

floris/optimization/load_optimization/load_optimization.py
```diff
@@ -0,0 +1,459 @@
...
+            test_power_setpoint[~update_mask, :] = final_power_setpoint[~update_mask, :]
+
+            # Update the final_power_setpoint
+            final_power_setpoint[:, :] = test_power_setpoint[:, :]
+
+            # Update the net_revenue
+            net_revenue[update_mask] = test_net_revenue[update_mask]
+
+    # Return the final power setpoint
+    return final_power_setpoint
```
misi9170, line 459: I think it's common for optimization routines to return both x^* (the solution to the optimization problem, in this case the power setpoints) and v^* (the optimal cost/value). Perhaps we could also return the final value of `net_revenue`?

floris/optimization/load_optimization/load_optimization.py
```diff
@@ -0,0 +1,459 @@
...
+    fmodel.set_operation_model("simple-derating")
+
+    # If initial set point not provided, set to rated (assumed max) power
+    if initial_power_setpoint is None:
+        max_power = get_max_powers(fmodel)
+        initial_power_setpoint = np.tile(max_power, (fmodel.n_findex, 1))
+
+    # Initialize the the test power setpoints
+    test_power_setpoint = initial_power_setpoint.copy()
+    final_power_setpoint = initial_power_setpoint.copy()
```
misi9170, line 404: could `final_power_setpoint` be renamed `optimal_power_setpoint` or `power_setpoint_opt`? "Final" isn't super clear to me

misi9170 (commented):
Hi @paulf81 , Overall this looks good to me. I left a couple of small comments in the load_optimization.py code, and made a few small changes directly for clarity, but overall the functionality looks good. 

I think where I feel there could be improvement is in the examples---I'm not sure that there need to be four separate examples (perhaps two would suffice?) and I'm also feel that they aren't the most intuitive to someone new coming to this. I'll think about that a bit more over the next few days and see if I think of anything.

examples/examples_load_optimization/00_load_ti_example.py
```diff
@@ -0,0 +1,80 @@
...
+ax.set_xlabel("Wind Direction (deg)")
+ax.legend()
+
+# CT SWEEP
+fmodel = FlorisModel(configuration="defaults")
+N = 50
+wind_directions = np.ones(N) * 270.0
+load_ambient_tis = np.ones(N) * 0.1
+power_setpoints = np.linspace(5e6, 1, N)
+power_setpoints_grid = np.column_stack([power_setpoints, power_setpoints * 0 + 1])
```
misi9170, line 55: Is there a reason that the downstream turbine has (essentially) a zero power setpoint here?

tests/load_optimization_test.py
```diff
@@ -0,0 +1,139 @@
...
+from floris.optimization.load_optimization.load_optimization import (
+    compute_farm_revenue,
+    compute_farm_voc,
+    compute_load_ti,
+    compute_net_revenue,
+    compute_turbine_voc,
+    find_A_to_satisfy_rev_voc_ratio,
+    get_max_powers,
+    get_rotor_diameters,
+    optimize_derate,
```
misi9170, line 17: There are not currently any tests for this---perhaps we could add a simple regression test?

floris/optimization/load_optimization/load_optimization.py
misi9170: I've now added this

floris/optimization/load_optimization/load_optimization.py
misi9170: I've added the optimal net revenue as an output (and renamed this function `optimize_power_setpoints`, which feels more explicit to me)

floris/optimization/load_optimization/load_optimization.py
```diff
@@ -0,0 +1,463 @@
...
+    if fmodel.wind_data is None:
+        raise ValueError("FlorisModel must have wind_data to compute net revenue")
+
+    # Ensure that fmodel.wind_data is not None
+    if fmodel.wind_data.values is None:
+        raise ValueError("FlorisModel wind_data.values must be set to compute revenue")
+
+    farm_power = fmodel.get_farm_power()
+    values = fmodel.wind_data.values
+    return farm_power * values
```
misi9170, line 277: Should there be some sort of dt in here? How does that work? Or is the output revenue treated as "per unit time", so to speak?

ejsimley: Could we just document that values are interpreted as revenue per Watt per findex? Same with VOC? Cost per  findex?

floris/optimization/load_optimization/load_optimization.py
```diff
@@ -0,0 +1,546 @@
...
+        power_setpoint_initial (np.array, optional): Initial power setpoint for each turbine.
+            If None, each turbine's rated power will be used. Defaults to None.
+        derating_levels (np.array, optional): Array of derating levels to consider in optimization,
+            represented as fractions of the power_setpoint_initial.
+            Defaults to np.linspace(1.0, 0.001, 5).
+
+    """
+
+    # Ensure we're in derating mode
+    fmodel.set_operation_model("simple-derating")
```
ejsimley, line 482: Are/will there be other derating models besides "simple-derating"? If so, it would be good to document that this module only considers simple-derating.

floris/optimization/load_optimization/load_optimization.py
```diff
@@ -0,0 +1,546 @@
...
+        wake_slope=wake_slope,
+        max_dist_D=max_dist_D,
+        exp_ws_std=exp_ws_std,
+        exp_thrust=exp_thrust,
+    )
+
+    return (farm_revenue.sum() / farm_voc.sum()) / target_rev_voc_ratio
+
+
+def find_A_to_satisfy_target_VOC_per_MW(
```
ejsimley, line 394: I added this function to help when we want to achieve a target VOC/MWh, which is something users might know. It seems like O&M costs are usually report as $/kW capacity/year, but this seems tricky to find unless the set of findices represents 1 year. So VOC/MW/findex seemed like a reasonable alternative (equivalent to VOC/MWh if a findex represents 1 hour of operation).

examples/examples_load_optimization/001_load_ti_and_voc.py
```diff
@@ -0,0 +1,131 @@
+"""Example: Load turbulence intensity
```
ejsimley, line 1: The file name was confusing to me at first because I was expecting this example to "load" or "import" Tis and VOCs. What do you think about changing to "001_investigate_load_ti_and_voc"?

examples/examples_load_optimization/02_sweep_values.py
```diff
@@ -0,0 +1,119 @@
+"""Example: Variable Operating Cost
```
ejsimley, line 1: To make the title more descriptive, how about changing this to "02_sweep_load_ti_and_value"?

examples/examples_load_optimization/02_sweep_values.py
```diff
@@ -0,0 +1,119 @@
...
+# Set the turbine layout to be a simple two turbine layout using the
+# time series object
+fmodel.set(layout_x=[0, D * 7], layout_y=[0.0, 0.0], wind_data=time_series)
+fmodel.run()
+
+# Compute the load turbulence intensity
+voc = compute_turbine_voc(fmodel, 2e-5, load_ambient_tis)
+
+# Compute net revenue
+net_revenue = compute_net_revenue(fmodel, 2e-5, load_ambient_tis)
```
ejsimley, line 52: Note that I changed the ratio between A and values to make net revenue more sensitive to load ambient TI to make the example a little more interesting.

examples/examples_load_optimization/03_row_opt_example.py
```diff
@@ -0,0 +1,149 @@
...
+time_series = TimeSeries(
+    wind_directions=wind_directions, wind_speeds=8.0, turbulence_intensities=0.06, values=values
+)
+
+# Run the FLORIS model
+fmodel.set(wind_data=time_series)
+fmodel.run()
+
+# Set the initial power setpoints as no derating
+initial_power_setpoint = np.ones((N, n_turbines)) * 5e6
```
ejsimley, line 68: Would this be a good place to demonstrate the `get_max_powers` function from the load_optimization module?

examples/examples_load_optimization/03_row_opt_example.py
```diff
@@ -0,0 +1,149 @@
+"""Example: Optimize a row of turbines
```
ejsimley, line 1: This is an interesting example, but I think the plot is showing too many combinations of wind conditions and electricity values. Can this be reduced to somewhere around 50 combinations to make it easier to digest? For example, I tried limiting the wind directions to 248 and 266 degrees and think that already shows all the interesting trends we'd want to highlight.

I also think a little commentary would be helpful to explain why specific turbines are being curtailed/derated in response to the changing conditions. But maybe that's not necessary.

ejsimley (changes requested):
@paulf81, I think this is in pretty good shape overall. I made a bunch of changes to the doc strings and comments to explain arguments and steps in more detail, and added another function for finding the value of A for the VOC equation to target a particular $/MW/findex. I also updated the examples to make the dollar amounts more realistic (in the neighborhood of $20/MWh electricity prices).

I added some minor comments throughout, with the biggest one related to simplifying example 03. Note that I did not leave any comments for examples 01 and 02, thinking we'd use the new example 001 that @misi9170 added. 

I also think the comments in the load_optimization module should explain if/how our method for computing load TI differs from IEC 61400-1 Annex E. For example, does the standard use a wake slope to determine waked turbines (and if so, does it use a slope of 0.3?), or is that something we added?

floris/optimization/load_optimization/load_optimization.py
```diff
@@ -0,0 +1,546 @@
...
+    """Compute the turbine 'load' turbulence intensity for the current layout by combining the
+    'load' ambient turbulence intensity and wake added turbulence following Annex E in the
+    IEC 61400-1 Ed. 4 standard.
+
+    Args:
+        fmodel (FlorisModel): FlorisModel object
+        load_ambient_tis (list or np.array): Ambient 'load' turbulence intensity for each findex,
+            expressed as fractions of mean wind speed
+        wake_slope (float, optional): Wake slope, defined as the lateral expansion of the wake on
+            each side per unit downstream distance along the axial direction. Defaults to 0.3.
```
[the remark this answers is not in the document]
paulf81: I think I got the 0.3 from the FLASC implementation of `get_all_impacting_turbines_geometrical`, and was also refererring to the IEC annex but can't find a supporting reference

floris/optimization/load_optimization/load_optimization.py
```diff
@@ -0,0 +1,546 @@
...
+    x_sorted = np.mean(fmodel.core.grid.x_sorted, axis=(2, 3))
+    y_sorted = np.mean(fmodel.core.grid.y_sorted, axis=(2, 3))
+
+    # Put ct into sorted frame
+    ct_sorted = np.take_along_axis(cts, sorted_indices, axis=1)
+
+    # 2. Iterate over turbines from front to back across findices
+    for t in range(fmodel.n_turbines):
+        # Set D to the diameter of the current turbine
+        D = rotor_diameters[t]
```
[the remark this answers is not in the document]
paulf81: No, and also D is a scalar, where it would have to be n_findex long to match the case where the nth upstream turbine might have a different diameter depending on findex

floris/optimization/load_optimization/load_optimization.py
paulf81: I'll just move to raise error and not automatically change the mode

floris/optimization/load_optimization/load_optimization.py
```diff
@@ -0,0 +1,546 @@
...
+
+def compute_load_ti(
+    fmodel: FlorisModel,
+    load_ambient_tis: np.array,
+    wake_slope: float = 0.3,
+    max_dist_D: float = 10.0,
+):
+    """Compute the turbine 'load' turbulence intensity for the current layout by combining the
+    'load' ambient turbulence intensity and wake added turbulence following Annex E in the
+    IEC 61400-1 Ed. 4 standard.
```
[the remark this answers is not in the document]
paulf81: Tried to clarify a bit here.

floris/optimization/load_optimization/load_optimization.py
```diff
@@ -0,0 +1,546 @@
...
+
+        # Calculate total distance
+        distance = np.sqrt(dx**2 + dy**2)
+
+        # Set the minimum distance mask
+        max_dist_mask = distance < D * max_dist_D
+
+        # Compute the standard deviation of the wind speed owed to this wake following Annex E
+        # of the IEC 61400-1 Ed. 4 standard
+        ws_std = np.where(
```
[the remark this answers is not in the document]
paulf81: Good suggestion, thank you!

floris/optimization/load_optimization/load_optimization.py
```diff
@@ -0,0 +1,546 @@
...
+            wake_cone_mask & downstream_mask & max_dist_mask,
+            ambient_wind_speeds / (1.5 + 0.8 * (distance / D) / np.sqrt(ct_t)),
+            0.0,
+        )
+
+        # Combine with ambient TI to get the total TI from this wake following Annex E
+        # of the IEC 61400-1 Ed. 4 standard
+        ti_add_update = (
```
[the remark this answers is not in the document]
paulf81: Another good suggestion

floris/optimization/load_optimization/load_optimization.py
```diff
@@ -0,0 +1,546 @@
...
+def compute_turbine_voc(
+    fmodel: FlorisModel,
+    A: float,
+    load_ambient_tis: np.array,
+    wake_slope: float = 0.3,
+    max_dist_D: float = 10.0,
+    exp_ws_std: float = 1.0,
+    exp_thrust: float = 1.0,
+):
+    """Compute the turbine Variable Operating Cost (VOC) for each findex and turbine.
```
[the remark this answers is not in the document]
paulf81: Took a shot below

floris/optimization/load_optimization/load_optimization.py
```diff
@@ -0,0 +1,546 @@
...
+
+    In this first approximation, variable operating cost is computed as the
+    product of turbine thrust and wind speed standard deviation.
+
+    Args:
+        fmodel (FlorisModel): FlorisModel object
+        A (float): Coefficient for the VOC calculation
+        load_ambient_tis (list or np.array): Ambient 'load' turbulence intensity for each findex,
+            expressed as fractions of mean wind speed
```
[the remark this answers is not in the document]
paulf81: No, I think it's simpler to keep these totally separate for now

floris/optimization/load_optimization/load_optimization.py
```diff
@@ -0,0 +1,546 @@
...
+    # A should be a float and not a list or array
+    if not isinstance(A, (int, float)):
+        raise ValueError("A (coefficient of VOC) must be a float")
+
+    # Get the ambient wind speed and apply to each turbine per findex
+    ambient_wind_speeds = fmodel.wind_speeds
+    ambient_wind_speeds = np.tile(ambient_wind_speeds[:, np.newaxis], (1, fmodel.n_turbines))
+
+    # Compute the rotor area
+    D = fmodel.core.farm.rotor_diameters[0, 0]
```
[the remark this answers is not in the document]
paulf81: realizing this gets complicated with the sorting, maybe we should simply limit operations to one D farms for now and work out how to do variable D in a future PR

floris/optimization/load_optimization/load_optimization.py
paulf81: We should probably just try to make this clear in the examples since this is just re-using the underlying values architecture

floris/optimization/load_optimization/load_optimization.py
paulf81: This is helpful, thank you for adding!

floris/optimization/load_optimization/load_optimization.py
```diff
@@ -0,0 +1,546 @@
...
+def optimize_power_setpoints(
+    fmodel: FlorisModel,
+    A: float,
+    load_ambient_tis: np.array,
+    wake_slope: float = 0.3,
+    max_dist_D: float = 10.0,
+    exp_ws_std: float = 1.0,
+    exp_thrust: float = 1.0,
+    power_setpoint_initial: np.array = None,
+    derating_levels: np.array = np.linspace(1.0, 0.001, 5),
```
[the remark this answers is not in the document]
paulf81: Yeah, great catch!

examples/examples_load_optimization/001_load_ti_and_voc.py
paulf81: Ok should be a clearer now

floris/optimization/load_optimization/load_optimization.py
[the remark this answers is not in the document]
paulf81: simplified to assuming single D value

misi9170 (approved):

Issue #1039, opened 2024-11-26
Feature request: Enable Cumulative Curl model in `solve_for_points()`


misi9170:
Over in #1038, @reos-aburns pointed out that we have not yet enabled the Cumulative Curl model to be used with the `solve_for_points()` procedure. I believe that this should not be too difficult to enable, and I'm going to look into it to see what I can do.

PR #1058, opened 2025-01-30
Enable solve_for_points with CC model

misi9170:
Addresses #1039.


rafmudaf:
This also fixes places in the CC and TurbOPark solvers where the `awc_modes` settings were unsorted.

rafmudaf (approved):

PR #1060, opened 2025-02-04
Add automatic benchmarking to FLORIS

paulf81:
# Add automatic benchmarking to FLORIS

The PR will supercede #992  and seeks to avoid issues running some of the github actions from a fork.  This shouldn't be an issue going forward since the action is specified to run on develop at a specific time and is not connected to pulls or pushes.

This draft PR is meant to add automatic code benchmarking to FLORIS.  Proposed solution is to use pytest-benchmark to implement set timing tests:

https://pytest-benchmark.readthedocs.io/en/latest/

https://github.com/ionelmc/pytest-benchmark

And then try to schedule some semi-daily execution of these tests with logged performance checks so we can track changes over time.  Here focused on:

https://github.com/benchmark-action/github-action-benchmark

To this end I added a first test to the tests folder including benchmarking to the `tests/` folder and confirm the command line:

`pytest floris_benchmark_test.py `

Produces a benchmark result.  At this point might open up for discussion, or others research:

1. Do these benchmarks belong in tests?  Won't that cause them to be run with normal CI process?
2. What should we test?
3. We should set a test results page
4. Test parallel?
5. Want to track changes coming from our own work, and also those coming from improvement to python, should make sure CI checks most recent version of python?

To include:

- [ ] AEP for 100-turbine farm
- [ ] Analysis for different wake models
- [ ] Parallel FLORIS


misi9170:
@paulf81 heads up that if we want to make this a patch release without including other changes, we'll need to change the merge target to main. Let me know when you're ready for that and we can work through it.

paulf81:
ok @rafmudaf  and @misi9170 , I think there is a bit of an overwrite tension between the deploy pages and this benchmarking in that they both write to gh-pages.  However, I think there is a pretty easy workaround.  If we specify both to deploy within a subdirectory of gh-pages, they will keep out of each others way.   I think we can use `destination_dir` in the deploy pages workflow:

[[actions-gh-pages](https://github.com/peaceiris/actions-gh-pages)](https://github.com/peaceiris/actions-gh-pages?tab=readme-ov-file#%EF%B8%8F-deploy-to-subdirectory-destination_dir)

Maybe `docs/`

And then similar for benchmark.  Then we'd only need to specify the location here:

image

Work for you guys?

paulf81:
Closing in favor of #1062

PR #1064, opened 2025-02-11
Add dependabot dependent version updates

paulf81:
# Add dependabot dependent version updates

This PR introduces automated updating of dependent versions via github's [dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates)

Following basic settings from here:

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates

When this file is in github then dependabot should generate PRs we can consider to update dependencies.  This should save us from having to do this manually.

Could consider patching this into main as well.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+- package-ecosystem: "pip"
+  directory: "/" # Location of package manifests
+  schedule:
+    interval: "weekly"


paulf81:
> Oooh that's a spicy one I think. My knee-jerk reaction is a strong "heck no". However, I'm happy to have the discussion and I can definitely be convinced. @paulf81 what's the context and motivation for this?

Thanks for be open to discussion!  So here's my pitch points:

1) Saves us from having to ourselves (or our users) identify when our dependencies are obsolete: (eg #939 or more recently #1051)
2) Automatically generates the PR with the fix which we can review, saving some time for us to write these ourselves
3) Doesn't actually change any code if we don't accept the PR.  We can just note in the PR we don't want to make the change now and close the request

misi9170:
@rafmudaf that was my original reaction too! But I think @paulf81 has a point---it will only open a PR (which we can always reject if needed). If we find that this is happening too often or is overbearing, we can always remove it again.

misi9170:
By my reading, it looks like PRs will be opened targeting the main branch, rather than the develop branch. Perhaps this is correct---if there is a true dependency conflict, it may need to be resolved in a patch release. Otherwise, it looks like we can add `target-branch: "develop"` to the dependabot.yml to have it open PRs to the develop branch.

paulf81:
> By my reading, it looks like PRs will be opened targeting the main branch, rather than the develop branch. Perhaps this is correct---if there is a true dependency conflict, it may need to be resolved in a patch release. Otherwise, it looks like we can add `target-branch: "develop"` to the dependabot.yml to have it open PRs to the develop branch.

Good catch, added

paulf81:
>  If we find that this is happening too often or is overbearing, we can always remove it again.

Or move it from weekly to monthly

paulf81:
hi @rafmudaf, these are good points, thanks for them!  I guess my main thinking is these concerns mainly apply to the point we decided to merge a PR.  Implementing this change will only create the PR, and we could have a policy of "good to know" and simply close 90% of them according to a conservative approach of only updating failing packages.

I think there's a good motivation to be made aware at least of when packages we import have moved a version.  It's usually a heads up that conflicts will be coming sooner or later as eventually more and more of the ecosystem makes the jump.  Take for example the numpy -> 2 PR.  We could fairly claim that there was "no problem" in the sense the FLORIS itself works fine with numpy 1, but as one user commented our back dating was making complicated problems for them:

> I think there will be more difficulties for the majority of users if you do not upgrade to numpy 2.x than if you upgrade to numpy 2.x. There are less and less packages not compatible with numpy 2.0, Floris is one of them...
I think it's better to encourage users to upgrade to numpy 2 (for their own good) if they want to use new floris functionalities than to keep them in their "dead end" with numpy 1.20...

So my main argument is that since we aren't forced to merge the PRs, at least gaining the information they provide by their creation is useful and keeps us from having to learn these manually or through user-submitted issues.

Then going through your other points:

- Establishing a process for evaluating a dependency version update and the associated pull request

This will probably be case by case.  Numpy/pandas being probably merit a big discussion/think, where coloredlogs can be simpler.

- Reducing the frequency of checks to "monthly"; "quarterly" would be better but it's not an option

Dang

- Configure the [update-types option](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#update-types-ignore) to skip major versions

This won't work, FLORIS is configured to automatically accept all minor and patch versions within a major version.  The only thing that dependabot can alert us to our major versions.

- Configure the [versioning-strategy option](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#versioning-strategy--)

No strong feeling on this if you'd like to propose something?

misi9170:
Hi @paulf81 , @rafmudaf ,

I'm going through older pull requests and came across this one. I'm willing to give this a try. However, I think we should make the interval `monthly`, and I'm not sure about the versioning-strategy:
- `increase` seems wrong, I don't think we want that
- `increase-if-necessary` seems ok, perhaps the best
- `widen` doesn't seem like it would do much for us
- `auto` is the default, but the documentation doesn't seem to specify what the default behavior actually is?

More info [here](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/controlling-dependencies-updated#defining-a-versioning-strategy). It's not really clear what the distinction between "apps" and "libraries" is though.

paulf81:
HI @misi9170, I think since it just creates PRs we can ignore, it's low risk enough to try.  My understanding is that things on pip are usually libraries, since others might install them into their apps, and so FLORIS is a kind of library and should use the "widen" approach to be more permissive rather then less.

Another question, since we use the ~= style we're already pretty widely permissive so this might not do too much.  We could have to do something more like the minimum by having it only sweep for major versions:

```yaml

version: 2
updates:
  - package-ecosystem: "pip"
    directory: "/"
    schedule:
      interval: "monthly"
    ignore:
      - dependency-name: "*"
        update-types: ["version-update:semver-minor", "version-update:semver-patch"]
```

Since we already allow any new minor version

misi9170:
Ok, sounds good. So, let's leave the versioning-strategy as default; change the interval to monthly, and add those `ignore` cases. I can make those changes now.

misi9170:
[Here](https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide) is the quick-start guide. I will now merge this PR and Enable version updates (so that Dependabot can open PRs).

misi9170:
Now merged; however, to enable PRs to open, we may need to wait until this reaches the main branch (not totally clear).

paulf81:
I think that works, this wasn't urgent and when they arrive they arrive.  Will be nice to get a heads up on new major versions, thanks for merging @misi9170 !

Issue #1033, opened 2024-11-20
[BUG] Missed dimension reduction in `YawOptimization` base class


misi9170:
[These lines](https://github.com/NREL/floris/blob/440549ce6a603eb726971ccec645d3f7e86e249f/floris/optimization/yaw_optimization/yaw_optimization_base.py#L251-L264) 
```python
        idx = (minimum_yaw_angle_subset > 0.0) | (maximum_yaw_angle_subset < 0.0)
        if np.any(idx):
            # Find bounds closest to 0.0 deg
            combined_bounds = np.concatenate(
                (
                    np.expand_dims(minimum_yaw_angle_subset, axis=3),
                    np.expand_dims(maximum_yaw_angle_subset, axis=3)
                ),
                axis=3
            )
            # Overwrite all values that are not allowed to be 0.0 with bound value closest to zero
            ids_closest = np.expand_dims(np.argmin(np.abs(combined_bounds), axis=3), axis=3)
            yaw_mb = np.squeeze(np.take_along_axis(combined_bounds, ids_closest, axis=3))
            yaw_angles_template_subset[idx] = yaw_mb[idx]
```
were missed in the v4 dimension reduction process #764 , and can be fixed by replacing with
```python
        idx = (minimum_yaw_angle_subset > 0.0) | (maximum_yaw_angle_subset < 0.0)
        if np.any(idx):
            # Find bounds closest to 0.0 deg
            combined_bounds = np.concatenate(
                (
                    np.expand_dims(minimum_yaw_angle_subset, axis=2),
                    np.expand_dims(maximum_yaw_angle_subset, axis=2)
                ),
                axis=2
            )
            # Overwrite all values that are not allowed to be 0.0 with bound value closest to zero
            ids_closest = np.expand_dims(np.argmin(np.abs(combined_bounds), axis=2), axis=2)
            yaw_mb = np.squeeze(np.take_along_axis(combined_bounds, ids_closest, axis=2), axis=2)
            yaw_angles_template_subset[idx] = yaw_mb[idx]
```

(that is, switching from `axis=3` to `axis=2` throughout, and also specifying the axis for the `np.squeeze` operation).

The reason this hasn't come up before is that this code only runs if `minimum_yaw_angle` is strictly greater 0 or `maximum_yaw_angle` is strictly less than 0, which is an unusual circumstance.

I will create a bugfix PR in due course that will include tests for these "unusual" minimum and maximum angles.

misi9170:
Closed by #1067

PR #1067, opened 2025-02-19
[BUGFIX] Dimension update in `LayoutOptimizationBase`

misi9170:
As described in #1033 , there was a section of code that was missed in the dimension reduction from 5D to 4D as part of the release of Floris v4. The missed code is very much an edge case, where the _minimum_ yaw angle is set to be a stricly positive number (and vice versa, if the maximum yaw angle is set to be strictly negative).

This PR updates that section of code, and also adds an integration test that tests for these unusual situations.

- Existing tests pass.
- New tests added (yaw_optimization_integration_test.py) fail without changes to `YawOptimizationBase` (with an "axis out of bounds" error due to indexing a non-existent dimension 3---see screenshot below).
- New tests pass with changes to `YawOptimizationBase`.

## Notes
1. Since this is such an edge bug, I don't think we need to target main and release a patch. This can simply be bundled into the next minor release of FLORIS.
2. I'm also using this PR as an opportunity to fix a typo in the axis labels of one of the yaw optimization examples.

Screenshot of failing integration test prior to fixing dimensions:
image

### DIFF
--- a/examples/examples_control_optimization/002_opt_yaw_single_ws_uncertain.py
+++ b/examples/examples_control_optimization/002_opt_yaw_single_ws_uncertain.py
@@ -76,7 +76,7 @@ for tindex in range(3):
         color="r",
         marker="x",
     )
-    ax.set_ylabel("Yaw Offset (deg")
+    ax.set_ylabel("Yaw Offset T{0:03d} (deg)".format(tindex))
     ax.legend()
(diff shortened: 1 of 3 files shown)


paulf81 (approved):
Looks good, reviewed the change and agree the new test covers the condition.  All pass on my end.

PR #1068, opened 2025-02-21
Widen numpy requirement

paulf81:
# Widen numpy requirement

Since raising numpy requirement to 2.0 in the develop branch of FLORIS in #1051 , have noted local FLORIS/FLASC workflows are difficult with main FLORIS supporting 1.0 and develop supporting 2.0.  Each virtual environment needs a little manual editing to come to a workable state when working with develop branches.

This made me think a gentler transition allows both 1 and 2 for one version iteration (or at least on develop) and that's what this PR does.

Notes:

1. The branch does not put in an open-ended >, such that the problem identified in https://github.com/NREL/floris/issues/1030#issuecomment-2538806480 is repeated
2. I confirm tests pass with numpy 1.20 reinstalled since #1051 doesn't change the code 
3. The softer requirement for one version might be gentler for users, allowing the upgrade to numpy 2 while not requiring it immediately

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ dependencies = [
     "attrs",
     "pyyaml~=6.0",
     "numexpr~=2.0",
-    "numpy~=2.0",
+    "numpy >=1.20, <3",
     "scipy~=1.1",
     "matplotlib~=3.0",
     "pandas~=2.0",


paulf81:
I should note also acceptable to me would to merge this into develop, and then double back and force the requirement back to 2 right before merging to main, it's just the develop/main split that's making my workflows difficult.

misi9170:
I guess a problem with this is that, if we are mostly using FLORIS in other environments with numpy < 2, we somewhat miss the point of having the numpy~=2.0 on develop (because we would mostly use numpy 1 in the other environments, and then miss any bugs/slowdowns that numpy 2 might produce).

That being said, I think we've used numpy 2 enough now on the develop branch that I'm not too concerned about anything with it, and I'm ok with having the develop branch have the `"numpy >=1.20, <3",` form (although, is this equivalent to "numpy >=1.20, <=2"? that seems like a cleaner notation to me)

paulf81:
I think <=2 differs from <3 in that the latter allows 2.1 but not the former.  

Your top point though does make a good case for ripping the band-aid and just pushing everything up to 2, without my proposal of a midway release that allows 1.  I guess I could either way.  Maybe we could hash it out today in quick meeting, I think neither option clearly dominates the other so probably just a choice to make

paulf81:
Nevermind, stick to sem var and 4.3 will be numpy 2 dependent

PR #1079, opened 2025-03-12
Drop support for python 3.9 and add python 3.14

misi9170:
Following #1078 , and based on comments https://github.com/NREL/floris/pull/953#issuecomment-2714568167 and https://github.com/NREL/floris/issues/951#issuecomment-2714573015, this PR drops support for python version 3.9. This should remain as an open pull request until we feel that it is appropriate to drop support as python 3.9 nears [end-of-life](https://devguide.python.org/versions/).

This PR also:
- [x] Adds testing for mac-os and windows, thus superseding #953 
- [x] ~Adds example checks for mac-os and windows as well as ubuntu (currently failing!)~ Runs examples check on ubuntu only, using the oldest supported python version (3.10) and newest support python version (3.14)
- [x] Adds testing for python 3.14
- [x] Removes the `__future__ import annotations` needed to make type hinting compatible with python 3.9. This is not possible everywhere: see comments below.
- ~Begins using the [`match`/`case` structure](https://docs.python.org/3/tutorial/controlflow.html#match-statements)~ EDIT: After discussion, decided against a concerted change to switch case since it is not super common in python yet.

Open questions:
- Should we consider limiting the check-working-examples workflow to a single python version (e.g. 3.13) across all operating systems to limit the total number of actions? YES---chose 3.10, under the notion that all later versions 3.x should be back-compatible. EDIT: Decided to add a second version to the matrix, 3.14, to test both ends of spectrum of currently supported python versions.
- Can we change the `python-version: "3.13"` to `python-version: "3.x"` in the deploy-book workflow to allow this to float up with the latest python releases? EDIT: possibly `python-version: "3"` would work? Trying this.

### DIFF
--- a/floris/type_dec.py
+++ b/floris/type_dec.py
@@ -160,22 +160,22 @@ def convert_to_path(fn: str | Path) -> Path:
     # located (e.g., testing via pytest), and [-1] is where a direct call to the function via an
     # analysis script will be located (e.g., running an example).
     base_fn_script = Path(inspect.stack()[-1].filename).resolve().parent
-    # base_fn_sys = Path(inspect.stack()[1].filename).resolve().parent
+    base_fn_sys = Path(inspect.stack()[1].filename).resolve().parent
 
     if isinstance(fn, Path):
         absolute_fn = fn.resolve()
         relative_fn_script = (base_fn_script / fn).resolve()
-        # relative_fn_sys = (base_fn_sys / fn).resolve()
+        relative_fn_sys = (base_fn_sys / fn).resolve()
         if absolute_fn.exists():
             return absolute_fn
         if relative_fn_script.exists():
             return relative_fn_script
-        # if relative_fn_sys.exists():
-        #     return relative_fn_sys
+        if relative_fn_sys.exists():
+            return relative_fn_sys
         raise FileExistsError(
             f"{fn} could not be found as either a\n"
             f"  - relative file path from a script: {relative_fn_script}\n"
-            # f"  - relative file path from a system location: {relative_fn_sys}\n"
+            f"  - relative file path from a system location: {relative_fn_sys}\n"
             f"  - or absolute file path: {absolute_fn}"
         )
     raise TypeError(f"The passed input: {fn} could not be converted to a pathlib.Path object")


misi9170:
At this stage, I have updated the check-working-examples workflow to include mac and windows. However, this is causing a bit of a headache, because there are a number of bash commands that are OS-specific and are currently not working. We could consider running tests on mac, windows, and ubuntu, but only running examples on ubuntu. This was the approach taken in #953.

rafmudaf:
I love the evolution of the commit messages. YOLO force-pushes with coming soon :)

misi9170:
> I love the evolution of the commit messages. YOLO force-pushes coming soon :)

Haha nothing like pushing changes in a bash language I don't use to see what works!

misi9170:
After some back and forth and trying different things, we have decided to simplify the workflows while providing some coverage across OSs:
- examples run only with `ubunutu-latest`, and only with python `3.10` (the thinking being that if the examples run on the minimum supported python version, they should also run on more recent python versions)
- tests run on all supported OSs (`ubuntu-latest`, `macos-latest`, `windows-latest`) and all support python versions (`3.10`--`3.13`)

misi9170:
I've now also removed _most_ `from __future__ import annotations` imports, since python 3.10 enabled the type hinting we use throughout FLORIS. However, there is a catch: Any class methods that instantiate the class and have an output type hint like:
```
def instantiation_method(...) -> ClassName
```
still require this `annotations` import. See [here](https://stackoverflow.com/a/53455562) for an explanation.

This currently affects two files, where I've left the `from __future__ import annotations` line in:
- floris/core/core.py [here](https://github.com/NREL/floris/blob/0a250744b2bbf5d3163949c193f53efdedde1085/floris/core/core.py#L354)
- floris/wind_data.py [here](https://github.com/NREL/floris/blob/0a250744b2bbf5d3163949c193f53efdedde1085/floris/core/core.py#L354) and [here](https://github.com/NREL/floris/blob/0a250744b2bbf5d3163949c193f53efdedde1085/floris/wind_data.py#L1986)

Alternatively, the explicit return type hints could be removed in these locations, which would free us up to remove the `annotations` import. However, that doesn't seem like a great solution. Unfortunately, we may just need to add the `from __future__ import annotations` back to any files containing classes with some "self-instantiation" method.

paulf81:
> Alternatively, the explicit return type hints could be removed in these locations, which would free us up to remove the `annotations` import. However, that doesn't seem like a great solution. Unfortunately, we may just need to add the `from __future__ import annotations` back to any files containing classes with some "self-instantiation" method.

I think this is ok if it is what it is, but nice to have a bunch of them removed, thanks @misi9170 !

misi9170:
Python 3.14 has been released and python 3.9 is now officially at end-of-life. Removing "on-hold" flag and proceeding with the PR.

https://devguide.python.org/versions/

pyproject.toml
```diff
@@ -7,7 +7,7 @@ name = "floris"
 version = "4.4.1"
 description = "A controls-oriented engineering wake model."
 readme = "README.md"
-requires-python = ">=3.9"
+requires-python = ">=3.10"
```
rafmudaf: I suggest to make this `requires-python = ">=3.10,<=3.13"` or triple check that 3.14 works. The latest Python version is a huge change despite it carrying a minor version number change.

misi9170: good call, thanks! I've just added 3.14 to the testing matrix, so we'll see if those run through. I'll also run the examples with 3.14 to see if there are any issues.

misi9170:
Thanks for the suggestion @rafmudaf . I've now tested running the examples in ubuntu with 3.10 and 3.14, and both work (I'm thinking I should leave both in the examples testing matrix, to "bookend" the python versions). To be explicit, I'll also add the upper version to the pyproject.toml at `requires-python = ">=3.10, <=3.14".`

I've also run tests locally with a conda environment with a 3.14 installation and built the docs locally with the same conda environment.

rafmudaf:
Thanks for checking that @misi9170 and I'm happy to hear it all works with the new version of Python!

floris/type_dec.py
misi9170, line 1: @rafmudaf , you spent some time looking into this (see [this comment](https://github.com/NREL/floris/issues/951#issuecomment-2484057737))---just wondering if you know whether we can go ahead and remove the lines that are commented out here?

EDIT: tracking this back further, these changes were proposed in #953 , which this PR supersedes

rafmudaf: I *think* you can add those commented line back in. You could try it and see if the windows tests still fail. The issue was specifically related to a pathlib implementation that was improved in Python 3.10+, so by dropping support for 3.9 in FLORIS I think that's resolved.

misi9170: Trying now

misi9170: Ok, looks good, the tests with windows and pythons version 3.10 to 3.14 are all passing.

misi9170:
One small unknown is whether the docs will build correctly with `python-version: "3"` in deploy-pages.yaml. I think we can go ahead and merge the PR into develop and find out---if not, I can push a change to develop to make this `python-version: "3.14"` (which worked fine for building the docs on my laptop).

rafmudaf:
Yea I think the docs build will work, but it'll be easy to fix if it doesn't

misi9170:
Post merger (while I'm still waiting for the docs to build), it seems that adding mac and windows OS testing has caused a different issue to do with codecov: https://github.com/NREL/floris/actions/runs/18693264948. I think we can fix this by adding `if: matrix.os == 'ubuntu-latest'` to the Codecov upload task.

misi9170:
Unfortunately a couple of direct pushes to try things on develop didn't work; I'll have to dig into this further.

misi9170:
Ok, after a couple of bad attempts (and annoying commits to develop), the issue is now fixed in b1ae666.

Issue #1080, opened 2025-03-13
Bug in AWC operation model


gyalla:
# Bug in AWC operation model

There appears to be a bug in the 'awc' operational model in floris concerning the computation of power and thrust of actuated turbines (turbines set to 'helix' as opposed to 'baseline').

The error stems from the use of any() here: https://github.com/NREL/floris/blob/c50389cfba596e29dcbdfa7670db591b2b65da3e/floris/core/turbine/operation_models.py#L540 
Instead of only reducing the power of 'helix' turbines, if any turbine is labelled as 'helix' in awc_modes, the power of all turbines will be reduced. 

## How to reproduce
Consider the following minimal example: 
```
import numpy as np 
from floris import (
    FlorisModel,
    WindRose,
)
cases = ["Baseline","AWC",]
fmodels = {}
fmodels['Baseline'] = FlorisModel("../floris/examples/inputs/emgauss_helix.yaml")
fmodels['AWC'] = FlorisModel("../floris/examples/inputs/emgauss_helix.yaml")
awc_array = ['helix','baseline','baseline']
baseline_array = ['baseline','baseline','baseline']
awc_amplitudes = [4,4,4]

fmodels['Baseline'].set_operation_model("awc")
fmodels['Baseline'].set(
    awc_modes=np.array([baseline_array]),
    awc_amplitudes=np.array([awc_amplitudes])
    )

fmodels['AWC'].set_operation_model("awc")
fmodels['AWC'].set(
    awc_modes=np.array([awc_array]),
    awc_amplitudes=np.array([awc_amplitudes])
    )

turbine_powers = {}
for case in cases:
    floris_model = fmodels[case]
    floris_model.run_no_wake()
    turbine_powers[case] = floris_model.get_turbine_powers()/1000
    print(case,": ",turbine_powers[case])
```
which results it: 
`Baseline :  [[6301.1399872 6301.1399872 6301.1399872]]`
`AWC :  [[5908.03837044 5908.03837044 5908.03837044]]`
wheras the correct answer should be: 
`Baseline :  [[6301.1399872 6301.1399872 6301.1399872]]`
`AWC :  [[5908.03837044 6301.1399872  6301.1399872 ]]`

Additionally, this will cause an error when performing a wind sweep or estimating AEP. If 'helix' occurs for any turbine at any wind condition, then all turbines at all wind conditions see a reduction in power. Consider the following minimal example with two wind conditions:  
```
for caseiter, case in enumerate(cases):
[...]

PR #1081, opened 2025-03-14
[BUGFIX] AWCTurbine operation model with mixed `awc_modes` inputs

gyalla:
# AWC Operation Mode Bug Fix

This PR addresses a bug in the 'awc' operation model identified in issue #1080.

### DIFF
--- a/floris/core/turbine/operation_models.py
+++ b/floris/core/turbine/operation_models.py
@@ -550,7 +550,7 @@ class AWCTurbine(BaseOperationModel):
                 + power_thrust_table['helix_power_c']*base_powers
                 )
                 *awc_amplitudes**power_thrust_table['helix_a']
-            ) 
+            )
 
         mask = (awc_modes == 'helix')
         if np.any(np.isclose(base_powers[mask]/1000,np.max(power_thrust_table['power']))):
@@ -587,7 +587,11 @@ class AWCTurbine(BaseOperationModel):
                 )
 
         mask = (awc_modes == 'helix')
-        return np.where(mask, apply_awc(base_thrust_coefficients,power_thrust_table,awc_amplitudes), base_thrust_coefficients)
+        return np.where(
+            mask,
+            apply_awc(base_thrust_coefficients,power_thrust_table,awc_amplitudes),
+            base_thrust_coefficients
+        )
 
     def axial_induction(
         power_thrust_table: dict,
--- a/floris/core/turbine/operation_models.py
+++ b/floris/core/turbine/operation_models.py
@@ -537,29 +537,29 @@ class AWCTurbine(BaseOperationModel):
             cubature_weights=cubature_weights
         )
 
-        if (awc_modes == 'helix').any():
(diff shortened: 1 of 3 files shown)


misi9170:
@gyalla Thank you again for raising this bug and providing the fix. I made a couple of changes that I hope you and @jfrederik-nrel will agree with, using the code you provided:
- Created a standalone method `AWC_model()` on the `AWCTurbine` operation model, which can be used in the calculation of both the `thrust_coefficient` and `power`.
- Switched from using `np.where()` to explicit masking for two reasons: 1) to allow easy expansion to the base where there may be other possible values for `awc_modes` (no top of `"baseline"` and `"helix"`) and 2) so that only the findex/turine combinations where AWC is active are actually evaluated (`np.where()` evaluated both of its arguments in full before applying masking), which should mean that the `AWCTurbine` evaluations are faster in cases where there are many `"baseline"` entries and few `"helix"` entries.

I also added a couple of tests where there are multiple turbines and multiple findices, and the `awc_modes` are not uniform across all findices/turbines. These tests were informed by the very helpful script you provided in #1080. As expected, these tests _fail_ on the develop branch (i.e., prior to the bugfix) but pass with the updated code.

As long as you are both happy with the changes I've made here, we can get this merged in and it will be part of the next FLORIS release.

misi9170 (approved):

floris/core/turbine/operation_models.py
```diff
@@ -544,22 +547,27 @@ def power(
...
+        helix_mask = (awc_modes == 'helix')
+        if np.any(np.isclose(base_powers[helix_mask]/1000,np.max(power_thrust_table['power']))):
             raise UserWarning(
                 'The selected wind speed is above or near rated wind speed. '
                 '`AWCTurbine` operation model is not designed '
                 'or verified for above-rated conditions.'
             )
-        return np.where(mask, apply_awc(base_powers,power_thrust_table,awc_amplitudes), base_powers)
+
+        awc_powers = AWCTurbine.AWC_model(
```
jfrederik-nrel, line 561: You use the wrong values from the `power_thrust_table` here. It should be:
`power_thrust_table['helix_thrust_b'],`
`power_thrust_table['helix_thrust_c'],`
in line 563-564

jfrederik-nrel (changes requested):
@misi9170 one tiny comment in `operation_models.py`.

floris/core/turbine/operation_models.py
misi9170: Oof, good catch, thanks @jfrederik-nrel ! You mean below (in `thrust_coefficient`) I've passed the wrong values, right?

jfrederik-nrel: Oh sorry you're right, I put the comment in the wrong place. Line 596-597

misi9170: Ok, fixed now, thanks for catching that!

gyalla:
These changes all look good to me, thanks @misi9170!

Issue #1089, opened 2025-03-24
Bug in AWC Mixing Factor


gyalla:
# Bug in AWC Mixing Factor

There is a bug in the AWC wake mixing model, similar to issue #1080, stemming from the use of any() here: 

https://github.com/NREL/floris/blob/b9c57a13b34035f2a741f3aa568a881e87016e7e/floris/core/wake_velocity/empirical_gauss.py#L297

Whereas issue #1080 was related to the power and thrust of turbines labeled as 'helix', this issue affects their velocity deficit and the subsequent power of downstream turbines. If a turbine is labeled as 'helix' at any one wind condition, the mixing factor associated with that turbine is increased at all wind conditions. 

## How to reproduce
Consider the following minimal example

```
fmodel = FlorisModel('./emgauss_helix.yaml')
awc_modes = np.array([['helix','baseline','baseline'],['baseline','baseline','baseline']])
awc_amplitudes = np.array([[4.0,4.0,4.0]]*2)
fmodel.set_operation_model('awc')
fmodel.core.wake.wake_velocity_parameters['empirical_gauss']['mixing_gain_velocity']=2.0
fmodel.set(
wind_directions=np.array([270.0, 270.0]),
wind_speeds=[8.0, 8.0],
turbulence_intensities=np.array([0.06, 0.06]),
awc_modes=awc_modes,
awc_amplitudes=awc_amplitudes
)
fmodel.run()
turbine_powers = fmodel.get_turbine_powers() / 1000.0
print(turbine_powers)
```
which outputs: 
```
[[5908.03837044 2616.78227155 2931.02406667]
 [6301.1399872  2547.88521286 2930.8745969 ]]
```
Compare this to the output of the following code, which is nearly identical but labels all turbines as 'baseline':
```
fmodel = FlorisModel('./emgauss_helix.yaml')
awc_modes = np.array([['baseline','baseline','baseline'],['baseline','baseline','baseline']])
awc_amplitudes = np.array([[4.0,4.0,4.0]]*2)
fmodel.set_operation_model('awc')
fmodel.core.wake.wake_velocity_parameters['empirical_gauss']['mixing_gain_velocity']=2.0
fmodel.set(
wind_directions=np.array([270.0, 270.0]),
wind_speeds=[8.0, 8.0],
turbulence_intensities=np.array([0.06, 0.06]),
awc_modes=awc_modes,
awc_amplitudes=awc_amplitudes
)
fmodel.run()
[...]

PR #1090, opened 2025-03-25
 [BUGFIX] AWC wake mixing factor with mixed awc_modes inputs #1089

gyalla:
This PR addresses a bug in the mixing factor when using AWC that was identified in issue https://github.com/NREL/floris/issues/1089.

### DIFF
--- a/floris/core/wake_velocity/empirical_gauss.py
+++ b/floris/core/wake_velocity/empirical_gauss.py
@@ -294,12 +294,7 @@ def awc_added_wake_mixing(
 ):
 
     # TODO: Add TI in the mix, finetune amplitude/freq effect
-    if (awc_mode_i == "helix").any():
-        return awc_amplitude_i[:,:,0,0]**awc_wake_exp/awc_wake_denominator
-    elif (awc_mode_i == "baseline").any():
-        return 0
-    else:
-        raise NotImplementedError(
-            'Active wake mixing strategies other than the `helix` mode '
-            'have not yet been implemented in FLORIS.'
-        )
+    awc_mixing_factor = np.zeros_like(awc_amplitude_i[:,:,0,0])
+    helix_mask = (awc_mode_i[:,:,0,0] == 'helix')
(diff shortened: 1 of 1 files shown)


misi9170:
@gyalla , totally fair, yes I'll take a look now

misi9170:
@gyalla , your proposal looks good to me. I've reformatted it a little bit, but all works fine. I was going to add another test to ensure that checks this functionality, but I realized we don't yet have a place where we test how the operation models affect flows, so to speak---I'll open an issue for that, but likely that'll happen at a later date.

Thanks again for providing an example runscript in #1089---that's really helpful.

misi9170 (approved):

gyalla:
Thanks for taking a look @misi9170, and for reformatting the code. All looks good to me, so feel free to merge whenever makes sense.

misi9170:
Making a note here that this change means that if any entries in `awc_mode_i` are something other than `"helix"`, zeros will be returned. This works as intended (for instance, if the entry is `"baseline"`), but will also return zeros if the `awc_mode_i` contains anything other than `"helix"` or `"baseline"`. Previously, passing anything else would have raised a `NotImplementedError`. Still, I think that's fine, since there is already a [warning raised in the `AWCTurbine` model](https://github.com/NREL/floris/blob/b9c57a13b34035f2a741f3aa568a881e87016e7e/floris/core/turbine/operation_models.py#L545) if this occurs. Moreover, I'd like to incorporate this kind of "checking" into a future class that is dedicated to handling control setpoints #1032.

gyalla:
@misi9170, we need to specify the mixing factor as 

`awc_mixing_factor = np.zeros_like(awc_amplitude_i,dtype=float)`

Otherwise, if `awc_amplitudes` are specified with integers, which they commonly are, then `awc_mixing_factor` will be rounded down and likely zeroed out. Thanks to @moprak-nrel for helping me track this one down!

misi9170:
Good catch @gyalla ! Actually, I was just talking to @rafmudaf about exactly this type of thing, and what we should really be using is the [`floris_float_type`](https://github.com/NREL/floris/blob/main/floris/type_dec.py#L23)---I'm going to quickly change it to that, but you're right about specifying the type!

Issue #1094, opened 2025-04-03
[BUG] demo_cp_ct_surfaces not installed correctly from PyPI.


misi9170:
It appears that Cp/Ct surface data that was introduced in #832 is not correctly installed from PyPI. For example, [automated tests in FLASC are failing](https://github.com/NREL/flasc/actions/runs/14251162947/job/39944008877?pr=238) because this data is missing.

Possibly, we need to add the turbine_library/demo_cp_ct_surfaces/ data to the `[tool.setuptools.package-data]` section of the pyproject.toml?

## Steps to reproduce
Create a new conda environment, pip install FLORIS, and try to instantiate a `FlorisModel`

```
conda create --name temp python
conda activate temp
pip install floris
```
Then, in a python terminal with the `temp` environment activated
```python
from floris import FlorisModel
fmodel = FlorisModel("defaults")
```
This causes an error:
```
FileNotFoundError: [Errno 2] No such file or directory: '/opt/miniconda3/envs/temp/lib/python3.13/site-packages/floris/turbine_library/demo_cp_ct_surfaces/nrel_5MW_demo_cp_ct_surface.npz'
```

## Notes
I think this should not affect users that clone directly from github; just those installing Floris from PyPI. Current workaround is to specify `pip install floris==4.3` to stay on the previous Floris version before the bug was introduced.

misi9170:
Should be addressed by #1095. I'll know for sure once we release 4.4.1 and will close this PR at that time.

misi9170:
[Release 4.4.1](https://github.com/NREL/floris/releases/tag/v4.4.1) fixes this issue with the inclusion of #1095.

PR #1095, opened 2025-04-03
[BUGFIX] Add demo_cp_ct_surface location to data files.

misi9170:
I believe this should address #1094 . I'm not sure what the best way to test the fix is, though, short of merging this and creating a new release (UPDATE: Read comments below for how to test locally).

UPDATE:
To avoid this type of bug being introduced again, as well as adding the missing data to the `[tool.setuptools.package-data]` list of pyproject.toml, this PR:
- Makes github workflow installations of floris non-editable by removing `-e` flag
- Moves tests into a subdirectory to avoid finding the data (still don't totally understand why this is necessary, but it has the desired effect).


rafmudaf:
You can test this locally by doing `pip install floris/` without the typical `-e` in a new Python environment. Then, you can search for the file in the installation directory to verify it's there and also run the demo case.

misi9170:
@rafmudaf 's testing suggestion did work locally. I'm now testing removing the `-e` flag from github workflows to see if that would have identified this issue earlier (thanks, @rafmudaf , for the suggestion). Assuming the tests fail in the expected way, I'll re-add the package data. If test/examples then pass, this should be good to go.

misi9170:
Expected failures are now happening. Now re-adding the data.

misi9170:
Ok, after a lot of digging, @rafmudaf and I found and a couple of other issues:
- added wrg_test.wrg to tests/data to avoid relying on data in the examples directory
- added `-rL` flag to copies to resolve symbolic links
- add turbine_library/*.csv to `[tool.setuptools.package-data]` so that multimensional turbine data is available with Floris is installed using PyPI
- created more robust paths using the package root for data loading in tests

rafmudaf (approved):

Issue #1106, opened 2025-05-06
Bug report?: optimize yaw with wind direction uncertainty does not work in with ParFlorisModel


gogannes:
# optimize yaw with wind direction uncertainty does not work in with ParFlorisModel

It seems that it is not possible to run a parallel yaw optimization with wind direction uncertainty.

This is unexpected to me, as the UncertainFlorisModel allows for parallelization (`floris/examples/examples_uncertain/003_uncertain_model_with_parallelization.py`). 

I put a question mark into the title, as I'm not sure it is indeed a bug or just a missing feature (in that case I think a warning should be displayed for the given example below).

## How to reproduce

```
from time import perf_counter as timerpc

import numpy as np

from floris import (
    ParFlorisModel,
    TimeSeries,
    WindRose,
)
from floris.optimization.yaw_optimization.yaw_optimizer_sr import YawOptimizationSR


# When using parallel optimization it is important the "root" script include this
# if __name__ == "__main__": block to avoid problems
if __name__ == "__main__":

    # Load the wind rose from csv
    wind_rose = WindRose.read_csv_long(
        "../inputs/wind_rose.csv", wd_col="wd", ws_col="ws", freq_col="freq_val",
        ti_col_or_value=0.06
    )

    # Load FLORIS as a parallel model
    max_workers = 16
    pfmodel = ParFlorisModel(
        "../inputs/gch.yaml",
        max_workers=max_workers,
        n_wind_condition_splits=max_workers,
        interface="pathos",
        print_timings=True,
    )

    # transform the model into the uncertain model
    from floris import UncertainFlorisModel
    pfmodel = UncertainFlorisModel(pfmodel)


    # Specify wind farm layout and update in the floris object
    N = 2  # number of turbines per row and per column
    X, Y = np.meshgrid(
        5.0 * pfmodel.core.farm.rotor_diameters_sorted[0][0] * np.arange(0, N, 1),
        5.0 * pfmodel.core.farm.rotor_diameters_sorted[0][0] * np.arange(0, N, 1),
    )
    pfmodel.set(layout_x=X.flatten(), layout_y=Y.flatten())

    # Get the number of turbines
    n_turbines = len(pfmodel.layout_x)

    # Optimize the yaw angles.  This could be done for every wind direction and wind speed
    # but in practice it is much faster to optimize only for one speed and infer the rest
    # using a rule of thumb
    time_series = TimeSeries(
[...]

PR #1107, opened 2025-05-06
[BUGFIX] `UncertainFlorisModel` not correctly using `ParFlorisModel`

misi9170:
@gogannes pointed out a bug in #1106 that the `UncertainFlorisModel` did not appear to be using parallelization when passed an instantiated `ParFlorisModel` as an input during yaw optimization routines. Looking into this further, I believe that the `UncertainFlorisModel` did not correctly assign the `ParFlorisModel`, and instead (silently) converted it into a `FlorisModel` type object.

At the root of the issue is that the `ParFlorisModel`, which is a child of `FlorisModel`, did not implement it's own `copy()` method; thus, when the `UncertainFlorisModel` calls `ParFlorisModel.copy()`, a `FlorisModel` is returned rather than a `ParFlorisModel`.

This PR corrects this issue by defining `copy()` on `ParFlorisModel`. In doing so, I believe it addresses #1106. 
EDIT: I have now switched to a more refined approach where `copy()` uses a property dictionary `secondary_init_kwargs`. This obviates the need for a `copy()` method on `ParFlorisModel`, and I have removed it again in favor of the `secondary_init_kwargs` approach.

Check list:
- [x] Implement fix for bug
- [x] Update examples if needed (example 009_parallel_models.py was not running correctly, but did not need to be changed; while examples_uncertain/003_uncertain_model_with_parallelization.py was not running correctly and needed a change so that the parallel model runs correctly).
- [x] Add test to check that the `ParFlorisModel` is correctly assigned to the `UncertainFlorisModel` (this new test fails on the develop branch, and only passes with the bug fix included)
- [x] Check that the expected behavior is occurring that, for large problems, `UncertainFlorisModel` running in parallel (using `ParFlorisModel`) is significantly faster than `UncertainFlorisModel` running using `FlorisModel`.

The underlying bug here was the `copy()` method of the FlorisModel object and its variants. This method has some other issues, in particular, it also does not copy over wind data structures and control setpoints #997. We may need to consider revamping how `copy()` works in future.

### DIFF
--- a/floris/floris_model.py
+++ b/floris/floris_model.py
@@ -1576,7 +1576,12 @@ class FlorisModel(LoggingManager):
         )
 
     def copy(self):
-        """Create an independent copy of the current FlorisModel object"""
+        """Create an independent copy of the current FlorisModel object
+
+        When creating the copy, this method uses self.__class__(), rather than FlorisModel()
+        directly, so that subclasses of FlorisModel can inherit this method and return
+        instantiations of their own class, rather than the FlorisModel class.
+        """
         return self.__class__(self.core.as_dict(), **self.secondary_init_kwargs)
 
     def get_param(
@@ -1665,8 +1670,8 @@ class FlorisModel(LoggingManager):
     @property
     def secondary_init_kwargs(self):
         """
-        FlorisModel takes only the configuration argument. This method is a placeholder
-        for children of FlorisModel.
+        FlorisModel takes only the configuration argument. This method is a placeholder for
+        subclasses of FlorisModel.
         """
         return {}
 
--- a/floris/floris_model.py
+++ b/floris/floris_model.py
@@ -1577,7 +1577,7 @@ class FlorisModel(LoggingManager):
 
     def copy(self):
         """Create an independent copy of the current FlorisModel object"""
-        return FlorisModel(self.core.as_dict())
+        return self.__class__(self.core.as_dict(), **self.secondary_init_kwargs)
 
     def get_param(
         self,
@@ -1616,7 +1616,7 @@ class FlorisModel(LoggingManager):
         """
         fm_dict_mod = self.core.as_dict()
         nested_set(fm_dict_mod, param, value, param_idx)
-        self.__init__(fm_dict_mod)
+        self.__init__(fm_dict_mod, **self.secondary_init_kwargs)
 
     def get_turbine_layout(self, z=False):
         """
@@ -1661,6 +1661,14 @@ class FlorisModel(LoggingManager):
         self.show_config(full=True)
 
     ### Properties
+
+    @property
+    def secondary_init_kwargs(self):
+        """
+        FlorisModel takes only the configuration argument. This method is a placeholder
+        for children of FlorisModel.
+        """
+        return {}
 
     @property
     def layout_x(self):
--- a/examples/examples_uncertain/003_uncertain_model_with_parallelization.py
+++ b/examples/examples_uncertain/003_uncertain_model_with_parallelization.py
@@ -13,73 +13,73 @@ from floris import (
 )
 from floris.par_floris_model import ParFlorisModel
 
+if __name__ == "__main__":
+    # Following the refactoring of ParFlorisModel, the UncertainFlorisModel can be
+    # parallelized by passing the ParFlorisModel as the model to be run.  This example
+    # demonstrates the usage and shows that the result obtained from the UncertainFlorisModel
+    # with and without parallelization is the same.  The results are compared to the nominal
+    # results.
 
-# Following the refactoring of ParFlorisModel, the UncertainFlorisModel can be
-# parallelized by passing the ParFlorisModel as the model to be run.  This example
-# demonstrates the usage and shows that the result obtained from the UncertainFlorisModel
-# with and without parallelization is the same.  The results are compared to the nominal
-# results.
+    # Instantiate a FlorisModel and ParallelFlorisModel using the GCH model
+    fmodel = FlorisModel("../inputs/gch.yaml")
+    pfmodel = ParFlorisModel("../inputs/gch.yaml")
 
-# Instantiate a FlorisModel and ParallelFlorisModel using the GCH model
-fmodel = FlorisModel("../inputs/gch.yaml")
-pfmodel = ParFlorisModel("../inputs/gch.yaml")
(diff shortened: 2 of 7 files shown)


misi9170:
I added a couple more tests and noticed a bug could occur also in `UncertainFlorisModel.copy()`, which I've now fixed.

misi9170:
To round this out, I think I have confirmed that indeed, the `UncertainFlorisModel` was not using parallelization.

To see this, run the following script:
```python

import numpy as np

from floris import (
    FlorisModel,
    ParFlorisModel,
    UncertainFlorisModel,
    WindRose,
)
from floris.optimization.yaw_optimization.yaw_optimizer_sr import YawOptimizationSR

use_parallel = False

if __name__ == "__main__":

    # Load the wind rose from csv
    wind_rose = WindRose.read_csv_long(
        "../inputs/wind_rose.csv", wd_col="wd", ws_col="ws", freq_col="freq_val",
        ti_col_or_value=0.06
    )

    # Load FLORIS as a parallel model
    max_workers = 16

    if use_parallel:
        fmodel = ParFlorisModel(
            "../inputs/gch.yaml",
            max_workers=max_workers,
            n_wind_condition_splits=max_workers,
            interface="pathos",
            print_timings=True,
        )
    else:
        fmodel = FlorisModel("../inputs/gch.yaml")

    # Set wind rose and layout
    n_turbines = 100
    fmodel.set(
        wind_data=wind_rose,
        layout_x=np.linspace(0, 1000*n_turbines, n_turbines),
        layout_y=np.zeros(n_turbines),
    )

    # Set up uncertain model
    ufmodel = UncertainFlorisModel(fmodel)

    # Time a power calculation
    start = timerpc()
    ufmodel.run()
    aep = ufmodel.get_farm_AEP()
    total = timerpc() - start
    print(f"Number of turbines: {ufmodel.n_turbines}. Number of findices: {ufmodel.n_findex}")
    print(f"Power calculation took {total:.2f} seconds (Calculated AEP: {aep/1e9:.2f} GWh)")
```

On the develop branch, prior to the bugfix, with the line `use_parallel = False`, this results in
```
Number of turbines: 100. Number of findices: 1327
Power calculation took 91.75 seconds (Calculated AEP: 1974.92 GWh)
```
while with `use_parallel = True`,
```
Number of turbines: 100. Number of findices: 1327
Power calculation took 98.30 seconds (Calculated AEP: 1974.92 GWh)
```
(i.e., no obvious change).

On this bugfix branch, the results for `use_parallel = False` are
```
Number of turbines: 100. Number of findices: 1327
Power calculation took 99.62 seconds (Calculated AEP: 1974.92 GWh)
```
as before, while with `use_parallel = True`, I get
```
Number of turbines: 100. Number of findices: 1327
Power calculation took 39.38 seconds (Calculated AEP: 1974.92 GWh)
```
on my Mac laptop (i.e., a significant speed improvement due to parallelization, as well as outputs reflecting the use of the `ParFlorisModel` under the hood).

___

While creating the script above, I noticed another subtle bug: as explained in the main
PR description, `FlorisModel.copy()` does _not_ copy over an assigned `wind_data` object.
Because the above script assigns `wind_data` to the (certain) `FlorisModel`/`ParFlorisModel`,
prior to instantiation of the `UncertainFlorisModel`, the `wind_data` is not carried over,
and in particular, the frequency values of the bins are lost. Running the above script on 
the develop branch prints the warning:
```
floris.uncertain_floris_model.UncertainFlorisModel WARNING Computing AEP with uniform frequencies. Results results may not reflect annual operation.
```
to notify users, which is how I noticed the issue. I have now updated
`UncertainFlorisModel.__init__()` to additionally copy over any `wind_data` and control setpoints,
as was done in #997 for the `ParFlorisModel`. I've also added a test for this (fails on develop; passed on this branch).

misi9170:
Note also, that, after updates, the expected line
```
 "Total time spent for parallel calculation (.. workers): ..."
 ```
 is produced by the script provided by @gogannes in #1106 .
 
 Finally, I added a small piece of code to quiet an unnecessary warning produced by the `YawOptimization` base class when running with an `UncertainFlorisModel`. This is equivalent to the quieting implemented in #1045, but that only worked for "certain" `FlorisModel`s.

**Comment by misi9170** [comment]
Thanks for reviewing @paulf81 !

In light of @gogannes [recent finding with `set_param()`](https://github.com/NREL/floris/issues/1106#issuecomment-2857501602) (see also [my response](https://github.com/NREL/floris/issues/1106#issuecomment-2858894597), where I found that the issue is that the "secondary arguments" of `ParFlorisModel.__init__` are not properly carried over), I've pivoted my approach here a bit to something that is a bit more elegant, if possibly a bit less obvious (jury is out on the "pythonicness").

In the new approach, each of `FlorisModel` and its derivatives contain a new `@property` `self.secondary_init_kwargs`. This returns a dictionary of the optional `kwargs` that follow `configuration` for instantiation of any of these classes. For `ParFlorisModel` and `UncertainFlorisModel`, this is non-empty---for `FlorisModel`, it is an empty dict and simply a placeholder for child classes. 

The benefit here is that now any time there is some sort of "reinstantation" procedure, it is easy to also copy over these secondary keyword arguments. We can now do away with a dedicated `copy` method on `ParFlorisModel`, simply inheriting `FlorisModel`'s `copy` as before. The same now works for `set_param` (this was not working correctly before, which caused this second look).

I think this works pretty nicely---all tests pass without needing any updates. It may also be an approach that could later work for copying over `wind_data` objects and `control_setpoints` (if we have a `set_kwargs` `@property`, but I will not include that in this PR because I think that's enough of a separate issue).

File: floris/floris_model.py [hunk 1]
```diff
@@ -1577,7 +1577,7 @@ def set_operation_model(self, operation_model: str | List[str]):
 
     def copy(self):
         """Create an independent copy of the current FlorisModel object"""
-        return FlorisModel(self.core.as_dict())
+        return self.__class__(self.core.as_dict(), **self.secondary_init_kwargs)
```
rafmudaf, line 1580: I would note here in the docstring that `self.__class__` is used so that subclasses will return themselves. It's clear what you did when I come to this from the pull request, but I think it would be odd if I came to this without the added context. I love the idea of this fwiw.

rafmudaf (approved):
I like it - nicely done @misi9170

PR #1113, opened 2025-05-23
fix: update Python version requirement and add uv.lock for compatibility

yuanchima:
- Set requires-python to ">=3.10" in pyproject.toml to match bokeh>=3.7 requirements
- Add uv.lock for reproducible dependency management with uv

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,5 @@ tests/v3_to_v4_convert_test/convert_turbine_v3_to_v4.py
 tests/v3_to_v4_convert_test/convert_floris_input_v3_to_v4.py
 tests/v3_to_v4_convert_test/gch_v4.yaml
 tests/v3_to_v4_convert_test/nrel_5MW_v3_v4.yaml
+
+.venv
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ name = "floris"
 version = "4.4.2"
 description = "A controls-oriented engineering wake model."
 readme = "README.md"
-requires-python = ">=3.9"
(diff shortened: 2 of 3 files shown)


misi9170:
Thanks for the PR @yuanchima ! Please also see #1079, where we are formally planning to drop support for earlier python versions (although I had missed the update to pyproject.toml, so I've now fixed that). Good to know that the `bokeh` requirement is not compatible with python 3.9---since `bokeh` is only used for building the documentation, I feel that this incompatibility is not currently critical, but I'll also call out that #1064 might have caught this inconsistency.

Regarding package management with uv---this is something we've talked about internally but have not yet figured out if we'll explicitly support. I will check with a few people on the development team and let you know what we decide on that.

misi9170:
Hi @yuanchima ,

We have now bumped up the python version requirement to `>=3.10, <=3.14` in the [develop branch](https://github.com/NREL/floris/blob/develop/pyproject.toml#L10) and will appear on main in the next minor version release.

Regarding adding a uv.lock file to FLORIS: although the uv developers suggest committing the lock file (see [here](https://docs.astral.sh/uv/guides/projects/#uvlock)), this seems to be more aimed at "project"-type repositories that call various different libraries. We consider FLORIS to be more like a "library" that users will install and use within their own project repositories, and so we currently feel that the uv.lock file is more appropriately located "up a level" from FLORIS, in the user's project repository. As such, at this stage, we will not include the uv.lock file in FLORIS itself.

Thanks again for opening this PR, and I'm sorry that I'm turning it down---not necessarily because it's wrong, but just a bit redundant with #1079.

Best,
Misha

PR #1120, opened 2025-06-18
Set keepdims to True so that we don't have to add again those dimensions

RRiva:
# Set keepdims to True so that we don't have to add again those dimensions

Instead of doing
```python
b = np.mean(a, axis=(2, 3))
b = b[:, :, None, None]
```
do `b = np.mean(a, axis=(2, 3), keepdims=True)`.

## Related issue

None

Solver and a couple of Gauss modules.

### DIFF
--- a/floris/core/solver.py
+++ b/floris/core/solver.py
@@ -253,8 +253,9 @@ def sequential_solver(
     flow_field.turbulence_intensity_field_sorted = turbine_turbulence_intensity
     flow_field.turbulence_intensity_field_sorted_avg = np.mean(
         turbine_turbulence_intensity,
-        axis=(2,3)
-    )[:, :, None, None]
+        axis=(2,3),
+        keepdims=True
+    )
 
 
 def full_flow_sequential_solver(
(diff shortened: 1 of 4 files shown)


misi9170:
Thanks @RRiva ! I generally agree with the proposal here to use `keepdims = True` rather than taking the mean and then reexpanding the dimensions using `[:, :, None, None]`. I'll take a look through and fix the tests.

There are other locations in the code where there isn't necessarily a mean or any other dimension-reducing operation, but we still need to expand the dimensions so we're still using `[:, :, None, None]`. This isn't the most obvious notation to a new python user (although it's common enough in numpy that perhaps its not a big deal). However, if we want to increase clarity, there are a couple of other options that numpy provides for this:
- [`expand_dims`](https://numpy.org/doc/stable/reference/generated/numpy.expand_dims.html), used like `x = np.expand_dims(x, (2,3))`
- [`np.newaxis`](https://numpy.org/doc/2.2/reference/constants.html#numpy.newaxis), which is a direct replacement (and actually just an alias for) `None`, used like `x = x[:, :, np.newaxis, np.newaxis]`.

My feeling is that perhaps the `np.newaxis` option is nice; nothing really changes at all, but it's clearer what is happening in the operation. What do you think?

RRiva:
Hi @misi9170, thans a lot 🙂 In PyWake we use `np.newaxis` all the time, except that Mads abbreviated it as `na`. I don't think it's important though, since it's just syntactic sugar for `None`. This PR was simply about doing 1 thing rather than 2.

misi9170:
Ok, understood, thanks @RRiva . I've now pushed up a couple more instances where `keepdims` could have been used, and also made a change to wind_data.py that (although it's not strictly about `keepdims`) I think is in keeping with the intent of this PR: `x[:, :, 0][:, :, None]` -> `x[:, :, 0:1]`.

Provided you are happy with this, I think we can get this PR merged into the develop branch and the changes will appear in the next FLORIS release.

RRiva:
Nice! Thanks a lot 🙂

PR #1121, opened 2025-06-18
Convert angles from deg to rad fewer times

RRiva:
# Convert angles from deg to rad fewer times

Convert angles from deg to rad once and re-use the result.

## Related issue

None

`plot_turbine_rotors` and `controller_dependent_operation_model`.

### DIFF
--- a/floris/core/turbine/controller_dependent_operation_model.py
+++ b/floris/core/turbine/controller_dependent_operation_model.py
@@ -520,6 +520,7 @@ class ControllerDependentTurbine(BaseOperationModel):
 
             # Yawed case
             mu = np.arccos(cosd(gamma) * cosd(tilt))
+            pitch_in_plus_beta = np.deg2rad(pitch_in + beta)
             data = (
                 sigma,
                 cd,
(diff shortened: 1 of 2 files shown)


RRiva:
My bad 😅  I'll fix the CI

misi9170:
Sorry about the waiting here between CI runs; that's a little annoying, let me see if I can just get them to run automatically without me having to approve them.

Overall though, I'm less sure about this PR than the `keepdims` one. In general, we prefer to keep things as clear as possible in FLORIS unless there is a significant performance benefit from less clarity. I guess what is "more clear" is always up for debate, but in my opinion the changes here move away from clarity for the sake of performance. 

I'm a little embarrassed to say I have to look up what the difference between `.ravel()` and `.flatten()`---`.ravel()` seems like a very useful alternative!

misi9170:
Ah, regarding the CI runs, we have the FLORIS settings set to require approval for first-time contributors. I guess since the other PR has not yet been merged, github still considers you a first-time contributor. I think that setting makes sense, so I'll keep it that way, but hopefully you won't need to deal with this CI run approval process next time.

RRiva:
No problem, at DTU we very much favor speed over clarity 🙂 

Regarding your doubt, both `ravel()` and `flatten()` will give you a 1D array, but `ravel()` returns a view, while `flatten()` a copy. So, if you only need to access the data, you should use `ravel()`.

misi9170:
Hi @RRiva , Coming back through and clearing up some older PRs, and as I said, I think I will reject this one on the grounds that although I don't doubt that the performance might improve, I think that improvement is likely very slight. Feel free to reopen if you disagree.

Issue #1117, opened 2025-06-16
Feature request: Sample Turbulence Intensity at x, y, z locations


hannahgielen:
# Sample Turbulence Intensity at x, y, z locations

Hi, I would like to propose a feature similar to the existing `sample_flow_at_points(...)` function, but for turbulence intensity (TI). This would allow users to easily retrieve TI values at arbitrary points (`x`, `y`, `z`).
Currently, I think the only way to extract TI at specific locations is to manually access the internal flow field arrays after calling `solve_for_points(...)`. 

## Proposed solution

Add a method that is similar to `sample_flow_at_points`, but returns turbulence intensity. I'm not completely sure how you would implement it internally, but I imagine something like the following, based on `sample_flow_at_points(...)`:

```python
def sample_ti_at_points(self, x, y, z):
    if not len(x) == len(y) == len(z):
        raise ValueError("x, y, and z must be the same size")

    self.core.solve_for_points(x, y, z)
    return self.core.flow_field.turbulence_intensity_field_sorted[0, 0, :, 0]
```

## Alternatives considered

Currently, I try to work around this by doing something like:
```python
self.core.solve_for_points(x, y, z)
ti = self.core.flow_field.turbulence_intensity_field_sorted[0, 0, :, 0]
```

misi9170:
Hi @hannahgielen ,

Thank you for the suggestion! It's a good idea, and I think your code at the top level would work as you describe. Unfortunately, in the underlying solver, the turbulence intensity is not evaluated at each location, only at the turbine locations. For instance, the [`full_flow_sequential_solver()`](https://github.com/NREL/floris/blob/a6fc5d35aa32614edc450dc399c42af60a816887/floris/core/solver.py#L263) is called by `Core.solve_for_points()`, but this does not actually re-evaluate the turbulence; the turbulence is calculated beforehand at the turbine locations in the `sequential_solver()` and then [used in the main `full_flow_sequential_solver()`](https://github.com/NREL/floris/blob/a6fc5d35aa32614edc450dc399c42af60a816887/floris/core/solver.py#L374-L375) code. 

[...]

PR #1129, opened 2025-07-25
TI sampling at arbitrary points 

hannahgielen:
## Turbulence Intensity Sampling at Arbitrary Points 

This pull request adds a function to compute the turbulence intensity at arbitrary points in the flow field the same way ```sample_flow_at_points()``` does. This is done by adding the turbulence intensity calculation from ```sequential_solver()``` to ```full_flow_sequential_solver()```. It introduces the functions ```sample_ti_at_points()```, which returns a 3D array of turbulence intensity values at specified x,y,z points and ```solve_ti_at_points()```, which uses ```PointsGrid``` to compute the TI field. Both work similar to ```sample_flow_at_points()``` and ```solve_at_points()```.  

## Related issue

Issue #1117

```floris/floris_model.py```
```floris/core/core.py```
```floris/core/solver.py```

Only the ```full_flow_sequential_solver()``` is modified,  so the function is currently not working for the ```cc_solver``` and ```full_flow_empirical_gauss_solver```.

### CODE REVIEW

**Comment by misi9170** [comment]
Thanks @hannahgielen , this looks great!

I'll work with this a bit and push some changes directly to your branch as I move through some bits and pieces:
- [x] Get existing tests/checks passing
- [x] Add new tests
- [x] Add code to other full flow solvers where appropriate based on your implementation in `full_flow_sequential_solver()`
- [x] Possibly add an example in or add to an existing example
- [x] Add equivalent functionality to `ParFlorisModel`, as in #1059 

Once I've done that (hopefully in the next week or two), I'll ask you to look over the changes I've made and get a final review from another FLORIS developer. Assuming that all goes smoothly, I'm hoping we can get this merged in in the next month or so and have it appear in the next minor FLORIS version (v4.5)

**Comment by misi9170** [comment]
I've now added the code to `full_flow_cc_solver()` so that the cumulative curl (CC) model can also be run. The code now supports running with:
- Gauss, GCH, TurboparkGauss*, and Jensen models (in `full_flow_sequential_solver()`)
- CC model (in `full_flow_cc_solver()`)

The legacy Turbopark model is not supported because it does not have a full flow solver implemented and is deprecated by the TurboparkGauss model. The Empirical Gaussian (EmG) model is minimally supported in `full_flow_empirical_gauss_solver()`, but note that since the EmG model does not explicitly use TI, the ambient TI is simply returned at all sample points. It is possible that we could add a similar approach for EmG that instead returns the [wake induced mixing](https://nrel.github.io/floris/empirical_gauss_model.html#wake-induced-mixing) factor, but I will leave that as future work.

I've now added an example examples/examples_get_flow/003_extract_turbulence_intensity_at_points.py (note the similarity to 002_extract_wind_speed_at_points.py). This produces the following figure when run using the GCH model (with the Crespo-Hernández added TI model):
image

I understand this to be correct; there is a top-hat shape with a slight "dip" at the top caused by a shortening of the "downstream" distance when turbines are not directly aligned (see similar shape using the Jensen velocity model in the 002_extract_wind_speed_at_points.py example).

I still need to add an equivalent TI sampling method to the `ParFlorisModel`. Once that is done, and if we confirm that the sampled TI in the figure above looks consistent with the form of the Crespo-Hernández model, I will ask others for review.


*the TurboparkGauss model runs but simply returns the ambient TI unless the `turbulence_model` is changed to `"crespo_hernandez"` in the input file.

**Comment by misi9170** [comment]
I've now also added equivalent functionality to `ParFlorisModel`. This was slightly more complex than I had originally imagined, but I've abstracted the `ParFlorisModel.solve_flow_at_points()` method into a new `ParFlorisModel._solve_value_at_points` method that can now handle velocities or turbulence intensity, and in future may support sampling other quantities. I've added a test for this, which compare the TI sample with `FlorisModel` to that sampled with `ParFlorisModel`. I also fixed a small typo bug in the test for `ParFlorisModel.solve_flow_at_points`.

This PR is now ready for final review.

File: floris/core/solver.py [hunk 1]
```diff
@@ -441,10 +447,37 @@ def full_flow_sequential_solver(
...
+            x_i,
+            rotor_diameter_i,
+            axial_induction_i,
+        )
+
+        # Calculate locations where wake-added turbulence (WAT) applies
+        area_overlap = np.where(velocity_deficit * flow_field.u_initial_sorted > 0.05, 1, 0)
+
+        # Modify wake added turbulence by wake area overlap
+        downstream_influence_length = 15 * rotor_diameter_i
```
paulf81, line 462: Sorry if this is a dumb question, but why 15 specifically?

floris/core/solver.py
```diff
@@ -871,9 +910,37 @@ def full_flow_cc_solver(
...
+            x_i,
+            rotor_diameter_i,
+            axial_induction_i
+        )
+
+        # Calculate wake overlap for wake-added turbulence (WAT)
+        area_overlap = np.where(turb_u_wake > 0.05, 1, 0)
+
+        # Modify wake added turbulence by wake area overlap
+        downstream_influence_length = 15 * rotor_diameter_i
```
paulf81, line 925: Same probably not brilliant question

paulf81 (approved):

paulf81 (approved):
Looks great, thank you.  Just had one part I didn't understand so left a small question

floris/core/solver.py
misi9170: Not a dumb question...

15D is how the [influence of TI is defined in the main `sequential_solver()`](https://github.com/NREL/floris/blob/a6fc5d35aa32614edc450dc399c42af60a816887/floris/core/solver.py#L238), and [similarly for the `cc_solver()`](https://github.com/NREL/floris/blob/a6fc5d35aa32614edc450dc399c42af60a816887/floris/core/solver.py#L675). 

For the purpose of this PR, that influence is just being matched in the `full_flow` versions of those solvers. Will dig a bit more on the original 15D specification.

misi9170: The 15D influence was introduced in the `sequential_solver` at the [same time as the Crespo-Hernández turbulence intensity model](https://github.com/NREL/floris/commit/869d0a1d83f0d471aaf2f34e1047ba9b31802696), and was then copied over to the `cc_solver()` when added turbulence was [added to the CC model](https://github.com/NREL/floris/commit/3bc5a6ac7c7fe87a1d7228c3efa56218d9db5801). 

The first introduction was during the run-up to release v3, and there wasn't a ton of documentation and formal pull requests at that time. However, looking back at the [original 1996 Crespo-Hernández paper](https://www.sciencedirect.com/science/article/pii/016761059500033X), it seems that the experiments used to parametrize the model ran to a maximum of 15D downstream of the turbine; so it seems we also limited the influence to that range to avoid extrapolating on the reported results.

PR #1134, opened 2025-08-04
Curled wake model wind farm solver implementation

tonyinme:
This is a draft PR to get the conversation started on the implementation of the curled wake model in floris. 

@misi9170 and @bayc have been helping with the implementation. 
The backbone code structure was suggested by @misi9170 

Here are some of the things that need to be addressed. 

- [x] Extract CWM code into a separate file
- [x] Modify code and solver to use multiple indices (currently only supports one wspd and wdir).
- [ ] Use the background flow features from flow_field.


tonyinme:
@misi9170 , thanks for meeting with me earlier today and talking through the changes. 
I pushed some of the changes we discussed (new solver_cwm file, extracted plane function to curled_wake, etc). 
More to come in the next few days.

misi9170:
@tonyinme , I know you're still working here---I've just updated some formatting and requirements so that the tests run through. If you want to be able to address the formatting yourself, you can pip install floris with the develop option
```
pip install -e ".[develop]"
```
and then run
```
pytest .
```
to run the tests and 
```
pre-commit run --all-files
```
to check formatting. That will catch most formatting issues (although it only runs in the current python version, which isn't completely bulletproof because the tests/checks run in various python versions 3.9--3.13, which is why I had to make a couple of pushes here to get the numba install right...).

misi9170:
Looks like numba version 0.61.2 is installed for all python versions 3.10 and up; but [0.60.0 is installed for python 3.9](https://github.com/NREL/floris/actions/runs/17044869626/job/48317914975?pr=1134).

We are planning to [remove support for python 3.9](https://github.com/NREL/floris/pull/1079) in the coming months as it reaches end-of-life. We can update the pyproject toml with an exact specification for numba later, once we get closer to final implementation and review.

tonyinme:
Perfect, thanks @misi9170 !

I've made some improvements today to the turbulence model. 
I only have a few more updates to make (add veer and curl) and then we should be good to move forward with the next steps.

tonyinme:
@misi9170 , thanks for the help with this implementation. 
I now have a working implementation that is ready for review.

misi9170:
Hi @tonyinme ,

I've now gone in and reworked the code to allow it to operate over multiple wind conditions (findices) at once. That's working fine, although I wasn't able to get away from at least some sort of loop, so there is a loop that runs n_findex x n_turbines times (but at least it doesn't run n_findex x n_planes times :) ).

I also refactored the code in curled_wake.py to use vectorized operations and remove the numba decorators and import. I still need to investigate whether this is slower than using numba though.

I'll now need to go through everything slowly and clean things up. As I'm going that, I'll make a list of items that we still need to address.

But, in the meantime, you can check out the operation over multiple conditions in examples/examples_curled_wake/01_test_cwm_solve.py

tonyinme:
Hi @misi9170 , excellent! Thanks for working your magic and updating the code to use the proper indexing. 
In my experience numba always makes these specific operations of the curled wake model faster (~2X or more) on most systems. 
But we can use whatever works best for the FLORIS infrastructure. 

I tested the code and it is similar than before so all is good from my end. 
Let me know when you have the tasks defined and I can go in and make any changes/adjustments. 

Thanks!

PR #1135, opened 2025-08-09
Add `turbine_average_velocities` property to UncertainFlorisModel

paulf81:
# Add turbine average velocity property to UncertainFLorisModel

This PR adds a property to allow getting the per-turbine average velocities from an UncertainFlorisModel.  This property exists for FlorisModel.  To provide this capability in UncertainFLorisModel requires mirroring the logic in get_turbine_powers to go from the expanded internal FlorisModel (which includes all the unique conditions after expanding across neighboring wind directions according to the wd_std) back to the un_expanded originally requested wind directions.  

Several tests are included to check the functino works as expected

## Impacted areas of the software
uncertain_floris_model.py

## Test results, if applicable
New tests pass

PR #1136, opened 2025-08-12
[BUGFIX] Specify resolution in scipy yaw optimization regression test

misi9170:
Recent PRs #1126 , #1129 , #1134 #1135 have had the `test_scripy_yaw_opt` [fail](https://github.com/NREL/floris/actions/runs/16893988771) for python versions 3.11, 3.12, and 3.13. It appears there is some minor numeric change in the outcome of the scipy optimization. This PR simply adds a resolution to the regression test to specify that the values only need to match to a relative tolerance of 1x10^-5 and an absolute tolerance of 0.01. For optimal yaw angles, a difference at the third decimal place is inconsequential.

I've already made a fix for this on #1126 and #1129, but I decided to fix it in a standalone PR. Once this is merged, I will back merge develop into those branches.

### DIFF
--- a/tests/reg_tests/yaw_optimization_regression_test.py
+++ b/tests/reg_tests/yaw_optimization_regression_test.py
@@ -183,4 +183,5 @@ def test_scipy_yaw_opt(sample_inputs_fixture):
         print(baseline_scipy.to_string())
         print(df_opt.to_string())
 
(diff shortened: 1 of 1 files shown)


paulf81 (approved):
Agree with this solution

Issue #1138, opened 2025-09-01
Bug in yaw_optimizer_geometric.py


ziyuwind:
In  yaw_optimizer_geometric.py, there is a bug:
"in_Jensen_wake = (abs(y_dists) < spread * x_dists + rotor_diameter)"

According to the Stanely et al. (2023) https://wes.copernicus.org/articles/8/1341/2023/，
it should be "in_Jensen_wake = (abs(y_dists) < spread * x_dists + rotor_diameter/2.)"

misi9170:
Thanks for the bug report @ziyuwind !

[This line](https://github.com/NREL/floris/blob/41d09b46c88034b93caea370640ae0c7da98c6e8/floris/optimization/yaw_optimization/yaw_optimizer_geometric.py#L184) is the one mentioned. As you point out, the expression from [Stanley et al., 2023](https://wes.copernicus.org/articles/8/1341/2023/) is $r_\text{wake} = 0.1x + r_\text{turbine}$ (see top paragraph of right column on page 1343). Here, $0.1$ is the `spread` and $r_\text{turbine}$ is the rotor radius (= `rotor_diameter/2`).

This bug was introduced during the implementation of the geometric yaw optimization in #681. I'll look around there to see if there are any reasons that this shouldn't simply be updated to correct the discrepancy.

misi9170:
The implementation in #681 was originally based on [this repository](https://github.com/pjstanle/GeometricYaw), but significantly refactored. It does appear that I introduced a bug in translating that code; in particular, [this file](https://github.com/pjstanle/GeometricYaw/blob/master/geometric_yaw.py#L24) seems to use the expected `rotor_diameter/2`. 

I'll create a bugfix for this.

misi9170:
Addressed in #1143 and will appear in next FLORIS release.

PR #1143, opened 2025-09-18
[BUGFIX] Fix incorrect implementation in geometric yaw optimization

misi9170:
Addresses bug flagged in #1138 by adding appropriate `/2.0` divisor and adds a test for the `_process_layout()` function. 

The updated implementation passes the new tests; the old version (without the `/2.0`) fails the test.

I also made a couple of updates to use `np.inf` rather than `0.` as an indicator of "far downstream" turbines. I checked that this does not affect outputs of regression tests and examples, and it simplifies the code somewhat and (in my opinion) makes the method clearer. Affected code: `_process_layout()` and `_get_yaw_angles()`. As these are both "hidden" functions, I'm comfortable that we can change their behavior without considering this a breaking change.

Note that all existing tests still pass (see tests/geometric_yaw_unit_test.py and tests/reg_tests/yaw_optimization_regression_test.py). Also, no change appears in examples/examples_control_optimization/006_compare_yaw_optimizers.py

### DIFF
--- a/floris/optimization/yaw_optimization/yaw_optimizer_geometric.py
+++ b/floris/optimization/yaw_optimization/yaw_optimizer_geometric.py
@@ -188,11 +188,6 @@ def _process_layout(
     dx = x_dists.min(axis=1)
     dy = y_dists[range(len(turbine_x)), x_dists.argmin(axis=1)]
 
-    # Handle last turbine downstream
-    furthest_ds_turb_idx = np.where(dx == np.inf)[0]
-    dx[furthest_ds_turb_idx] = 0.
-    dy[furthest_ds_turb_idx] = 0.
-
     return dx/rotor_diameter, dy/rotor_diameter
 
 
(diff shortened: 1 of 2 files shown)


floris/optimization/yaw_optimization/yaw_optimizer_geometric.py
```diff
@@ -219,7 +195,7 @@ def _get_yaw_angles(
     |.......................................
```
paulf81: What is this a picture of?

paulf81: Maybe could give a short description of this function here

paulf81 (approved):
Tests pass, reviewed code changes and new tests

floris/optimization/yaw_optimization/yaw_optimizer_geometric.py
misi9170: Good idea---I don't know (or recall) what the picture is. I've updated the docstrings across the document.

PR #1145, opened 2025-10-03
Added run_solver flag to sample_flow_at_points and sample_ti_at_points

RRiva:
# Added `run_solver` flag to `sample_flow_at_points` and `sample_ti_at_points`

The functions `sample_flow_at_points` and `sample_ti_at_points` are fine when used independently. However, if the user needs both the wind speed and TI field for the same grid, then the solver is called twice. I've added a flag to prevent that.

## Related issue

None.

- `FlorisModel`
- `Core`

### DIFF
--- a/floris/core/core.py
+++ b/floris/core/core.py
@@ -264,8 +264,6 @@ class Core(BaseClass):
         else:
             full_flow_sequential_solver(self.farm, self.flow_field, field_grid, self.wake)
 
-        return self.flow_field.u_sorted[:,:,0,0] # Remove turbine grid dimensions
-
     def solve_for_velocity_deficit_profiles(
         self,
         direction: str,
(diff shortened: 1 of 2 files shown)


misi9170:
Hi @RRiva ,

I think this makes sense, but it seems like we'll need some sort of error handling for the case where a user sets `run_solver = False` but a solve has not already happened? I think that should be possible, and I'm happy to put something in and write tests for it, but just wanted to see if you'd already thought of this.

RRiva:
Hi @misi9170, thanks for the feedback, I'll add some error handling 🙂

misi9170:
> Hi @misi9170, thanks for the feedback, I'll add some error handling 🙂

Thanks @RRiva ! In case you hadn't already found it, I think you should be able to get most of what you need with something like `if self.core.state is not State.USED:` (where `self` is the `FlorisModel`), although you might also have to check that the output variables are the correct shape to be consistent with a `solve_for_points`-type solve, since `self.core.state` will also be `State.USED` under a normal `run()`.

PR #1146, opened 2025-10-17
Add IEA 22MW reference wind turbine to turbine library

misi9170:
This PR adds the [IEA 22MW reference wind turbine](https://iea-wind.org/wp-content/uploads/2024/04/iea-22-280-rwt-report.pdf) to FLORIS' turbine library. ~Data is from the IEA: https://github.com/IEAWindSystems/IEA-22-280-RWT/blob/main/Documentation/IEA-22-280-RWT_tabular.xlsx.~ Data generated by @ptrbortolotti using OpenFAST v4.1.2 and ROSCO v2.10.2. OpenFAST input files used to run the simulations are attached below.

The documentation is also updated.

To view the data, run examples/examples_turbine/001_check_turbine.py. This prints the power and thrust coefficient curve for the added 22MW turbine, as well as the other reference turbines available in FLORIS:

image

Note similarity to the curves in Figure 20 of the [definition document](https://iea-wind.org/wp-content/uploads/2024/04/iea-22-280-rwt-report.pdf):
image
image

### OpenFAST input data
[IEA22_Floris.zip](https://github.com/user-attachments/files/23756212/IEA22_Floris.zip)


floris/turbine_library/iea_22MW.yaml
```diff
@@ -0,0 +1,178 @@
...
+    - 21999.90199
+    - 22000.08648
+    - 22000.0445
+    - 22000.0247
+    - 22000.01503
+    - 22000.00988
+    - 22000.00673
+    - 22000.00464
+    - 22000.00305
+    - 22000.00173
```
paulf81: Recommend you shave off these decimal places above rated, or at least make monotonic, there is a slight wiggle which messes up some of the optimizers

paulf81 (approved):
made myself a reviewer, all in all looks great!  Had a small comment to remove minor wiggles in the above rated power.  but I checked the values and the example runs on my computer

paulf81:
Above rated power zoomed in: 
image

floris/turbine_library/iea_22MW.yaml
misi9170: Good call, will do that

misi9170:
Thanks @paulf81 ! I'm also taking this opportunity to rename the example 001_reference_turbines.py rather than 001_check_turbine.py, since I think it's more about plotting the response of the reference turbines than "checking" anything.

misi9170:
To do:
- [x] Check whether rotor diameter includes coning, rotor pre-bend
- [x] Confirm that data is for a fixed tower
- [x] Check reference air density, reference shear

misi9170:
After discussion with @dzalkind , our current understanding is that:
- The rotor diameter (284.0m) includes the rotor cone angle but does not include pre-bend, as expected.
- The data appears to be for a fixed tower (no floater or extra platform tilt other than the shaft tilt).
- The data appears to be generated with a reference air density of 1.225 kg/m^3 and a shear exponent of 0.14. Note that FLORIS does not currently handle "reference" shear exponents. I will open a separate issue to discuss possible future handling for reference wind shears. #1147 

EDIT: new OpenFAST data uses zero shear.

misi9170:
After further discussion with @ptrbortolotti and @dzalkind , the power and thrust coefficient curves were regenerated using OpenFAST to ensure that all relevant degrees of freedom are included and at sufficient wind speed resolution (particularly near the knee of the power curve).

PR #1152, opened 2025-11-07
Allow `multidim_conditions` to be specified across multiple findices

misi9170:
## Description

Until now, users were limited to a single condition tuple for `multidim_conditions` per `run()` of the `FlorisModel`. For instance, users could set `fmodel.set(multidim_conditions={"Tp": 1, "Hs": 5})` for an arbitrary number of findices (wind speed/wind direction/turbulence intensity combinations), but would need to call `.set()` and `.run()` again to change the `multidim_conditions`. This PR allows `multidim_conditions` to be set as a dictionary where each value is an array of length `n_findex`; these conditions are then paired to the wind speed/wind direction/turbulence intensity at that findex. 

Scalar specification of `mutlidim_conditions` values is still allowed; these are broadcast across all findices, as before. 

## Affected areas of the code
The majority of the working code is limited to floris/core/turbine.py, in particular, the `power()`, `thrust_coefficient()`, and `axial_induction()` functions. This is where the `multidim_conditions` are interpolated, grouped, evaluated, and rebroadcast.

Additionally, there will be light changes to user-level classes (`FlorisModel` and its subclasses and the `WindData` objects). For the `WindData` objects, while it would have been convenient to add `multidim_conditions` to the `.unpack()` methods, this would be a breaking change, so I've instead opted to create a new `unpack_multidim_conditions()` method.

## Tests
All existing tests pass. Newly added tests for the core code pass. New integration tests running through the user-level classes are on their way.

## Examples
I've updated the 002 and 003 examples in examples/examples_multidim/ to demonstrate the new functionality. Their outputs do not change, but there is now less looping needed. For example 003, I decided to leave the looping option available, set by the flag `loop_over_ti` near the top of the script. Setting this to `True` uses the existing FLORIS functionality, and loops over the two values of turbulence intensity being evaluated. Setting to `False` instead uses the new functionality. The resulting plot is the same for both, as expected:
image

#### Core code

- [x] Implement core code

#### User-level code
- [x] `FlorisModel`
- [x] `ParFlorisModel`
- [x] `UncertainFlorisModel`
- [x] `TimeSeries`
- [x] `WindRose`
- [x] `WindTIRose`
- [x] `WindRoseWRG` (no change; simply returns `None`)

#### Testing
- [x] Test for core code
- [x] Integration tests for `FlorisModel` and subclasses; `WindData` objects


misi9170:
Just pointing out a couple of things here that might not be immediately obvious:
- In the `WindData` classes, I've added a new method `unpack_multidim_conditions()` to "unpack" the multidim_condition dictionary. This unpacking would more naturally occur in the main `unpack()` methods, but this would be a breaking change on the user-level `WindData` objects, so I created a new method. In a future (major) FLORIS version, we should remove the `unpack_multidim_conditions()` method and incorporate it into the `unpack` method.
- In `WindRose` and `WindTIRose`, I've introduced new temporary variables `_n_wd`, `_n_ws`, and `_n_ti` to help perform checks more succinctly. These are not kept after `__init__` is complete for these classes.

misi9170:
python 3.14 macos tests fail due to how python versions are specified. This is being addressed in #1160

misi9170:
Still to do: run some timed tests for `.set()` and `.run()` when `multidim_conditions` is scalar, and specified directly or specified on a `WindData` object. Ideally, there would be no impact to speed in these cases. Speed may decrease somewhat for array `multidim_conditions`; this is acceptable.

misi9170:
I have now built a script for timing the `set()` and `run()` methods in various configurations: with a scalar or vector `multidim_conditions`; and with `multidim_conditions` specified on the `WindRose` or passed directly to `set`. The timing script is as follows:
```python
import timeit
import numpy as np

from floris import FlorisModel, WindRose

fmodel = FlorisModel("defaults")

running_baselines = True

np.random.seed(0)

n_turbines = 10
n_wd = 72 # 100 doesn't work---bug in WindRose??
n_ws = 10

wr_basic = WindRose(
    wind_directions=np.linspace(0.0, 360.0, n_wd, endpoint=False),
    wind_speeds=np.linspace(4.0, 16.0, n_ws),
    ti_table=0.06,
    freq_table=(1.0 / (n_wd * n_ws)) * np.ones((n_wd, n_ws)),
)

fmodel.set(
    layout_x=np.linspace(0.0, n_turbines * 500.0, n_turbines),
    layout_y=np.zeros(n_turbines),
    wind_data=wr_basic,
    turbine_type=["iea_15MW_multi_dim_cp_ct"] * n_turbines,
    reference_wind_height=150.0,
)

scalar_md_condition = {"Tp": 3, "Hs": 2}

def md_scalar_set():
    fmodel.set(multidim_conditions=scalar_md_condition)

execution_time_sss = timeit.timeit(md_scalar_set, number=100)
print(f"MD scalar set avg. time: {execution_time_sss / 100 :.4f} seconds per call")
# ~ 0.02 seconds per call

def md_scalar_run():
    fmodel.run()

fmodel.set(multidim_conditions=scalar_md_condition)
execution_time_ssr = timeit.timeit(md_scalar_run, number=10)
print(f"MD scalar run avg. time: {execution_time_ssr / 10 :.4f} seconds per call")
# ~0.13 seconds per call

vector_md_condition = {
    "Tp": np.random.uniform(2, 4, size=n_wd * n_ws),
    "Hs": np.random.uniform(1, 6, size=n_wd * n_ws),
}

if running_baselines:
    exit()

def md_vector_set():
    fmodel.set(multidim_conditions=vector_md_condition)

execution_time_vss = timeit.timeit(md_vector_set, number=100)
print(f"MD vector set avg. time: {execution_time_vss / 100 :.4f} seconds per call")
# ~ 0.02 seconds per call

def md_vector_run():
    fmodel.run()

fmodel.set(multidim_conditions=vector_md_condition)
execution_time_vsr = timeit.timeit(md_vector_run, number=10)
print(f"MD vector run avg. time: {execution_time_vsr / 10 :.4f} seconds per call")
# ~0.2 seconds per call

## Now test with multidim conditions specified on wind rose
wr_md_single = WindRose(
    wind_directions=np.linspace(0.0, 360.0, n_wd, endpoint=False),
    wind_speeds=np.linspace(4.0, 16.0, n_ws),
    ti_table=0.06,
    freq_table=(1.0 / (n_wd * n_ws)) * np.ones((n_wd, n_ws)),
    multidim_conditions=scalar_md_condition,
)

def md_wr_scalar_set():
    fmodel.set(wind_data=wr_md_single)

execution_time_sws = timeit.timeit(md_wr_scalar_set, number=100)
print(f"MD wind rose scalar set avg. time: {execution_time_sws / 100 :.4f} seconds per call")
# ~ 0.02 seconds per call

def md_wr_scalar_run():
    fmodel.run()

fmodel.set(wind_data=wr_md_single)
execution_time_swr = timeit.timeit(md_wr_scalar_run, number=10)
print(f"MD wind rose scalar run avg. time: {execution_time_swr / 10 :.4f} seconds per call")
# ~0.2 seconds per call [!]

shaped_md_condition = {
    "Tp": np.random.uniform(2, 4, size=(n_wd, n_ws)),
    "Hs": np.random.uniform(1, 6, size=(n_wd, n_ws)),
}

wr_md_vector = WindRose(
    wind_directions=np.linspace(0.0, 360.0, n_wd, endpoint=False),
    wind_speeds=np.linspace(4.0, 16.0, n_ws),
    ti_table=0.06,
    freq_table=(1.0 / (n_wd * n_ws)) * np.ones((n_wd, n_ws)),
    multidim_conditions=shaped_md_condition,
)

def md_wr_vector_set():
    fmodel.set(wind_data=wr_md_vector)

execution_time_vws = timeit.timeit(md_wr_vector_set, number=100)
print(f"MD wind rose vector set avg. time: {execution_time_vws / 100 :.4f} seconds per call")
# ~ 0.02 seconds per call

def md_wr_vector_run():
    fmodel.run()

fmodel.set(wind_data=wr_md_vector)
execution_time_vwr = timeit.timeit(md_wr_vector_run, number=10)
print(f"MD wind rose vector run avg. time: {execution_time_vwr / 10 :.4f} seconds per call")
# ~0.2 seconds per call
```

When running on the default branch and with `running_baselines = True`, I see the following output:
```
MD scalar set avg. time: 0.0203 seconds per call
MD scalar run avg. time: 0.1343 seconds per call
```

When running on this feature branch and with `running_baselines = True`, this becomes
```
MD scalar set avg. time: 0.0214 seconds per call
MD scalar run avg. time: 0.1459 seconds per call
```
which I would say is minimal change.

Running again with `running_baselines = False` to run the full test suite, I get
```
MD scalar set avg. time: 0.0230 seconds per call
MD scalar run avg. time: 0.1416 seconds per call
MD vector set avg. time: 0.0228 seconds per call
MD vector run avg. time: 0.2006 seconds per call
MD wind rose scalar set avg. time: 0.0230 seconds per call
MD wind rose scalar run avg. time: 0.2130 seconds per call
MD wind rose vector set avg. time: 0.0208 seconds per call
MD wind rose vector run avg. time: 0.2096 seconds per call
```

This is all essentially as expected. `set()` timing is not impacted. `run()` timing increases when using the vector conditions, taking approximately 30% longer (increasing from ~0.15 seconds to ~0.20 seconds) in the non-wind-rose case. 

The one "outlier" here is that running scalar `multidim_conditions` passed to the `WindRose`, rather than directly to `set()`, takes ~0.20 seconds (similar runtime to vector `multdim_conditions`). I somewhat expected this, since the `WindRose` currently [broadcasts the scalar condition](https://github.com/misi9170/floris/blob/7dbac6d0f1e8bd59ba5a8bed67d24b0f7356a467/floris/wind_data.py#L308-L312) before passing it to `FlorisModel`, so makes the operation a vector operation. Perhaps this is an acceptable penalty for simplicity; however, I will see if this performance hit is negated if I leave the scalar `multidim_conditions` as a special case.

misi9170:
I have now pushed a change to allow scalar `multidim_conditions` to be retained on the `WindData` objects and passed through to `FlorisModel`. This does indeed speed up performance, with the `MD wind rose scalar run avg. time` improving from ~0.2 seconds to ~0.15 seconds (that is, reverting to the `MD scalar run avg. time`). Although this makes the code slightly clunkier, I consider this worthwhile for a ~30% speed-up. Also, these tests are only running a smallish number of conditions (720), and I expect the difference might grow for many more conditions.

paulf81:
@misi9170 : Recommend updating this file: docs/multidimensional_wind_turbine.ipynb

examples/examples_multidim/002_multi_dimensional_cp_ct_2Hs.py
```diff
@@ -15,45 +15,46 @@
```
paulf81, line 20: Recommend updating the docstring of this file to mention changes made in FLORIS vx.x and that this illustrates using array-style multidim to enable simulating two different conditions in one run

examples/examples_multidim/002_multi_dimensional_cp_ct_2Hs.py
```diff
@@ -15,45 +15,46 @@
...
-fmodel_dict_mod["flow_field"]["multidim_conditions"]["Hs"] = 1.0
-fmodel_hs_1 = FlorisModel(fmodel_dict_mod)
-
-# Set both cases to 3 turbine layout
 fmodel.set(layout_x=[0.0, 500.0, 1000.0], layout_y=[0.0, 0.0, 0.0])
-fmodel_hs_1.set(layout_x=[0.0, 500.0, 1000.0], layout_y=[0.0, 0.0, 0.0])
 
-# Use a sweep of wind speeds
-wind_speeds = np.arange(5, 20, 1.0)
+# Conditions to evaluate
```
paulf81, line 25: Recommend like docstring above to descibe a bit more how arrays in `multidim_conditions` conditions are being used in this case to enable one-run of FLORIS.  Could be in comparison to pre vx.x where seperate runs of scalar `multidim_conditions` would be required

examples/examples_multidim/003_multi_dimensional_cp_ct_TI.py
```diff
@@ -20,25 +20,54 @@
 # Set both cases to 3 turbine layout
```
paulf81, line 26: Recommend updating the docstring of this file as well (appears at head of file in example docs)

paulf81 (approved):
Left a few small documentation-focused comments here and there but all in looks great so hitting approve now!

misi9170:
Thanks @paulf81 ! Good ideas; I've done both. I'll take a last look over this tomorrow morning and then merge it in.

PR #1153, opened 2025-11-10
[BUGFIX] Heterogeneous inflows with `ParFlorisModel` [FLORIS v4.5.1]

misi9170:
While working on #1152, I realized that the `heterogeneous_inflow_config` is not properly handled on the `ParFlorisModel`. In particular, the `heterogeneous_inflow_config` is not correctly broken down into subsets to pass off to the parallelized `FlorisModel` workers, meaning that all workers may be running the same subset of the `heterogeneous_infow_config` (instead of their correct subset).

This PR fixes the issue; since I consider this to be a fairly critical bug and the develop branch currently have some other changes that would be better to release in a minor version release, I'm directing this PR to main; once (squashed and) merged, I will:
- [x] increment the version number on main
- [x] create a patch release (v4.5.1) pointing at main
- [x] merge main back into develop.

To fix the issue and hopefully avoid future similar issues, I've moved from making changes to the `fmodel_dict` to pass to the parallel workers to explicitly passing information through the keyword arguments to `set()`. This should be done for all future arguments where there are findex dimensions (scalar arguments to `set()`, such as `reference_wind_height` or `wind_shear`, remain as being passed on `fmodel_dict`).

I've added a test for the new capability, as well as adding tests for passing control set points. These all now pass. I also made the type-hinting of `FlorisModel` for `heterogeneous_inflow_config` explicit, since this was missing before.

### DIFF
--- a/floris/floris_model.py
+++ b/floris/floris_model.py
@@ -1,3 +1,6 @@
+
+from __future__ import annotations
+
 import copy
 import inspect
 from pathlib import Path
--- a/floris/floris_model.py
+++ b/floris/floris_model.py
@@ -1,6 +1,3 @@
-
-from __future__ import annotations
-
 import copy
 import inspect
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


paulf81 (approved):
Looks great @misi9170 , I reviewed the code and the new tests which look nicely structured to test this issue.  Tests pass on my machine

PR #628, opened 2023-04-22
Add Empirical Gauss wake model for fitting SCADA data

misi9170:
This pull request is to incorporate the new Empirical Gaussian wake model (working name Geometric Model).

This PR includes:
* The new parametrization of the Gaussian wake model designed to be simpler to tune and more flexible for fitting observed wind farm data, consisting of 
    - a new velocity deficit model
    - a new velocity deflection model
    - a new "wake induced mixing" model to replace wake-added turbulence
* Vertical wake deflections due to rotor tilt
* Mirror wakes to address stumpiness and ground effects

TO DO before merge:
- [x] Add example to demonstrate usage of empirical_gauss model
- [x] Check default tuning, ensure defaults match the example input file emgauss.yaml
- [x] Check documentation in docs/empirical_gauss_model.md and develop docstrings for new functions, classes, and methods
- [x] Check/update current tests,
- [x] Develop regression tests

NOT included in this PR, but for possible future developments:
- Implement unit tests for lower-level functions as part of the simulation tools, e.g. `sigmoid_integral`
- Profile code and look for speed-up opportunities

PR #1158, opened 2025-11-13
[DOCS] Fix Empirical Gaussian wake-induced mixing term definition

misi9170:
The current version of the [documentation on the Empricial Gaussian model](https://nrel.github.io/floris/empirical_gauss_model.html), in particular the [wake-induced mixing (WIM) terms](https://nrel.github.io/floris/empirical_gauss_model.html#wake-induced-mixing), is inaccurate. This PR updates the documentation so that it matches the source code. The source code itself is not changed at all; this PR simply updates the documentation so that the description matches the form of the Empirical Gaussian model in the code. There were two errors in the documentation: first, the implementation uses a root-sum-squared approach for computing the WIM term, rather than a sum of (positive) values (that is, a 2-norm is used rather than a 1-norm); and second, the denominator of the off-diagonal terms in the sum (the downstream distances between turbines) are now squared. I apologize for any confusion caused---these errors came in with the original merger of the EmGauss model #628 , but[I only noticed them recently](https://github.com/NREL/floris/discussions/1148#discussioncomment-14765911). 

Because the FLORIS documentation builds off the develop branch, once this is merged into develop the docs page will be updated. Before I merge, I plan to add a test to confirm that the form of the WIM is as expected and explained in the documentation.

### DIFF
--- a/docs/empirical_gauss_model.md
+++ b/docs/empirical_gauss_model.md
@@ -113,7 +113,7 @@ the empirical model, explicit dependencies on turbulence intensity are removed
 completely to aid in tuning. Instead, a non-physical "wake-induced mixing
 factor" is specified for turbine $j$ as
 
-$$ \text{WIM}_j = \sqrt{\sum_{i \neq j} 
+$$ \text{WIM}_j = \sqrt{\sum_{i \neq j}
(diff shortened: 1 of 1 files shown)


paulf81 (approved):
Thanks for fixing!

PR #1160, opened 2025-12-06
Update name of laboratory

misi9170:
NREL has recently been renamed the National Laboratory of the Rockies (NRL). Further, the Alliance for Sustainable Energy has been renamed the Alliance for Energy Innovation. This pull request reflects these name changes. "NREL" is replaced with "NLR" where appropriate. There are a few places where the change is _not_ made:
- URLs containing NREL that continue to be correct (notably, github.com/NREL/...). 
- References to past NREL reports, most notably the NREL 5MW reference turbine.

I'm also taking this opportunity to bump up a few version numbers and years on the license file etc. Finally, the turboparkguass.py and layout_optimization_random_search.py files contained license headers that were removed from most files when we moved to the BSD 3-Clause license #810 , but these slipped back in on this file so I'm removing them now.

A second round of updates may be needed if URLs change in future.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -2,4 +2,4 @@ blank_issues_enabled: false
 contact_links:
   - name: Usage question
     url: https://github.com/NREL/floris/discussions
-    about: Have any questions about using FLORIS? Post in Discussions to engage with the NREL team and FLORIS community.
+    about: Have any questions about using FLORIS? Post in Discussions to engage with the NLR team and FLORIS community.
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -64,7 +64,7 @@ Add the results from unit tests and regression tests here along with justificati
 -->
 
 <!--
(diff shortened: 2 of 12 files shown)


paulf81 (approved):
Thanks for catching this!

misi9170:
In testing, [macos tests are now installing python version 3.14.1](https://github.com/misi9170/floris/actions/runs/20037757922/job/57463736037), which is failing the check for `requires-python = ">=3.10, <=3.14"` in pyproject.toml. On the other hand, windows and ubuntu tests are still installing python version 3.14.0, which satisfies `requires-python = ">=3.10, <=3.14"`. Although I'm having a hard time finding an explicit answer, it seems that [PEP 440](https://peps.python.org/pep-0440/#inclusive-ordered-comparison) treats `<=3.14` as `<=3.14.0` (adds `.0` as needed). Our intention was to allow the patch version to float upwards, i.e. to allow all `3.14.x`. To allow this as intended, I'm updating the pyproject.toml from 
```
requires-python = ">=3.10, <=3.14"
```
to
```
requires-python = ">=3.10, <3.15"
```
(i.e., strictly less than 3.15).

PR #1163, opened 2025-12-19
Instantiate `TimeSeries` from a csv file

misi9170:
`WindRose` and `WindTIRose` have capabilities of reading data from CSVs using their own (static) `read_csv_long()` methods. However, the `TimeSeries` class, until now, has not supported such functionality. This PR adds a new static method `read_csv()` to the `TimeSeries` class to enable such functionality.

To be discussed:
- The new `TimeSeries.read_csv()` allows users to specify a `value_col`, whereas `WindRose.read_csv_long()` and `WindTIRose.read_csv_long()` do not support reading in a `value_col`. Do we want to add this optional keyword arguments to  `WindRose.read_csv_long()` and `WindTIRose.read_csv_long()`?
- I chose to use `TimeSeries.read_csv()` rather than `TimeSeries.read_csv_long()` because I think the "long" format is implied by the "series" nature of the class. However, we could use `read_csv_long` instead to match the other wind data classes.
- None of the `WindData` classes have a `to_csv()` method, but this also seems like a natural counterpart. We could consider adding one for `TimeSeries` and possibly also `WindRose` and `WindTIRose`

Tagging @paulf81 , @ejsimley in case they have thoughts on this.

### DIFF
--- a/floris/wind_data.py
+++ b/floris/wind_data.py
@@ -2692,6 +2692,53 @@ class TimeSeries(WindDataBase):
             self.heterogeneous_map,
         )
 
+    @staticmethod
+    def read_csv(
+        file_path: str,
+        ws_col: str = "wind_speeds",
+        wd_col: str = "wind_directions",
+        ti_col_or_value: str | float = "turbulence_intensities",
+        value_col: str | None = None,
+        sep: str = ",",
+    ) -> TimeSeries:
+        """
+        Reads a CSV file and returns a TimeSeries object.
+
+        Args:
+            file_path (str): Path to the CSV file.
+            ws_col (str, optional): Column name for wind speeds. Defaults to "wind_speeds".
+            wd_col (str, optional): Column name for wind directions. Defaults to "wind_directions".
+            ti_col_or_value (str | float, optional): Column name for turbulence intensities
+                or a single float value. Defaults to "turbulence_intensities".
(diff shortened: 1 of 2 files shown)


paulf81:
> * The new `TimeSeries.read_csv()` allows users to specify a `value_col`, whereas `WindRose.read_csv_long()` and `WindTIRose.read_csv_long()` do not support reading in a `value_col`. Do we want to add this optional keyword arguments to  `WindRose.read_csv_long()` and `WindTIRose.read_csv_long()`?

Sure, does seem reasonable!

> * I chose to use `TimeSeries.read_csv()` rather than `TimeSeries.read_csv_long()` because I think the "long" format is implied by the "series" nature of the class. However, we could use `read_csv_long` instead to match the other wind data classes.

`TimeSeries.read_csv()` does make sense would always be long.  Could alias `TimeSeries.read_csv_long()` for completeness?  Or maybe the more perfect is the function is defined in `TimeSeries.read_csv_long()`  and  `TimeSeries.read_csv()` is *its* alias?

> * None of the `WindData` classes have a `to_csv()` method, but this also seems like a natural counterpart. We could consider adding one for `TimeSeries` and possibly also `WindRose` and `WindTIRose`

I could see it.  In FLASC I do sometimes call up FLORIS just for the WindRose classes and a work flow of loading a `TimeSeries` object, converting to `WindRose` and then exporting the `WindRose` to csv could have a place.

Issue #970, opened 2024-08-22
Bug report: Cumulative Curl does not work with the turbine_cubature_grid solver


Bartdoekemeijer:
# The CC model does not support turbine_cubature_grid solver
An error is raised when trying to run the cc model with the `turbine_cubature_grid` solver.

## How to reproduce
Here's a minimal reproducible example:
```
from floris import FlorisModel

fmodel = FlorisModel("inputs/cc.yaml")
fmodel.set(solver_settings={"type": "turbine_cubature_grid", "turbine_grid_points": 3})
fmodel.run()
fmodel.get_turbine_powers()

```

## Relevant output
This is the output of the example code, raising an error in `turbine.py` at line 301 (definition of `thrust_coefficient`):
```
Exception has occurred: ValueError
non-broadcastable output operand with shape (1,3) doesn't match the broadcast shape (1,1,9,3)
  File "/mnt/c/Users/Bart.Doekemeijer/OneDrive - Shell/Documents/python_scripts/floris_v4/floris/core/turbine/turbine.py", line 301, in thrust_coefficient
    thrust_coefficient += (
  File "/mnt/c/Users/Bart.Doekemeijer/OneDrive - Shell/Documents/python_scripts/floris_v4/floris/core/solver.py", line 511, in cc_solver
    turb_Cts = thrust_coefficient(
               ^^^^^^^^^^^^^^^^^^^
  File "/mnt/c/Users/Bart.Doekemeijer/OneDrive - Shell/Documents/python_scripts/floris_v4/floris/core/core.py", line 177, in steady_state_atmospheric_condition
    cc_solver(
  File "/mnt/c/Users/Bart.Doekemeijer/OneDrive - Shell/Documents/python_scripts/floris_v4/floris/floris_model.py", line 483, in run
    self.core.steady_state_atmospheric_condition()
  File "/mnt/c/Users/Bart.Doekemeijer/OneDrive - Shell/Documents/python_scripts/floris_v4/examples/test_cc_with_cubature_grid.py", line 5, in <module>
    fmodel.run()
ValueError: non-broadcastable output operand with shape (1,3) doesn't match the broadcast shape (1,1,9,3)
```

## Floris version
FLORIS v4.1.1, on commit `044ab0ff31`.

## System Information

 - OS: Ubuntu 22.04.3 LTS
 - Python version: 3.11.8
 - Library versions
   - `numpy==1.26.4`
   - `numexpr==2.10.1`
   - `scipy==1.14.0`

[...]

PR #1170, opened 2025-12-24
[BUGFIX] Enable Cumulative Curl model to work with a cubature grid

misi9170:
As raised in #970 and originally in #709 , the Cumulative Curl model does not currently work with a cubature grid. It took me quite a while to track the issue down, but finally, the solution was simply to add dimensions to `turb_avg_vels` for use in the initial computation of `thrust_coefficient` and `axial_induction`.

@Bartdoekemeijer kindly provided a test script:
```python

fmodel = FlorisModel("inputs/cc.yaml")
fmodel.set(solver_settings={"type": "turbine_cubature_grid", "turbine_grid_points": 3})
fmodel.run()
print(fmodel.get_turbine_powers())
```

On the develop branch, this fails, producing the error reported in #970. After this bugfix, the code runs as expected, producing
```
[[1746971.6830192   471716.37117722  467636.89587481]]
```

Moreover, commenting out the line `fmodel.set(solver_settings={"type": "turbine_cubature_grid", "turbine_grid_points": 3})` (which is to say, using the "standard" square grid) produces
```
[[1753954.45917917  381082.81148362  446768.73505829]]
```
on both the develop branch and this bugfix branch (that is, no change to results when not using a cubature grid). The difference in powers between the cubature grid and square grid is expected.

All tests pass and this PR is ready for review and merger.

### DIFF
--- a/floris/core/solver.py
+++ b/floris/core/solver.py
@@ -529,9 +529,9 @@ def cc_solver(
             + (flow_field.u_initial_sorted - turb_u_wake) * mask2
         )
 
-        turb_avg_vels = average_velocity(turb_inflow_field)
+        turb_avg_vels = average_velocity(turb_inflow_field)[:, :, None, None]
         turb_Cts = thrust_coefficient(
-            turb_avg_vels[:, :, None, None],
+            turb_avg_vels,
             flow_field.turbulence_intensity_field_sorted,
             flow_field.air_density,
             farm.yaw_angles_sorted,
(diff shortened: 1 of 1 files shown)


Bartdoekemeijer:
Thanks for fixing this!!

misi9170:
@Bartdoekemeijer , I've added you as a reviewer (something I should have thought of earlier)---if you'd rather not be, let me know and I'll ask someone else for a quick review. After that we should be able to get this merged into develop.

misi9170:
Going ahead and merging, as this change is very minor and tests are passing.

PR #1172, opened 2025-12-30
Update package dependences (dependabot)

misi9170:
We recently introduced dependabot #1064 , and v4.6 is the first FLORIS version to include it; since v4.6 was released, dependabot opened four PRs to update dependencies: #1166, #1167, #1168, #1169. I have now tried various installations of FLORIS to confirm these, and the updates to `isort`, `pytest`, and `sphinx-autodoc-typehints` seem fine; however, the update to `jupter-book` I am not currently OK with (see below). This PR combines the acceptable updates into a single PR to avoid having to merge three separate PRs, and also updates the `labels` field of the dependabot.yml to prevent dependabot from automatically creating new labels (once this PR is merged and the ones dependabot opened are closed without merger, I will delete the automatically created labels).

Regarding `jupyter-book`: `jupyter-book` v2 seems to require that node.js is installed. See [here](https://blog.jupyterbook.org/posts/2024-11-15-jupyter-book-2-alpha/) for more info. This seems an unnecessary extra dependency (that I'm not sure can be installed via pip), so I've left `jupyter-book` fixed at `jupyter-book~=1.0` for the time being. @rafmudaf , would you like to weigh in on this at all?

Further, our current set-up of dependabot widens the version requirements to allow multiple major versions (e.g. from `isort~=5.0` to `isort>=5,<8`). @paulf81 , back in #748, I think you were making an argument for trying to stick to a single major version for each requirement using the compatible-release specifier. Any opinions on whether we should stick to a single major version per requirement or are happy with a range? If we want to go with a single major version, we could move up to the latest available (that works), which would be `isort~=7.0`, `pytest~=9.0`, `sphinx-autodoc-typehints~=3.0`.

Note that this PR only affects the "docs" and "develop" dependencies, not the main dependencies for simple installation of FLORIS. I believe this is simply luck this time around.

I have tried:
- `isort` versions `5.13.2` (installed under current specification), `6.1.0`, and `7.0.0`.
- `pytest` upgraded to `9.0.2`
- `sphinx-autodoc-typehints` upgrade to `3.6.0`

When installing with `jupyter-book~=1.0`, version `1.0.4.post1` is installed and works (including with the upgraded `sphinx-autodoc-typehints`); upgrading to `jupyter-book` version `2.1.0` triggered a prompt to install node.js, which I aborted.

This is on my 2023 M3 Mac Pro.

_Alternatively_, we could simply take the dependabot PRs as informational only and non-critical, and simply close them (and this PR) without merger at this time.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,8 @@ updates:
   target-branch: "develop"
   schedule:
     interval: "monthly"
+  labels:
+    - "package"
   ignore:
     - dependency-name: "*"
       update-types: ["version-update:semver-minor", "version-update:semver-patch"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,17 +46,17 @@ dependencies = [
 docs = [
     "jupyter-book~=1.0",
     "sphinx-book-theme~=1.0",
-    "sphinx-autodoc-typehints~=2.0",
+    "sphinx-autodoc-typehints>=2,<4",
     "sphinxcontrib-autoyaml~=1.0",
     "sphinxcontrib.mermaid~=1.0",
     "bokeh~=3.7",
 ]
 develop = [
-    "pytest~=8.0",
+    "pytest>=8,<10",
     "pytest-benchmark~=5.1",
     "pre-commit~=4.0",
     "ruff~=0.9",
-    "isort~=5.0"
+    "isort>=5,<8"
 ]
 
 [tool.setuptools.packages.find]


rafmudaf:
Jupyter Book v2 uses MyST instead of Sphinx. At least, I think it's *instead of* rather than *in addition to*. See https://jupyterbook.org/stable/community/ecosystem/#a-comparison-with-jupyter-book-1-and-sphinx.

For FLORIS, this means that everything in the [sphinx config block](https://github.com/NREL/floris/blob/main/docs/_config.yml#L42) will have to be mapped to the MyST equivalent. FWIW some of that may be available by default. In any case, I think it may be worthwhile, if there's time.

On another note, I had some trouble using node and npm (node package manager) on the NREL VPN in the past. If you are interested in going down that path, double checking that you can actually install node and project dependencies might be a good first step. Node is not installed via pip, you could think of it as a the pip-equivalent for javascript/typescript.

misi9170:
@rafmudaf , Thank you very much for the additional context here. Jumping down the rabbit hole (or at least sticking my head in), it looks like the Jupyter Book v2 workflow will be easier once we get it set up. _However_, at this time, it appears that Jupyter Book v2 does not support automatically generating API documentation (because MyST doesn't yet support this), which I would say is a critical piece of the FLORIS documentation. Based on that, I think we need to wait on moving to Jupyter Book v2.

See:
- https://jupyterbook.org/stable/resources/faq/#known-limitations
- https://github.com/jupyter-book/mystmd/issues/1259

paulf81:
@misi9170 , maybe the best solution for right now is to merge the updates minus the changes to jupyter-book and then the two of us can track https://github.com/jupyter-book/mystmd/issues/1259 (or open an issue on FLORIS tracking it) and then bump jupyter-book when that's ready?

misi9170:
@paulf81 That works for me. Any opinion on whether we should bump dependencies all up to strictly requiring the latest major version (`isort~=7.0`, `pytest~=9.0`, `sphinx-autodoc-typehints~=3.0`) vs. just widening the requirements to include the latest major versions as well as the ones we currently specify? I think both will work

paulf81:
I feel like the wider permissions on these less critical dependencies is acceptable, so would vote that way

misi9170:
Thanks @paulf81 , that works for me. I've listed you as a reviewer---hope that's ok.

paulf81 (approved):

misi9170:
#1173 records the issue to keep track of MyST updates and eventually move to Jupyter Book v2.

misi9170:
Grrr, documentation build broke upon merger to develop. I'm looking into it.

misi9170:
I've replicated the [documentation build failure](https://github.com/NREL/floris/actions/runs/20760615777/job/59614294593) on my machine. It seems to come from the install of `ruamel.yaml` version 0.19, rather than version 0.18. I'm not totally sure why this is now being installed, because it seems to be a dependency of `sphinxcontrib_autoyaml` rather than anything that was changed in this PR, but perhaps I'm just missing something. Specifying `ruamel.yaml<0.19` (following https://github.com/FunAudioLLM/CosyVoice/issues/1769) fixes the issue for me locally. I am going to try pushing that to develop to see if that works.

misi9170:
I've ended up pushing a requirement that `ruamel.yaml~=0.18.0`, that has [fixed the docs build](https://github.com/NREL/floris/actions/runs/20762047532/job/59619247516) for now (`ruamel.yaml` 0.18.7 is installed and runs as expected).

PR #1176, opened 2026-01-13
Allow users to pass in their own turbine operation models

misi9170:
FLORIS v4 introduced the notion of turbine "operation models", which model how a turbine responds to different control setpoints and atmospheric conditions. See [the documentation](https://nrel.github.io/floris/operation_models_user.html) for details.

Until now, users specify which operation model they'd like to use by either setting the `operation_model` key on the turbine input yaml, or passing one of the predefined strings (e.g. `"cosine-loss"`, `"simple-derating"`, etc) to [`FlorisModel.set_operation_model()`](https://github.com/NREL/floris/blob/bcfea64fd3fea6316e82fd33ce57553cd47260cc/floris/floris_model.py#L1575). 

This PR enables users to pass in a compatible operation model class of their own making. Currently included are the functional changes (which are minimal, and limited to turbine.py and floris_model.py), detailed documentation showing how this feature can be used, and tests. I also plan to make a small example showing how this can be done [EDIT: this is now done]. Note also that the existing, prepackaged operation models can also be passed in this way, if desired. 

Finally, this also adds tests to show that the operation model class passed can be a dynamic class (all prepackaged operation models are static classes at this time). [EDIT: now done]

As part of this PR, I'm also renaming `FlorisModel.get_turbine_ais()` to the more explicit `FlorisModel.get_turbine_axial_induction_factors()`, which is more consistent with other methods used to extract quantities from FLORIS (e.g. `FlorisModel.get_turbine_thrust_coeficients()`). I had originally added backwards compatibility, but this is now going to go into a new version of FLORIS, so I'll go ahead with the breaking change.

To do:
- [x] Check integration [good for now for a beta version]
- [x] Add or update examples to show usage
- [x] ~Update docs page with exact minor version number of expected release~
- [x] Clean up attrs declaration of `operation_model` attribute on `Turbine` class


misi9170:
For future reference, this PR also contains a new documentation page providing information on how to implement a new operation model, and a short example demonstrating the functionality (examples/examples_operation_models/002_define_operation_model.py)

skygering:
Still running on my end!

Issue #760, opened 2023-12-08
Velocities at the ground


rafmudaf:
# Treating velocities at z=0
The u-component of velocity is initialized with a shear profile that enforces a no slip condition at the ground (when `z_sorted` is 0, the shear profile is 0):
```python
wind_profile_plane = (grid.z_sorted / self.reference_wind_height) ** self.wind_shear
```

However, the integration of the wake deficit with the background flow simply subtracts the two:
```python
flow_field.u_sorted = flow_field.u_initial_sorted - wake_field
```

Given this solver structure, the wake models themselves are responsible for handing the velocity at the ground. This separation of concerns is valid since the solver should not contain the physical models but instead orchestrate the coupling between the models.

Additionally, the Gauss-Curl Hybrid model contains methods for calculating the transverse velocity components (`v` and `w` in addition to `u`) in [floris.simulation.wake_deflection.gauss.calculate_transverse_velocity](https://github.com/NREL/floris/blob/main/floris/simulation/wake_deflection/gauss.py#L350), but a no slip condition at the ground is not enforced.

This was found in #699 since @vallbog noted that the derivative of the streamwise velocity component (du/dz) [initialized in the FlowField](https://github.com/NREL/floris/blob/main/floris/simulation/flow_field.py#L131) results in a divide-by-zero at z=0 and the wind shear exponent is less than 1 (discussed in https://github.com/NREL/floris/pull/699#discussion_r1420977808). In that case, a runtime warning is issued and the transverse velocity components should be undefined (though I haven't tested that this is what FLORIS produces). #699 instead sets the derivative to 0 when the wind shear is 0 to avoid the divide by zero. This change is relevant to this issue because the ultimate question is how to treat this derivate at the ground since it's actually more appropriate to be undefined.

## Proposed solution
- [ ] All velocity components should always be 0 at a point where z=0
- [ ] Tests of the wake models at z=0 should show 0 velocity
- [ ] Tests of the transverse velocity components should show 0 velocity

## Additional information

Here's the beginning of a test for this:
```python
    # What happens when the wind shear and height are 0?
    tdict = turbine_grid_fixture.as_dict()
    tdict["grid_resolution"] = 3  # Use an odd number to put a point at the origin
    coords = np.array(tdict["turbine_coordinates"]).T  # Transpose to get (x, y, z)
    coords[2, :] = 0.0  # Set all z coordinates to 0.0
[...]

PR #1175, opened 2026-01-13
WindIO support in FLORIS

lejeunemax:
# WindIO support in FLORIS

This PR introduces support for [WindIO](https://github.com/IEAWindSystems/windIO) in FLORIS.  
It enables running FLORIS wake simulations directly from WindIO-compliant input files, while keeping WindIO as an optional dependency.

## Summary of changes

The `fmodel` now includes three new methods:

- **`from_windio`**  
  Reads a WindIO YAML farm file and runs a wake model simulation based on its contents.

- **`set_wind_data_from_windio`**  
  Updates the wind farm geometry and turbine data based on a specified WindIO *wind farm* file.

- **`set_farm_from_windio`**  
  Updates wind data based on a specified WindIO *wind resource* file.  
  Currently supported wind resource formats include:
  - Time series
  - Wind rose
  - Weibull distribution

## WindIO parsing and field tracking

WindIO input files are read using the modified dictionary class  
`read_windio/utils/TrackedDict`.

This class tracks which WindIO fields are accessed during parsing and emits a warning when keys from the WindIO input are not used. This mechanism helps:

- Identify which WindIO fields are currently mapped to FLORIS
- Highlight which inputs are ignored or not yet supported

## Dependency management

WindIO is imported at runtime and therefore remains an **optional dependency** of FLORIS.  
Users who do not rely on WindIO workflows are not affected.

## Impacted areas of the software

Most changes are contained within the `read_windio` directory and are therefore largely orthogonal to other ongoing or future developments.

To comply with the WindIO output standard, a new `keep_inertial_frame` option was added to the initialization of `FlowFieldPlanarGrid`.  
This option allows the FLORIS flow field to be computed on a fixed inertial grid, rather than in a frame aligned with the main wind direction, which is required for compatibility with WindIO outputs.

## Optional TODOs

- [ ] Review and update documentation as needed
- [ ] Add tests for WindIO-based workflows
- [ ] Check for unmapped or unsupported features (e.g. `blockage_model`, `axial_induction_model`, `rotor_averaging`)
- [ ] Enforce consistent FLORIS-style logging across the new code

PR #1183, opened 2026-02-11
Clean up grid classes

misi9170:
This PR makes several adjustments to the `Grid` classes:
- Removes `FlowFieldGrid` entirely, since it is not used anywhere in FLORIS and I believe was deprecated in the transition from v3 to v4
- Adds a warning to `Grid` classes (via an `attrs` validator) to let users know that issues may arise if `z` is less than or equal to zero (more on this below)
- Moves `grid_resolution` validators onto the subclasses of `Grid` (see below)

This originally came up because the `where=z != 0` line in `np.power()` on `FlowField` was raising a numpy user warning, but digging into this more, I realized that any time `z` is less than or equal to zero can cause FLORIS to fail. Rather than generate an error, which might be a breaking change in case there are some workflows where non-positive `z` works, I opted to raise a warning to help users track down any issue that might arise.

In doing so, I believe I am also sufficiently addressing #760.

Finally, I noticed a TODO in the validators for the `grid_resolution` attribute on `Grid`, and according to [this stackoverflow thread](https://stackoverflow.com/questions/48788285/python-attrs-validator-in-child-class), I decided to remove the generic `grid_resolution` validator and instead add checks for correctness to the `__attrs_post_init__` of each subclass.

To check the new behavior, the following script can be run (based on examples/examples_visualization/003_visualize_cross_plane):
```python

from floris.flow_visualization import visualize_cut_plane

fmodel = FlorisModel("../inputs/gch.yaml")

# Set a 1 turbine layout
fmodel.set(
    layout_x=[0],
    layout_y=[0],
    wind_directions=[270],
    wind_speeds=[8],
    turbulence_intensities=[0.06],
)

# Collect the cross plane downstream of the turbine
cross_plane = fmodel.calculate_cross_plane(
    y_resolution=100,
    z_resolution=100,
    downstream_dist=500.0,
    z_bounds=[0, 500], # Raises warning, and also errors out. Changing lower bound to a value greater than zero works
)

# Plot the flow field
fig, ax = plt.subplots(figsize=(4, 6))
visualize_cut_plane(
    cross_plane, ax=ax, min_speed=3, max_speed=9, label_contours=True, title="Cross Plane"
)

# plt.show()
```
This script errors out with an ambiguous error `ValueError: z array must not contain non-finite values within the triangulation` on the develop branch, which the new warning should help users identify.

With the proposed change, this error is still raised, but is preceded by the added warning:
```
floris.logging_manager.LoggingManager WARNING Non-positive z coordinates detected. This may cause issues with the flow model calculations. To fix this, consider adjusting the z coordinates to be positive.
```

I have added tests under a new test script grid_unit_test.py to ensure that the warning is correctly raised and to check the `grid_resolution` validators. Probably, more could be added here.


lejeunemax:
Maybe this PR is the opportunity to integrate the [keep_inertial_frame](https://github.com/NatLabRockies/floris/pull/1175/changes/b264d0f982dd7ec85693542ae0c7d0d641bc181c) argument of #1175.  I believe this would allow to isolate better responsibilities.

misi9170:
> Maybe this PR is the opportunity to integrate the [keep_inertial_frame](https://github.com/NatLabRockies/floris/pull/1175/changes/b264d0f982dd7ec85693542ae0c7d0d641bc181c) argument of #1175. I believe this would allow to isolate better responsibilities.

@lejeunemax good idea! And sorry that I've been slow to get to #1175. Would you like to add the `keep_inertial_frame` option to this branch? If that doesn't work (i.e. you can't commit to this branch), I can just grab those changes myself.

EDIT: actually, sorry, let me take a day to just look over #1175 in more detail and make sure I understand the purpose and usage of `keep_inertial_frame`. I'll try to get back to you by the end of the day

misi9170:
@lejeunemax , I'm realizing I never got back to you about including the `keep_inertial_frame` argument in this PR. I decided that, since this PR is more of a bug fix than adding any new features, I think we should keep that change grouped with #1191 . This PR will then be part of a patch release, whereas I think eventually #1191 should be part of a more significant release (either a minor release, e.g. FLORIS v4.7, or a farther out major release, [FLORIS v5](https://github.com/NatLabRockies/floris/discussions/categories/v5-design-discussion)).

floris/core/core.py
```diff
@@ -104,11 +103,9 @@ def __attrs_post_init__(self) -> None:
                 grid_resolution=self.solver["turbine_grid_points"],
             )
         elif self.solver["type"] == "flow_field_grid":
-            self.grid = FlowFieldGrid(
-                turbine_coordinates=self.farm.coordinates,
-                turbine_diameters=self.farm.rotor_diameters,
-                wind_directions=self.flow_field.wind_directions,
-                grid_resolution=self.solver["flow_field_grid_points"],
+            raise ValueError(
+                "The flow_field_grid solver type is deprecated as of FLORIS v4.6.6 in favor of "
```
rafmudaf: Just a nuance - "deprecated" means the thing is still there but shouldn't be used and will be removed in the future. It's a heads up that there's going to be a breaking change soon. Here, the breaking change is already in place. That being said, I think this breaking change is fine, and I might consider removing this branch of the `elif` entirely.

rafmudaf: Otherwise, this looks good @misi9170

misi9170: Thanks @rafmudaf---I had been using "deprecated" a bit incorrectly :) 

I've removed the `elif` case, as you suggested.

PR #1184, opened 2026-02-23
Use PyPI Trusted Publishing workflow

misi9170:
This PR updates the python-publish workflow to use PyPI Trusted Publishing. In addition to the changes to python-publish.yml, I have also created a new Environment in the repository named `release-pypi`.

Details on trusted publishers here: https://docs.pypi.org/trusted-publishers/

Thanks @RHammond2 for the help! I've requested you as a reviewer---would you mind taking a look over the changes here and see if they look right? I always find changes to the publishing workflow a bit scary, as I don't know how to test them prior to just releasing a new version and hoping that it works...

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,5 +1,7 @@
 # This workflows will upload a Python Package using Twine when a release is created
+# Published via GitHub Actions as a PyPI Trusted Publisher.
 # For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+# and: https://docs.pypi.org/trusted-publishers/
 
 name: Upload Python Package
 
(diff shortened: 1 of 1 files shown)


.github/workflows/python-publish.yml
```diff
@@ -9,22 +11,23 @@ on:
...
-        python -m pip install --upgrade pip
-        pip install build twine
-    - name: Build and publish
-      env:
-        TWINE_USERNAME: __token__
-        TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
-      run: |
-        python -m build
-        twine upload dist/*
+      - uses: actions/checkout@v3
```
RHammond2:
```suggestion
      - uses: actions/checkout@v6
```

Between this and the Python version update, you should be good to not think about it again for a while.

.github/workflows/python-publish.yml
```diff
@@ -9,22 +11,23 @@ on:
...
+      - name: Set up Python
+        uses: actions/setup-python@v4
```
RHammond2:
```suggestion
        uses: actions/setup-python@v6
```

.github/workflows/python-publish.yml
```diff
@@ -9,22 +11,23 @@ on:
...
+        run: |
+          python -m pip install --upgrade pip
+          pip install build twine
+          python -m build
+          twine check --strict dist/*
+      - name: Publish package to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          verbose: True
+          password: ${{ secrets.PYPI_PASSWORD }}
```
RHammond2:
```suggestion

```

This line will no longer be necessary because of the trusted publishing process, which is essentially saying that packages pushed from this repository with this action and the specified environment don't need further authentication.

RHammond2 (changes requested):
I left a couple suggestions and comments, and this should be good to go. Just be sure to add create an environment in the project's settings such that the name matches your trusted publisher (no changes required you name it).

RHammond2 (approved):

PR #1192, opened 2026-05-13
fix: correct tilt_interp parameter type in compute_tilt_angles_for_fl…

zredb:
…oating_turbines

- Change tilt_interp parameter type from dict[str, interp1d] to interp1d|None
- This aligns with the function's actual usage where it expects a single interpolation function or None


misi9170:
Good catch and thanks for fixing this @zredb ! I've made one very minor formatting change to the type hint, just to match with other type hints that accept a range of types (` | `). Once tests pass I'll get this merged into develop, and it will appear on the main branch in the next FLORIS release.

Issue #1193, opened 2026-05-29
Compatibility pandas >= 3.0


thebaptiste:
There is no hurry but a compatibility with `pandas >= 3.0` would be nice
Regards

misi9170:
Closed by #1194

thebaptiste:
Thanks. When will you release with this fix ?

misi9170:
Hopefully, by the end of the week or early next week. I'm planning to get #1183 merged (it's been sitting a long while), and then will create a patch release (4.6.5) once that's in.

PR #1194, opened 2026-05-29
Check allowing pandas v3

misi9170:
As raised in #1193 , [pandas v3](https://pandas.pydata.org/docs/whatsnew/v3.0.0.html) was released earlier this year and it would be nice to have FLORIS be compatible. 

I have so far simply updated the dependencies and run tests, but not examples; opening this PR should run the examples. Assume they all execute correctly without error, it's likely we can simply support both pandas v2 and pandas v3, which would be my preference over dropping pandas v2 support.


paulf81:
Works for me!

misi9170:
Within the core code, I think pandas is only used in two places:
- `solve_for_velocity_deficit_profiles()` on `Core` in floris/core/core.py, where the return object is a pandas dataframe
- `_initialize_multidim_power_thrust_table()` on `Turbine` in floris/core/turbine/turbine.py, where pandas is used to read and extract values from a csv.

It's also used rather minimally in two examples and one test (after I've now removed some unused imports). I'm comfortable allowing pandas 3 (but will continue to allow pandas 2, as I suspect many users will not have upgraded)

rafmudaf:
FWIW we could eliminate the second usage with [numpy.loadtxt](https://numpy.org/doc/stable/reference/generated/numpy.loadtxt.html#numpy.loadtxt). The first one might be more involved, but generally it feels like an anti pattern if that's the only place where we're using a dataframe.

@misi9170 what do you think about eliminating Pandas as a dependency all together? I can take a stab at this if you agree.

misi9170:
@rafmudaf , I'd be very on board, but my one hesitation is that the `solve_for_velocity_deficit_profiles` routines, which are user facing, actually _return_ a pandas dataframe to the user, so it'd be a breaking change to eliminate that and return something else instead. That being said, I'd be happy for you to remove all other uses, and then maybe add a note (and a deprecation warning?) to remove that in future... what do you think? Alternatively, maybe this is a little-enough-used method that we can break the rules on breaking changes in this case (for the greater good of removing a rather hefty dependency)

rafmudaf:
@misi9170 I made the change in the turbine file. As for the velocity deficit profile function, I'm happy to add a warning, but I also lean toward your latter perspective. I think it's a lightly enough used function and we could make the change clear in the release notes.
FWIW it's a bigger effort to completely remove Pandas, so maybe the warning log is good enough for now.

rafmudaf:
And just for documentation, I used [np.genfromtxt](https://numpy.org/doc/stable/reference/generated/numpy.genfromtxt.html) because it creates a named array from the header row directly whereas `np.loadtxt` does not.

misi9170:
I agree on all fronts, thanks @rafmudaf . Looks like all checks are passing, so I'll merge unless you are planning on any more changes

rafmudaf:
Good to go from my end. Thanks @misi9170

PR #503, opened 2022-09-20
Refactor the solver module to a more general and class-based system

RHammond2:
**Feature or improvement description**
This is a prototype for refactoring `floris/simulation/solver.py` to have less repeated code blocks by moving from function-based operations to class-based operations with a base `Solver` class to implement the initialization and expected methods. In particular, the full flow solvers all followed the same routine, so this is implemented once in `Solver`.

**Related issue, if one exists**

**Impacted areas of the software**

- `floris/simulation/solver.py`
- `floris/floris.py` (when new API is settled)

**Additional supporting information**

This relates to a discussion point where the solver needed to be reconsidered because it was growing too much and having too many of the same code patterns being reused, so moving to a different module-level architecture was desired. 

@rafmudaf , this is definitely in a prototype stage, and thus isn't integrated into the code, but I wanted to start this PR to get the conversation going on un/desirable design patterns in a solver refactor. This will also have a lot of merge conflicts with some of the other PRs in the pipeline, as an FYI.

**Test results, if applicable**

PR #1197, opened 2026-07-10
Refactor wake models and solvers

misi9170:
This is a major PR that restructures how FLORIS implements wake models and solvers. In particular, it removes the capability of mixing and matching wake deficit, deflection, and turbulence models, instead unifying these into a single "model," and it combined the "solver" with the wake model itself.

This has the following benefits:
- It makes the interface clearer for the vast majority of users, that were not "mixing" different types of model
- It allows user-defined wake models to be passed in directly
- Turbine quantities (e.g. powers) will be computed during the main turbine solve, and can then be accessed at the `FlorisModel` level, making the structure more flexible in being able to handle different ways in which the turbine quantities are evaluated.

However, the following support will be removed:
- Mixing different parts of different wake models (e.g. Jensen deficit model with Gauss deflection model). These were never tested and likely did not fully work, so I consider this to be a reasonable loss.

In making this change, I have also been working with the `Farm` and `Turbine` classes some. At this stage, I'm not sure whether I'll keep those changes or revert them, or create a separate PR for them. Time will tell. EDIT: these were separated into #1200, which is now merged.

Still to do:
- [x] Access turbine functions from the operation_model directly #1200 
- [x] Convert all inbuilt wake models to the new structure [remaining: CC, TurboparkGauss (remove TurboPark)]
- [x] Update FLORIS input files, and decide on back compatibility
- [x] Clean up `WakeModelManager` usage
- [x] Build converters from v4 input files to v5.
- [x] Decide whether to support various combination models [decided to continue to support various pre-packaged combination models, although I've refactored how this is done slightly]
- [x] Look for opportunities to cut down on repeated code [likely other opportunities, but the big one (matching deflection model in `Gauss`, `CumulativeCurl` wake models) is addressed]
- [x] ~Update documentation~ [I'm going to leave this out of this PR. It'll certainly be needed, but with other possible structural changes to come in v5, it'll be easier to document them all in one dedicated PR.]
- [x] Calculate turbine quantities (e.g. power) during solve and store on `Farm` [Leaving for a separate PR; edit: decided to go ahead with this after all, so it's now in]

Some items on this TODO list may be left for a future PR, depending on how I get on.

This PR supersedes #503.


misi9170:
I've run some profiling similar to what I did [here](https://github.com/NatLabRockies/floris/pull/1200#issuecomment-5007151391). 

Here is the profiling on the base dev/v5-beta branch:

image




And again on the refactored ehmt/solvers branch:

image

The upshot is that there is not much of a performance change, and I was not expecting there to be. The refactored code seems to run slightly faster, but that could be just a fluke.

misi9170:
Assuming all checks and tests pass, I'm will merge this without review, since it is going into a secondary development branch. There may be some bugs; if that is the case, I will open bugfixes to dev/v5-beta as they come up.

PR #1199, opened 2026-07-13
Fix NaN turbine power at 0 m/s inflow in Gauss wake models

### LINKED ISSUE #1069: Gauss velocity deficit fails with 0 m/s inflow
# Gauss fails at 0 m/s

If you run the Gauss velocity deficit model without secondary_steering, yaw_added_recovery, active_wake_mixing, enable_transverse_velocities, the model returns NaN when you provide an inflow of 0 m/s.

## How to reproduce

I took the gch.yaml in the inputs folder, and changed: 

  # Can be "true" or "false".
  enable_secondary_steering: false

  ###
  # Can be "true" or "false".
  enable_yaw_added_recovery: false

  ###
  # Can be "true" or "false".
  enable_transverse_velocities: false

Then, a simple script: 

## System Information

 - OS: <e.g. Ubuntu 20.04 or macOS 10.12>
 - Python version: <Result of `python --version`>
 - Library versions
   - Results of `pip freeze`, for example
   - matplotlib
   - numpy
   - numexpr
   - scipy
   - pandas

steps-re:
## Summary

At a freestream wind speed of 0 m/s, the Gauss velocity-deficit and deflection models return NaN for every turbine's power. This PR guards the zero-inflow case so a 0 m/s inflow produces zero wake, zero deflection, and zero power, which is the physically correct result.

Fixes #1069

## Root cause (two sites)

The initial Gaussian wake width uses `sqrt(uR / (u_initial + u0))`. Both `uR` and `u0` are proportional to the inflow velocity, so at 0 m/s the ratio is `0 / 0 = NaN`. That NaN flows into `sigma_y0`/`sigma_z0` and then into every downstream turbine power.

- `floris/core/wake_velocity/gauss.py` — `sigma_z0 = rotor_diameter_i * 0.5 * np.sqrt(uR / (u_initial + u0))`
- `floris/core/wake_deflection/gauss.py` — the same `sigma_z0` ratio, plus `C0 = 1 - u0 / freestream_velocity`, which is `1 - 0/0 = NaN` at zero inflow.

Fixing only one site still leaks NaN, because the deflection field is fed back into the velocity-deficit solve.

## The fix

Guard the division at every site so `den > 0` is computed normally and the zero-inflow entries get the physically correct fill:

- **Velocity model:** mask the `sigma_z0` ratio so a zero inflow gives zero wake width.
- **Deflection model:**
  - `C0`: analytically `u0 / freestream_velocity == sqrt(1 - ct)`, so `C0 -> 1 - sqrt(1 - ct)` in the zero-inflow limit. The guard fills that value rather than 0 (the deflection expressions need the correct `C0`/`M0`/`E0` for the `den > 0` case, and this preserves them exactly).
  - `sigma_z0`: mask the ratio (zero wake width at zero inflow).
  - Far-wake width ratio `sigma_y * sigma_z / (sigma_y0 * sigma_z0)`: masked to avoid a spurious divide warning once the widths are zero.
  - The returned deflection field is forced to zero where `freestream_velocity == 0`, since there is no wake and therefore no deflection at zero inflow. This also stops any residual NaN from propagating into the velocity solve.

The guards use `np.divide(num, den, out=..., where=den > 0.0)`, so behavior for all `den > 0` cases is bit-for-bit identical to before; only the previously-NaN zero-inflow entries change.

## Test

Adds `test_regression_zero_inflow` to `tests/reg_tests/gauss_regression_test.py`. It builds the gauss/gauss model with `enable_secondary_steering`, `enable_yaw_added_recovery`, and `enable_transverse_velocities` all disabled, runs at `wind_speeds=[0.0]`, and asserts the turbine powers are finite (`not np.isnan(...).any()`) and all zero (`np.allclose(..., 0.0)`).

- Fails on `main` (all-NaN power).
- Passes with this fix.

## Validation

- New regression test: fails before the fix, passes after.
- `pytest tests/ -k "gauss or velocity or deflection"`: 31 passed.
- `pytest tests/reg_tests/ tests/wake_unit_tests.py`: 50 passed (no regressions on any wake model or normal wind speed).

## AI disclosure

Developed with AI assistance (Claude); I have reviewed, understand, and tested every line.

### DIFF
--- a/floris/core/wake_deflection/gauss.py
+++ b/floris/core/wake_deflection/gauss.py
@@ -157,12 +157,25 @@ class GaussVelocityDeflection(BaseModel):
         ky = self.ka * turbulence_intensity_i + self.kb
         kz = self.ka * turbulence_intensity_i + self.kb
 
-        C0 = 1 - u0 / freestream_velocity
+        # Guard against the zero-inflow case (freestream_velocity == 0), where u0 is
+        # also zero and u0 / freestream_velocity would be 0/0 = NaN. Analytically
+        # u0 / freestream_velocity == sqrt(1 - ct_i), so C0 -> 1 - sqrt(1 - ct_i) in
+        # that limit. Fill the zero-inflow entries with that value.
+        C0_fill = (1 - np.sqrt(1 - ct_i)) * np.ones_like(freestream_velocity)
+        C0 = 1 - np.divide(
+            u0, freestream_velocity, out=1 - C0_fill, where=freestream_velocity > 0.0
+        )
         M0 = C0 * (2 - C0)
         E0 = ne.evaluate("C0 ** 2 - 3 * exp(1.0 / 12.0) * C0 + 3 * exp(1.0 / 3.0)")
 
         # initial Gaussian wake expansion
-        sigma_z0 = ne.evaluate("rotor_diameter_i * 0.5 * sqrt(uR / (freestream_velocity + u0))")
+        # Same zero-inflow guard: uR and u0 are both zero at freestream_velocity == 0,
+        # so the ratio is 0/0 = NaN. A zero inflow produces zero wake, so the initial
+        # wake width is zero there.
+        u_sum = freestream_velocity + u0
+        sigma_z0 = rotor_diameter_i * 0.5 * np.sqrt(
+            np.divide(uR, u_sum, out=np.zeros_like(u_sum), where=u_sum > 0.0)
+        )
         sigma_y0 = sigma_z0 * cosd(yaw_i) * cosd(wind_veer)
 
         # yR = y - y_i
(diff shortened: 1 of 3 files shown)


rafmudaf:
Just out of curiosity, what's the situation where you use 0 m/s wind speed? Also, did you check whether the other models also return nans or 0s?

misi9170:
See also further discussion on this topic in #1076

steps-re:
Thanks for the quick look, @rafmudaf and @misi9170.

On the use case: 0 m/s tends to show up in time-series / measured-data
runs rather than as a deliberate input. The #1069 reporter hit it
feeding measured data through FLASC, where calm periods come through
as 0 m/s rows (and more generally any workflow that ends up with a
zero-speed findex). The goal isn't to model 0 m/s meaningfully, just
to keep one calm timestamp from turning the whole farm's result into
NaN.

On the other models: I checked at 0 m/s (270 deg, TI 0.06, two
turbines). Most return a clean 0:

- jensen -> 0
- turbopark / turboparkgauss -> 0
- empirical_gauss -> 0
- gauss -> 0 with the default GCH corrections ON, but NaN with
enable_secondary_steering / enable_yaw_added_recovery /
enable_transverse_velocities all OFF (the #1069 config). That is what
this PR fixes.
- cumulative-curl (cc) -> NaN even in its default config.

So gauss is the main one, but a heads-up that CC also NaNs at 0 m/s,
in a separate path this PR doesn't touch. Happy to fold a CC guard
into this PR or open a separate one, whichever you prefer. I'll also
read through #1076.

Mike

Issue #847, opened 2024-03-18
Standardize shapes of `hub_heights`, `rotor_diameters`, etc. on Farm object.


misi9170:
And issue (#833) was raised which was caused by a change in the dimensions of `Farm.hub_heights` during the `Farm.finalize()` call. Essentially, some attributes which might be expected to have a single dimension of length `n_turbines` are "finalized" with dimensions (`n_findex` x `n_turbines`). The following code demonstrates this change:

```python
import numpy as np
from floris import FlorisModel

fmodel = FlorisModel("inputs/gch.yaml")

# Changing the wind farm layout uses FLORIS' set method to a two-turbine layout
fmodel.set(layout_x=[0, 500.0], layout_y=[0.0, 0.0])
fmodel.set(wind_directions=np.array([270.0, 270.0, 270.0, 270.0]),
        wind_speeds=[8.0, 8.0, 10.0, 10.0],
        turbulence_intensities=np.array([0.06, 0.06, 0.06, 0.06]))
print("Prior to run():", fmodel.core.farm.hub_heights.shape)

# After the set method, the run method is called to perform the simulation
fmodel.run()
print("After run():", fmodel.core.farm.hub_heights.shape)
```

misi9170:
Note that the high-level bug in #833 has been addressed in #846 , but the underlying inconsistency on the dimensions of `hub_height` remains.

PR #1200, opened 2026-07-17
Refactor the Turbine and Farm class

misi9170:
This pull request makes some structural changes to the `Farm` and `Turbine` classes, with the core idea being that the `Turbine` class (and the turbine operation models assigned to it) were underutilized and obscure.

In particular, this PR:
- Passes a list of instantiated `Turbine` objects to `Farm` in a new `turbines` attribute which is the same length as the number of turbines in the farm.
- Removes `turbine_power_functions`, ... as attributes on `Farm`, instead using the operation model functions found in `turbines` explicitly`
- Simplifies the attributes on `Farm` by making almost all of the "`sorted`" versions of attributes `@property`s, and using attributes available on the `Turbine`s directly when possible. 

This PR does not implement a [`ControlSetpoints` class](https://github.com/NatLabRockies/floris/discussions/1141#discussioncomment-15087847), but `Farm` will likely be further simplified if/when that happens.

I believe this PR addresses #847 .

TODO:
- [x] Check all examples still run [some minor fixes were needed to fix 3 examples]
- [x] Perform timing tests to for speed changes [see [detailed comment below](https://github.com/NatLabRockies/floris/pull/1200#issuecomment-5007151391)]

### Not done in this PR but other improvements that could be made

- Implement [`ControlSetpoints` class](https://github.com/NatLabRockies/floris/discussions/1141#discussioncomment-15087847)
- Store `power_thrust_table` on the operation models (i.e., make the operation models dynamic classes) to save passing them through function calls
- ~Move the `power()`, `thrust_coefficient()`, and `axial_induction()` _functions_ on turbine.py to methods of the `Turbine` class, again to save passing around extra information~ [after consideration, I believe this would be less powerful, as it would _require_ that function evaluations happen in a sequential manner, rather than a possible vectorized operation. Instead, I have left the functions in place but indicated they are mostly for user convenience.]


misi9170:
I've now run some code profiling to understand any performance effects from the change. The upshot is that there appears to be a small (<10%) negative effect. To test, I ran the test script [profiling/profiling.py](https://github.com/misi9170/floris/blob/fb403f1e8515d34447ea71e1d9415fc57843ab05/profiling/profiling.py) using cProfile, i.e.
```
python -m cProfile -o <out>.prof profiling.py
```
profiling/profiling.py essential runs the GCH model for a medium-large problem (100 turbines, 1800 findices).

I ran this on both the ehmt/refactor-farm and the develop branches of FLORIS, and then used snakeviz to visualize the profiling results, i.e.
```
snakeviz <out>.prof
```

On both branches, the solve takes roughly 20 seconds on my laptop, with the new branch taking 0.3s longer in total. Here is the profile breakdown on develop:
image

In particular, the breakdown of calls to functions on farm.py and turbine.py are as follows:
image

image

(I have truncated the calls to farm.py to show only the top fifteen entries or so, but they are ordered by total time spent in the function calls so these should account for the majority of time.)

Now, running the snakeviz analysis on the new branch with the implemented changes gives:
image

and the breakdown on farm.py (again, truncated) and turbine.py:
image

image

The main difference between the two is that there are now substantially more function calls on farm.py. In particular, each time a "`_sorted`" version of a quantity (such as `rotor_diameters_sorted`) is called, a `numpy.take_along_axis()` operation is performed to return the sorted quantity. This also involves the `Farm._sorted_indices` attribute, so this attribute is called many times. This _does_ appear to result in slightly more time spent in `Farm` in the new implementation (but should save slightly on memory, since we do not need to save both the unsorted and sorted quantities, although I'm not sure if new memory is allocated with `numpy.take_along_axis()` or whether this is just a new "view"). Either way, if we are concerned about extra computation here, we could consider swapping this so that the "`_sorted`" versions of the quantities are the permanent attributes, since those are the ones that are called during the solve procedure, and the "unsorted" ones are the unsaved "views" that are accessed as `@property`s. Since the performance hit appears small here, I will leave that as a future possible performance improvement if it is needed.

floris/core/turbine/turbine.py
```diff
@@ -177,663 +504,340 @@ def power(
...
 
-    @rotor_area.validator
-    def reset_rotor_area(self, instance: attrs.Attribute, value: float) -> None:
-        """
-        Resets the `rotor_radius` value to trigger the recalculation of
-        `rotor_diameter`, `rotor_radius` and `rotor_area`.
-        """
-        self.rotor_radius = (value / np.pi) ** 0.5
+    Args:
+        velocities (NDArrayFloat): The velocity field at each turbine; should be shape:
```
rafmudaf: Update docstring

floris/core/turbine/turbine.py
```diff
@@ -177,663 +504,340 @@ def power(
...
-        # Collect reference information
-        power_thrust_table_ref = copy.deepcopy(self.power_thrust_table)
-        self.power_thrust_data_file = power_thrust_table_ref.pop("power_thrust_data_file")
-
-        # Solidify the data file path and name
-        self.power_thrust_data_file = self.turbine_library_path / self.power_thrust_data_file
-
-        # Read in the multi-dimensional data supplied by the user.
-        data = np.genfromtxt(self.power_thrust_data_file, delimiter=',', names=True)
+        velocities (NDArrayFloat[findex, turbines, grid1, grid2]): The velocity field at
```
rafmudaf: Update docstring

floris/core/turbine/turbine.py
```diff
@@ -119,50 +456,40 @@ def _select_multidim_condition(
...
     multidim_condition: dict | None = None,
 ) -> NDArrayFloat:
     """Power produced by a turbine adjusted for yaw and tilt. Value
     given in Watts.
 
     Args:
+        turbines (list[Turbine]): Instantiated turbine objects for each turbine in the farm.
         velocities (NDArrayFloat[n_findex, n_turbines, n_grid, n_grid]): The velocities at a
             turbine.
         turbulence_intensities (NDArrayFloat[findex, turbines]): The turbulence intensity at
```
rafmudaf: While we're at it, I think the `turbines` in all of these arrays as meant to be `n_turbines`

floris/core/farm.py
```diff
@@ -237,138 +210,46 @@ def check_turbine_type(self, attribute: attrs.Attribute, value: Any) -> None:
...
-            axis=1,
-        )
-        self.state = State.INITIALIZED
+    def initialize(self):
+        if hasattr(self, "_sorted_indices"):
+            self.state = State.INITIALIZED
+        else:
+            raise ValueError(
+                "The Farm object must be initialized with the sorted indices from the FlowField "
+                "before it can be used. Please call Farm.set_sorted_indices() first."
```
rafmudaf, line 229: The sorted indices come from the grids, i.e. `turbine_grid_farm.set_sorted_indices(turbine_grid.sorted_coord_indices)`

rafmudaf (commented):
Hey @misi9170 this is a big change, nice job! Here are a few notes all related to code comments.

floris/core/farm.py
misi9170: Good catch---fixed.

